PR #28, opened 2013-07-04
Controllers for exporting the state of force-torque and IMU sensors to ROS topics

adolfo-rt:
Controllers fetch all available sensors in an interface (eg. all registered IMU sensors), and publish their state in separate topics named after the sensor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,4 @@
+ros_controllers
+===============
+
+See [ros_control documentation](http://ros.org/wiki/ros_control) on ros.org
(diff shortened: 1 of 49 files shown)


adolfo-rt:
@wmeeusse, @jbohren, @davetcoleman. Similar to my previous request on ros_control, I'd be interested in merging this into hydro-devel as well, but I don't have catkin workspace setup. It would be great if one of you has extra cycles to build and run the tests of **ros_controllers** on a catkin system just to make sure that I didn't mess up the package.xml scripts.

PR #57, opened 2013-10-24
Added robot namespace when getting URDF info about the joint in the effort_controllers/JointPositionController

fsuarez6:
The `effort_controllers/JointPositionController` gets the URDF info about the joint from the `robot_description` parameter. This should include the namespace in order to allow configurations with multiple robots. With this fix the info will be obtain from `ns/robot_description`.


bbrieber:
I created another request which fixes the same issue #58 with a different solution... 
sorry for the duplication but i had not seen this one before...

fsuarez6:
I seems that both solutions work, although I think using `searchParam` is the right way to go.

PR #58, opened 2013-10-25
Update joint_position_controller.cpp

bbrieber:
The robot_description parameter is search in the gazebo namespace and not in the namespace of the controller.
This creates Problems when working with multiple different robots.

This fix added a search for the parameter in the robots own namespace before trying gazebo's namespace

### DIFF
--- a/effort_controllers/src/joint_position_controller.cpp
+++ b/effort_controllers/src/joint_position_controller.cpp
@@ -80,9 +80,20 @@ bool JointPositionController::init(hardware_interface::EffortJointInterface *rob
 
   // Get URDF info about joint
   urdf::Model urdf;
-  if (!urdf.initParam("robot_description"))
+  std::string key;
+  if (n.searchParam("robot_description", key))
   {
-    ROS_ERROR("Failed to parse urdf file");
+    std::string val;
+    n.getParam(key, val);
+    urdf.initString(val);
+    
+  }
+  //SEARCHING FOR THE PARAMETER FROM "/gazebo" downwards does not seem useful 
+  //THIS BLOCK MIGHT BE DELETED (replaced with else return false; or somthing like this)
+  else if (!urdf.initParam("robot_description"))
+  {
+    
+    ROS_ERROR("Failed to parse urdf file");    
     return false;
   }
   joint_urdf_ = urdf.getJoint(joint_name);


bbrieber:
Nearly the same as request #57   but used another solution

wmeeusse:
The model we've been using is that every robot has it's own controller manager, with it's own set of hardware interfaces and it's own set controllers. In that model, each robot is namespaced, and every robot has a robot_description parameter in its own namespace.  So you don't have to search for the robot_description parameter, you know exactly where to find it, which is in the root_nh passed on to the controller. 

The code to get the robot_description looks like this:
std::string robot_description
root_nh.getPararm(robot_description, "robot_description");
urdf.initString(robot_description);

The reason why urdf.initParam() does not work in Gazebo is that the urdf library will create its own nodehandle, which ends up to be in the global namespace in case you're running multiple robots in Gazebo. 

Hope this helps,
Wim

adolfo-rt:
It seems that the problem comes from how you are namespacing things in your simulation setup, and not from ros_control(lers). More specifically, the Gazebo backend is pushing the controller manager to the `gazebo` namespace, while te robot model is getting loaded in the root namespace. They should live at the same level (whichever makes more sense to you). Is there any particular reason for separating the controller manager and the robot description in different scopes?. The rationale provided by @wmeeusse above works very well for both single- and multi-robot setups.

bbrieber:
Hi again,

in my setup each robot has his own namespace. 
But if the controller manager searches in /gazebo instead of the namespace where the plugin is loaded it can't find the robot description

the output of `rosservice list` shows the the controller manager is spawned in the right namespace:

/my_robot/controller_manager/list_controller_types
/my_robot/controller_manager/list_controllers
/my_robot/controller_manager/load_controller
/my_robot/controller_manager/reload_controller_libraries
/my_robot/controller_manager/switch_controller
/my_robot/controller_manager/unload_controller

the robot_description is placed at 
/my_robot/robot_description 

on the parameter server
but if you just use urdf.initParam() the controller_manager starts searching for the parameter in /gazebo instead of in the namespace of the NodeHandle (/my_robot) that in provided as method parameters(as @wmeeusse allready said). So the current implementation from the trunk does not work with different namespaces.

The search function might be useful if the robot_description is in a subnamespace.
This might be useful if you are using more than one instance of the same robot.  And the overhead in communication is only done in the init function...

adolfo-rt:
The problem seems to lie in the Gazebo backend, and not on ros_control or its controllers. What Gazebo backend are you using?, if it's `gazebo_ros_control`, you should be able to specify the namespace in the SDF to be whatever you want ([SDF example](http://gazebosim.org/wiki/Tutorials/1.9/ROS_Control_with_Gazebo#Add_the_gazebo_ros_control_plugin), [relevant code](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp#L83)).

If you're using a different Gazebo backend, you should patch it so a correctly scoped NodeHandle is passed to the controller manager upon construction.

wmeeusse:
@bbrieber urdf.initParam() is indeed not going to work here, due to a limitation of the urdf library (which can/should get fixed).  That's why I was suggesting to grep the robot_description from the parameter server yourself, and then use urdf.initString(robot_description) instead. Can you verify that you're able to grep the robot description from the parameter server without using search param? If you can do that, you have a full solution without searching. If that doesn't work (i.e. the root_nh passed to the controller is in the gazebo namespace instead of in the robot namespace), then you have a problem to solve with the gazebo plugin.  The plugin should set the root_ns of your robot correctly. 

I see the search param not as a fix for this problem (the actual fix is described above), but as a newly proposed feature that would allow you to share a robot description between different robots. That's definitely a fair proposal. I'm a little worried about that feature though, because it could lead to some very difficult debugging, since you wouldn't know where the controller has found its robot description.  Is there another motivation for this feature except for solving the problem in this ticket?

Hope this helps,
Wim

adolfo-rt:
> I see the search param not as a fix for this problem (the actual fix is described above), but as a newly proposed feature that would allow you to share a robot description between different robots. That's definitely a fair proposal. I'm a little worried about that feature though, because it could lead to some very difficult debugging, since you wouldn't know where the controller has found its robot description.

I vote against sharing the same robot_description parameter across multiple robots. As you suggest, it is likely to cause more problems than it solves.

P.S. Sorry for prematurely closing the PR. I clicked comment&close by mistake. Reopened until the discussion is finalized.

adolfo-rt:
I'm closing this PR because it has been inactive for about a year. Feel free to open it if further discussion is required.

PR #63, opened 2013-11-11
Position controller

progtologist:
Added support for position - position controller.
Tested it with gazebo hardware interface (pull request https://github.com/ros-simulation/gazebo_ros_pkgs/pull/135) and works fine.

### DIFF
--- a/position_controllers/CMakeLists.txt
+++ b/position_controllers/CMakeLists.txt
@@ -15,18 +15,36 @@ if(USE_ROSBUILD)
 else()
 
   # Load catkin and all dependencies required for this package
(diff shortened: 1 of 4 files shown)


adolfo-rt:
Could you please detail a bit more the motivation for this pull request?. It's modifying an existing controller, and changing its behavior. The original implementation forwards position commands to a position interface, while the proposed changes seem to add additional PID dynamics between the command and what gets sent to the interface.

progtologist:
You are correct, I have misunderstood of how the original implementation was working. Please ignore this request as it is neither correct (in terms of control) nor needed.

adolfo-rt:
No problem. Thanks anyway for the interest.

PR #66, opened 2013-11-27
Install default config files

po1:
I don't see a good reason not to have them in the debian packages, and they probably work perfectly as is in a majority of cases.


adolfo-rt:
The files were left out unintentionally, they should be installed. Thanks!.

PR #76, opened 2013-12-19
Run tests for ros_controllers during Travis build.

ahendrix:


ahendrix:
The tests are failing to run because there are no tests for the effort_controllers package.

I see a couple of ways to proceed here:
- Hardcode the set of tests to run
- Write tests for packages that don't have them
- Run all of the existing tests, including tests for packages in the workspace but outside of ros_controllers.

.travis.yml
```diff
@@ -20,6 +20,8 @@ before_install: # Use this to prepare the system to install prerequisites or dep
   - export CI_SOURCE_PATH=$(pwd)
   - export REPOSITORY_NAME=${PWD##*/}
   - echo "Testing branch $TRAVIS_BRANCH of $REPOSITORY_NAME"
+  - export TESTS=$(for f in `ls $CI_SOURCE_PATH` ; do [ -d $f ] && echo -n "run_tests_$f " ; done)
```
adolfo-rt, line 23: How about this:

``` bash
# Populate a '|'-separated list of packages in the source path
export PKGS=$(for f in `ls $CI_SOURCE_PATH` ; do [ -d $f ] && echo -n "$f|" ; done)

# Filter these packages from the list of available targets (ie. look for run_tests_<package_name>)
# NOTE 1: Maybe you can optimize this command better, I overspent time getting to this
# point, awk is not my thing
# NOTE 2: I'm using 'make -C build' instead of 'catkin_make' because the latter also produces
# an extra 'Invoking "make" failed' in TESTS
export TESTS=`make -C build -qp | awk -F':' "/^(run_tests_)($PKGS):([^=]|$)/" | awk '{split($1,A,/:/);for(i in A)print A[i]}' | grep -v "^$"`
```

adolfo-rt:
> -    Hardcode the set of tests to run
> -    Write tests for packages that don't have them
> -    Run all of the existing tests, including tests for packages in the workspace but outside of ros_controllers.

Please review my code comment, there's a fourth option. It should be possible to match a repo's package names against the list of available test targets. This should do the trick better than the above options, what do you think?.

efernandez:
What's the problem with running only `catkin_make -j2 run_tests`?

I've just tried it with `ros_control`, `realtime_tools` and `ros_controllers` in the same workspace. And `echo $?` returns `0` after it. I'm testing with `indigo-devel`, except for `ros_controllers`, which is at `jade-devel` by default.

BTW, I had some trouble with this rostest:

https://github.com/ros-controls/ros_controllers/blob/jade-devel/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp#L327

but it seems the problem is that **rostests cannot be run in parallel**. When I run that test alone, it passes. This means we'd have to use `catkin_make run_tests` instead of `catkin_make -j2 run_tests`, unless gtests and rostests are separated somehow. But still, for the rostests (which take most of the time) we'd have to run the tests w/o parallelization, so I guess it's fine to remove the `-j2` to avoid false negatives.

efernandez:
FYI @bmagyar This is the actual issue from `realtime_tools`. I suggest closing the PR in `realtime_tools` and continue the discussion here.

bmagyar:
Ah jeez guys! Running the tests should be with -j1... I totally forgot but we fought the same thing with `gazebo_ros_pkgs` not so long ago [here](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/353)

bmagyar:
Fixing hydro builds is out of scope now, also our builds for newer distros are ok. Closing...

PR #82, opened 2014-03-06
Added support to JointTrajectoryController for an joint interfaces are not inherited from JointHandle.

Igorec:


Igorec:
It should be applied after http://github.com/ros-controls/ros_control/pull/147

adolfo-rt:
Thanks for submitting this PR!

PR #94, opened 2014-05-05
Add base_frame_id param (defaults to base_link)

po1:
The `nav_msgs/Odometry` message specifies the `child_frame_id` field, which was previously not set.

This PR creates a parameter to replace the previously hard-coded value of the `child_frame_id` of the published `tf` frame, and uses it in the `odom` message as well.

Opened against `hydro-devel` because this change is minor. It does change the default value of a field, though, which is noteworthy.

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
@@ -138,6 +138,9 @@ namespace diff_drive_controller{
     /// Timeout to consider cmd_vel commands old:
     double cmd_vel_timeout_;
 
+    /// Frame to use for the robot base:
+    std::string base_frame_id_;
+
     // speed limiters
     Commands last_cmd_;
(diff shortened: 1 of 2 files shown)


adolfo-rt:
This was already mentioned in the fork where this PR comes from, but it's good to re-state it here as well: Not hardcoding the `child_frame_id` to be `base_footprint` is a good thing, as `base_footprint` does not comply with [REP 105](http://www.ros.org/reps/rep-0105.html). The current behavior does comply with REP 105 by default (sets the frame name to `base_link`), but allows to be overridden.

The controller is rather new, so it's unlikely that people are using it extensively or relying on this default behavior. Still, if there are objections against the new defaults, please raise them, otherwise I'll accept the patch.

po1:
@adolfo-rt I think this should be merged into `indigo-devel` as well.

adolfo-rt:
> @adolfo-rt I think this should be merged into indigo-devel as well.

Done

po1:
Thanks!

PR #95, opened 2014-05-19
check for CATKIN_ENABLE_TESTING

bulwahn:
This patch follows the idea of pull request #89, but has been adjusted for the indigo-devel branch. Adjusting package.xml is not required because indigo only provides catkin versions that support this feature.

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -56,35 +56,37 @@ catkin_package(
   LIBRARIES ${PROJECT_NAME}
 )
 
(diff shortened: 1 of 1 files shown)


adolfo-rt:
+1

I'll wait for Travis to report back and merge. I'm catching up on #89 as well; Why did you not add the minimum catkin version here, because Indigo's catkin already meets the requirements?.

bulwahn:
yes, that was the rationale. However, I am not aware of a guideline for when to set the minimum version and when not.

adolfo-rt:
Thanks, merging

PR #97, opened 2014-05-27
joint_trajectory_controller: Critical targets declared before calling catkin_package

jbohren:
This should also be backported to hydro-devel.


jbohren:
Related to catkin/catkin_tools#51

adolfo-rt:
We're no longer [getting away with murder](https://github.com/catkin/catkin_tools/issues/51), snif! ;-)

PR #98, opened 2014-05-27
joint_trajectory_controller: making it so the URDF parameter is configurable

jbohren:
This is necessary when using this controller in simulation with multiple robots.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -113,7 +113,20 @@ std::vector<UrdfJointConstPtr> getUrdfJoints(const urdf::Model& urdf, const std:
     }
     else
     {
-      ROS_ERROR_STREAM("Could not find joint '" << joint_names[i] << "' in URDF model.");
+      std::ostringstream joints_oss;
+      std::vector<boost::shared_ptr<urdf::Link> > links;
+      urdf.getLinks(links);
+      for(std::vector<boost::shared_ptr<urdf::Link> >::const_iterator link_it = links.begin();
+          link_it != links.end();
+          ++link_it)
+      {
+        if((*link_it)->parent_joint) {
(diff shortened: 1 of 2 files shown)


adolfo-rt:
How's your namespace layout when using multiple robots?. I believe that the following layout should work with the current implementation:

```
/
|
|-- one_robot
|   |
|   |-- robot_description
|   |-- controller_manager
|
|-- other_robot
|   |
|   |-- robot_description
|   |-- controller_manager
```

For each robot, the `root_nh` would be `one_robot` and `other_robot`, respectively, so the URDF should be found.

I'm interested in this usecase, because I'm presently coding an rqt plugin for this controller, and I need both controller_manager and URDF info. It would be a big time saver if a sane convention could be assumed.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -113,7 +113,20 @@ std::vector<UrdfJointConstPtr> getUrdfJoints(const urdf::Model& urdf, const std:
...
+          link_it != links.end();
+          ++link_it)
+      {
+        if((*link_it)->parent_joint) {
+          joints_oss << " - \"" << (*link_it)->parent_joint->name << "\"\n";
+        }
+      }
+      ROS_ERROR_STREAM(
+          "Could not find joint '" << joint_names[i] << "' in URDF model.\n"
+          << "Available joints include: \n" << joints_oss.str());
```
adolfo-rt, line 129: This is an extra treat! ;-)

adolfo-rt:
Bump.
@jbohren, could you please comment on my [past comment](https://github.com/ros-controls/ros_controllers/pull/98#issuecomment-44458604)?. I'm interested in understanding your usecase.

adolfo-rt:
Bump again.

jbohren:
So this was specifically a problem with simulation, it should be resolved by using the `<robotNamespace>` tag in the `gazebo_ros_control` plugin, but for some reason that didn't work. If that should work, then maybe that's where the bug is.

adolfo-rt:
Long time no change.

PR #99, opened 2014-05-27
[joint_trajectory_controller] Declare targets after catkin_package.

adolfo-rt:


adolfo-rt:
Hydro backport of #97.

PR #104, opened 2014-06-20
Allow an initial value in position controllers

benwr:
We (Personal Robotics Lab at CMU) were attempting to use a position controller for a robot with starting positions other than 0. This change doesn't change the implementation of any existing setup; it only allows users to specify arbitrary starting values for controllers based on forward_command_controller.

### DIFF
--- a/forward_command_controller/include/forward_command_controller/forward_command_controller.h
+++ b/forward_command_controller/include/forward_command_controller/forward_command_controller.h
@@ -79,12 +79,20 @@ public:
       ROS_ERROR("No joint given (namespace: %s)", n.getNamespace().c_str());
       return false;
     }
+    
+    
+    initial_val_ = 0.0;
+    if (!n.getParam("initial"), initial_val_)
+    {
+      ROS_INFO("No initial value given (namespace: %s); using 0.0", n.getNamespace().c_str())
+    }
+    
     joint_ = hw->getHandle(joint_name);
     sub_command_ = n.subscribe<std_msgs::Float64>("command", 1, &ForwardCommandController::commandCB, this);
     return true;
   }
 
-  void starting(const ros::Time& time) {command_ = 0.0;}
+  void starting(const ros::Time& time) {command_ = initial_val_;}
   void update(const ros::Time& time, const ros::Duration& period) {joint_.setCommand(command_);}
 
   hardware_interface::JointHandle joint_;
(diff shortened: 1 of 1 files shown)


adolfo-rt:
I agree with you that having the command start always at zero is inappropriate for most cases. To better understand your usecase, are you using [position_controllers::JointPositionController](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/position_controllers/include/position_controllers/joint_position_controller.h#L59)?.

How about initializing the command (position in your case) with the joint's current value, instead of reading an ad-hoc value from the parameter server?. In the case of a position controller, this would mean holding the actual position.

One way of making this happen would be to keep [ForwardCommandController::starting()](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/forward_command_controller/include/forward_command_controller/forward_command_controller.h#L87) **virtual** (it's originally inherited from `controller_interface::Controller<T>`), and then reimplement an appropriate `starting()` behavior, such as

``` c++
void starting(const ros::Time& time)
{
  command_ = joint_.getPosition();
}
```

As a sidenote, what's your usecase?. if you're interested in either controlling multiple joints, interpolating at a higher frequency than the one at which you refresh your commands, or having a preemptible action interface, you should take a look at the [joint_trajectory_controller](https://github.com/ros-controls/ros_controllers/tree/indigo-devel/joint_trajectory_controller) package.

benwr:
Thanks for your response!

I think you're right that, in the case of position controllers, it makes more sense to initially replicate the current value by default. I had assumed it would be more general and less ad-hoc to allow arbitrary starting values for all ForwardCommandControllers. If that's not the case, then modifying the position controller directly in the way you suggest suits our needs.

I'd like to briefly mention, though, that doing so will change the behavior of existing controllers on initialization, and it may be that some people are currently counting on the controller to home the robot.

We are aware of the joint_trajectory_controller, thanks! I just wanted to bring this issue to your attention as well.

adolfo-rt:
> I think you're right that, in the case of position controllers, it makes more sense to initially replicate the current value by default. I had assumed it would be more general and less ad-hoc to allow arbitrary starting values for all ForwardCommandControllers.

I actually think this is less ad-hoc. Using a fixed value fetched from the param server means that depending on the initial configuration of the robot, it might (or not) abruptly jump to this value. I think that doing nothing if no commands are received, for all possible initial joint configurations, conforms well to the principle of least surprise.

> I'd like to briefly mention, though, that doing so will change the behavior of existing controllers on initialization, and it may be that some people are currently counting on the controller to home the robot.

Indigo is not out yet, so now would be a good time to push this change. I also think this controller is not widely used. For most practical purposes the `joint_trajectory_controller` is a better choice.

adolfo-rt:
This was addressed in #135, so closing.

PR #109, opened 2014-07-22
add velocity interface for joint_trajectory_controller

fmessmer:
...in separate feature branch

This feature has been provided by @davetcoleman

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
@@ -217,7 +217,7 @@ public:
     // Update PIDs
     for (unsigned int i = 0; i < n_joints; ++i)
     {
-      const double command = pids_[i]->computeCommand(state_error.position[i], desired_state.velocity[i], period);
+      const double command = pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
       (*joint_handles_ptr_)[i].setCommand(command);
     }
   }
--- a/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
@@ -45,10 +45,10 @@
 /**
  * \brief Helper class to simplify integrating the JointTrajectoryController with different hardware interfaces.
  *
- * The JointTrajectoryController outputs position, velocity, command triplets, while the more common hardware interfaces
- * it is supposed to work with accept either position or effort commands.
+ * The JointTrajectoryController outputs position and velocity command triplets, while the more common hardware interfaces
+ * it is supposed to work with accept position, velocity, or effort commands.
  *
- * Use one of the avaialble template specializations of this class (or create your own) to adapt the
+ * Use one of the available template specializations of this class (or create your own) to adapt the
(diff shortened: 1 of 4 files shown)


joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
```diff
@@ -45,10 +45,10 @@
 /**
  * \brief Helper class to simplify integrating the JointTrajectoryController with different hardware interfaces.
  *
- * The JointTrajectoryController outputs position, velocity, command triplets, while the more common hardware interfaces
- * it is supposed to work with accept either position or effort commands.
+ * The JointTrajectoryController outputs position and velocity command triplets, while the more common hardware interfaces
```
adolfo-rt: I made a mistake in the original wording, it should actually be 'position, velocity and acceleration command triplets'.

joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
```diff
@@ -104,13 +131,112 @@ class HardwareInterfaceAdapter<hardware_interface::PositionJointInterface, State
...
+
+/**
  * \brief Adapter for an effort-controlled hardware interface. Maps position and velocity errors to effort commands
  * through a position PID loop.
  *
  * The following is an example configuration of a controller that uses this adapter. Notice the \p gains entry:
  * \code
  * head_controller:
- *   type: "position_controllers/JointTrajectoryController"
+ *   type: "effort_controllers/JointTrajectoryController"
```
adolfo-rt, line 225: Good catch here!

joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
```diff
@@ -71,6 +71,33 @@ class HardwareInterfaceAdapter
 
 /**
  * \brief Adapter for a position-controlled hardware interface. Forwards desired positions as commands.
+ *
+ * The following is an example configuration of a controller that uses this adapter. Notice the \p gains entry:
```
adolfo-rt: Please remove the 'Notice the \p gains entry' phrase.

If you want to add an example usage, could you please mimic the example used for the effort adapter as closely as possible?, so it becomes straightforward to spot the differences. With this I mean preserving the joint count and names, as well as constraint specification. The only things that should change are the controller type (namespace prefix) and the PID gains, which are not needed in for position joints.

adolfo-rt:
I made a few inline documentation comments. Otherwise it looks OK.

Can you confirm that the new plugin works as expected with velocity joints, and whether it has been tested in simulation and/or hardware?.

Thanks!

adolfo-rt:
Something else. New features go to `indigo-devel`, so you should rebase your branch and open a new PR against `indigo-devel`. Expect some conflicts at the CMake level, but not much more.

fmessmer:
The example configurations have been harmonized.
I use this controller for our robot in Gazebo and - as far as I can tell - it seems to work fine.
I haven't tested it with real hardware...

adolfo-rt:
The patch looks good. If you can make it target `indigo-devel` I'll merge it.

fmessmer:
#116  rebases this feature to the latest indigo-devel branch
In case this feature does not go into hydro-devel, please close this PR!

fmessmer:
As #116 is merged, I'll close this one!

PR #113, opened 2014-07-27
addressing test dependencies with -DCATKIN_ENABLE_TESTING=0

bulwahn:


gripper_action_controller/CMakeLists.txt
```diff
@@ -17,7 +17,6 @@ find_package(catkin
       realtime_tools
       control_msgs
       trajectory_msgs
-      rostest
```
adolfo-rt, line 20: You might want to also remove the `<build_depends>` element from the `package.xml`.

adolfo-rt:
See one inline comment, otherwise +1

gripper_action_controller/CMakeLists.txt
bulwahn, line 20: You are right, and I actually did remove the element from the package.xml in just that commit.

bulwahn:
See inline comment: the issue mentioned was already addressed.

adolfo-rt:
All good then. Merging.

Issue #118, opened 2014-09-02
odom tf publishing should be optional in diff_drive_controller


mikepurvis:
See the setup tutorial for robot_localization:

http://wiki.ros.org/robot_localization/Tutorials/Preparing%20Your%20Sensor%20Data#Odometry

[REP-105](http://www.ros.org/reps/rep-0105.html#odom) states that the odom -> base_link tf should come from "an odometry source". To my reading this says it could be diff_drive_controller _or_ a higher level component which also takes in IMU data.

I suggest a default-true parameter, `~enable_odom_tf`. If that's agreeable, I can implement it.

@bmagyar

bmagyar:
I see your point and agree. Adding the default-true parameter yields no changes for our current configuration files. 
There is however one question that I forgot to ask, that is the distro this should go, but I think it's also useful in hydro. 

@adolfo-rt I'm planning to add this to indigo-devel too, once this is merged.

mikepurvis:
I'm working in Indigo, so yes, it'd be great to have this merged there too.

adolfo-rt:
Correct, target `indigo-devel` for new features. `hydro-devel` is in bugfix-only mode.

bmagyar:
Sure, I'll update the PR.

PR #119, opened 2014-09-03
Add default-true parameter "~enable_odom_tf"

bmagyar:
https://github.com/ros-controls/ros_controllers/issues/118


mikepurvis:
+1

PR #127, opened 2014-09-22
added joint names filter parameter to joint state controller

mathias-luedtke:
This patch introduces a "joints" parameter that overwrites the joint names selection of the joint state controller.
It enables to run multiple joint state controllers with different publish rates and can be useful to harmonize bevahiour of simulation and real robots.
Defaults to publishing all joint names if no parameter is given.

### DIFF
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -39,8 +39,12 @@ namespace joint_state_controller
 
   bool JointStateController::init(hardware_interface::JointStateInterface* hw, ros::NodeHandle &root_nh, ros::NodeHandle& controller_nh)
   {
-    // get all joint names from the hardware interface
-    const std::vector<std::string>& joint_names = hw->getNames();
+    // try to get joint names from parameter server
+    std::vector<std::string> joint_names_param;
+    bool restricted = controller_nh.getParam("joints", joint_names_param) && !joint_names_param.empty();
+
+    // get all joint names from the hardware interface otherwise
+    const std::vector<std::string>& joint_names = restricted ? joint_names_param : hw->getNames();
     for (unsigned i=0; i<joint_names.size(); i++)
       ROS_DEBUG("Got joint %s", joint_names[i].c_str());
 
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -39,28 +39,51 @@ namespace joint_state_controller
 
   bool JointStateController::init(hardware_interface::JointStateInterface* hw, ros::NodeHandle &root_nh, ros::NodeHandle& controller_nh)
   {
-    // try to get joint names from parameter server
-    std::vector<std::string> joint_names_param;
-    bool restricted = controller_nh.getParam("joints", joint_names_param) && !joint_names_param.empty();
-
-    // get all joint names from the hardware interface otherwise
-    const std::vector<std::string>& joint_names = restricted ? joint_names_param : hw->getNames();
-    for (unsigned i=0; i<joint_names.size(); i++)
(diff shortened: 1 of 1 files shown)


joint_state_controller/src/joint_state_controller.cpp
```diff
@@ -39,8 +39,12 @@ namespace joint_state_controller
 
   bool JointStateController::init(hardware_interface::JointStateInterface* hw, ros::NodeHandle &root_nh, ros::NodeHandle& controller_nh)
   {
-    // get all joint names from the hardware interface
-    const std::vector<std::string>& joint_names = hw->getNames();
+    // try to get joint names from parameter server
+    std::vector<std::string> joint_names_param;
+    bool restricted = controller_nh.getParam("joints", joint_names_param) && !joint_names_param.empty();
```
adolfo-rt: Please use a more descriptive name for `restricted`. Eg. `use_custom_joints`. I you can think of something briefer yet descriptive of the purpose, even better.

joint_state_controller/src/joint_state_controller.cpp
```diff
@@ -39,8 +39,12 @@ namespace joint_state_controller
...
+
+    // get all joint names from the hardware interface otherwise
+    const std::vector<std::string>& joint_names = restricted ? joint_names_param : hw->getNames();
```
adolfo-rt: You should validate the contents of `joint_names_param` before committing to use them. You should at least check that the joints are exposed by `hw->getNames()`, and that there are no duplicate entries. Proper error (the former) and warning (the latter) log statements should be produced to provide user feedback. Also, `init` should fail when an error is encountered.

adolfo-rt:
Apart from inline comments, the new behavior should be documented to become useful. 

Currently, since the configuration is straightforward, there exists a [default config file](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_state_controller/joint_state_controller.yaml) which work for any robot.

One possibility is to add contents to the ROS wiki entry:
http://wiki.ros.org/joint_state_controller

The existing wiki macros for documenting ROS APIs can be used. An example documented controller can be found here
http://wiki.ros.org/joint_trajectory_controller

The `joint_state_controller` would have a much simpler page, but you get the idea.

mathias-luedtke:
I have implemented the requested behaviour, but I have no time to document it from scratch.
Anyway the additional parameter is not needed and the default config will work as expected.
We can document this special feature after it was merged/released.

adolfo-rt:
> I have implemented the requested behaviour, but I have no time to document it from scratch.

When you do have the time, it should not take more than half an hour.

> Anyway the additional parameter is not needed and the default config will work as expected.

Boy Scouts [have a rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule), which applies well to open-source projects: _"Always leave the campground cleaner than you found it."_

By adding a new interesting feature and not documenting it, you are making the project better for yourself, but not for others. Pretty much only the readers of this PR will be aware of the feature you are proposing, which is far less people than those that use the `joint_state_controller`, arguably the most widely used controller in all of `ros_controllers`.

> We can document this special feature after it was merged/released.

Should I read this as: _I will_ document this special feature after it is merged/released.? ;-)

One last thing. Considering that this this is likely the most used controller in `ros_controllers`, my concern for its stability is very high, so a unit test would also fast-track the merging of this pull request.

Thanks for your understanding. Making an open source project great not only takes writing great code, it also takes writing great documentation and great tests, and we want this project to be a good one.

mathias-luedtke:
Sorry, I have already spent too much time on this just-nice-to-have feature..

adolfo-rt:
That's fine. This PR can be picked up again if and when someone needs it enough to implement the feature more fully. Feel free to close it if you don't expect to work on it anymore.

PR #128, opened 2014-09-22
add forward_chain_command_controllers

fmessmer:
This PR introduces a new template class **forward_chain_command_controller**, with which it is possible to send commands to a kinematic chain consisting of several joints. Additionally three new controllers are added which use this class:
- effort_controllers/chain_effort_controllers
- velocity_controllers/chain_velocity_controllers
- position_controllers/chain_position_controllers

The controllers subscribe to a topic "command" of type std_msgs/Float64MultiArray and use the float64[] data field as command values interpreting the values as position, velocity and efffort depending on the control domain respectively.

### DIFF
--- a/effort_controllers/CMakeLists.txt
+++ b/effort_controllers/CMakeLists.txt
@@ -30,6 +30,7 @@ add_library(${PROJECT_NAME}
   src/joint_effort_controller.cpp
   src/joint_velocity_controller.cpp
   src/joint_position_controller.cpp
+  src/joint_group_effort_controller.cpp
 )
 
 target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
--- a/effort_controllers/effort_controllers_plugins.xml
+++ b/effort_controllers/effort_controllers_plugins.xml
@@ -18,4 +18,10 @@
     </description>
   </class>
 
+  <class name="effort_controllers/JointGroupEffortController" type="effort_controllers::JointGroupEffortController" base_class_type="controller_interface::ControllerBase">
+    <description>
+      The JointGroupEffortController tracks effort commands for a set of joints. It expects a EffortJointInterface type of hardware interface.
+    </description>
+  </class>
+
 </library>
--- a/effort_controllers/include/effort_controllers/joint_group_effort_controller.h
+++ b/effort_controllers/include/effort_controllers/joint_group_effort_controller.h
@@ -1,0 +1,63 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2008, Willow Garage, Inc.
+ *  Copyright (c) 2012, hiDOF, Inc.
+ *  Copyright (c) 2013, PAL Robotics, S.L.
+ *  Copyright (c) 2014, Fraunhofer IPA
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the Willow Garage nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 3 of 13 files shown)


fmessmer:
The message type is open for discussion!
I could also think of a new message type in control_msgs, e.g. ChainCommand.msg, which has the following definition:

```
string[] joint_names
float64[] command
```

This would allow for correctly mapping the commands to the handles of the joints...

effort_controllers/include/effort_controllers/chain_effort_controller.h
```diff
@@ -0,0 +1,63 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2008, Willow Garage, Inc.
+ *  Copyright (c) 2012, hiDOF, Inc.
+ *  Copyright (c) 2013, PAL Robotics, S.L.
+ *  Copyright (c) 2014, Fraunhofer IPA
```
bmagyar: Since this is a new file, there's no need to start it from 2008 with Willow :)

adolfo-rt:
This seems to be a generalization of the existing single-joint controllers to multiple joints. I don't see anywhere the restriction that the group must be a kinematic chain, or did I miss something?.

forward_command_controller/include/forward_command_controller/forward_chain_command_controller.h
```diff
@@ -0,0 +1,154 @@
...
+#include <std_msgs/Float64MultiArray.h>
+
+
+namespace forward_command_controller
+{
+
+namespace internal
+{
+
+std::vector<std::string> getStrings(const ros::NodeHandle& nh, const std::string& param_name)
```
adolfo-rt: roscpp now has a [convenience method](http://docs.ros.org/hydro/api/roscpp/html/classros_1_1NodeHandle.html#afe5b2bed420cb24e21797651a273f3ad) that does this for you without having to implement it again. You might have taken this from an existing controller implementation that was written back in the Fuerte days.

forward_command_controller/include/forward_command_controller/forward_chain_command_controller.h
```diff
@@ -0,0 +1,154 @@
...
+
+  bool init(T* hw, ros::NodeHandle &n)
+  {
+    // List of controlled joints
+    joint_names_ = internal::getStrings(n, "joints");
+    if (joint_names_.empty()) {return false;}
+    n_joints_ = joint_names_.size();
+    
+    for(unsigned int i=0; i<n_joints_; i++)
+    {  joints_.push_back(hw->getHandle(joint_names_[i]));  }
```
adolfo-rt: `getHandle()` will [throw an exception](https://github.com/ros-controls/ros_control/blob/indigo-devel/hardware_interface/include/hardware_interface/internal/hardware_resource_manager.h#L96) if the name does not exist. You should either first check for the existence of the names, or catch the exception and handle the error.

adolfo-rt:
I think quite a few persons might find this addition useful :)

effort_controllers/include/effort_controllers/chain_effort_controller.h
fmessmer: As I copied from `ForwardCommandController`, I'm fine giving those guys some credit...

fmessmer:
@adolfo-rt 
I think I covered your suggestions in the additional commits.

Yes, it not necessarily needs to be a chain. In fact, it can be used to control a list of joints.
Would you like "Chain" to be replaced by "MultiJoint" or something?

Any comments on the message type?
Should we define a new message for this in `control_msgs` or is it ok to use `Float64MultiArray`?

forward_command_controller/include/forward_command_controller/forward_chain_command_controller.h
```diff
@@ -0,0 +1,135 @@
...
+    for(unsigned int i=0; i<n_joints_; i++)
+    {
+      try
+      {
+        joints_.push_back(hw->getHandle(joint_names_[i]));  
+      }
+      catch (const hardware_interface::HardwareInterfaceException& e)
+      {
+        ROS_ERROR_STREAM("Exception thrown: " << e.what());
+        ROS_ERROR_STREAM("Not able to getHandle for '" << joint_names_[i]);
```
adolfo-rt: This statement is redundant, `e.what()` above will output ([source](https://github.com/ros-controls/ros_control/blob/indigo-devel/hardware_interface/include/hardware_interface/internal/resource_manager.h#L109)):

```
Could not find resource 'name' in 'interface_type'.
```

If you want to be specific that the missing resource is a joint, a variant of this statement is implemented [here](https://github.com/ipa-fxm/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L299), which yields

```
Could not find joint 'name' in 'interface_type'.
```

adolfo-rt:
> Yes, it not necessarily needs to be a chain. In fact, it can be used to control a list of joints.
> Would you like "Chain" to be replaced by "MultiJoint" or something?

How about joint group?. I believe that is the term used in MoveIt! to identify a set of joints. We could borrow it here.

> Any comments on the message type?
> Should we define a new message for this in control_msgs or is it ok to use Float64MultiArray?

Hmmm. Well, on one hand you're adding to an existing package, which uses plain `Float64` for single joint commands, so `Float64MultiArray` would be consistent with that. Also, `control_msgs` is a package used by a much larger community than that of `ros_control`, so adding a new type here would trigger a pretty large downstream rebuild.

On the other hand, it is a best practice to add some semantics to the message by means of a new type. Input from other readers of this discussion would be valued at this point.

fmessmer:
The controllers are now called JointGroup[Position|Velocity|Effort]Controller

Shall we wait for more input on the message type or shall I squash to be ready for merging?

adolfo-rt:
> The controllers are now called JointGroup[Position|Velocity|Effort]Controller

Looks better now. There's more parallelism with existing single-joint controllers: `joint_position_controller` vs `joint_group_position_controller`.

> Shall we wait for more input on the message type or shall I squash to be ready for merging?

If you're in no hurry, we can give it another day or two.

fmessmer:
Ping? Any comments on this PR?

adolfo-rt:
Sorry for the latency. Just came out to the surface to breathe ;-).

Since no further feedback has been received on the message type, and considering that the original classes this PR generalizes use `Float64`s, I'd say to give it a go.

A commit squash would be appreciated.

fmessmer:
Squashed!

adolfo-rt:
Thanks for putting this together.

PR #135, opened 2014-10-10
use current position when starting (forward_command) position_controllers

fmessmer:
effort and velocity (forward_command) controllers still use 0.0

The template class now only contains the declaration for `starting()`. The implementation moved to the respective controllers.

This is related to #104

### DIFF
--- a/position_controllers/src/joint_group_position_controller.cpp
+++ b/position_controllers/src/joint_group_position_controller.cpp
@@ -38,4 +38,16 @@
 #include <position_controllers/joint_group_position_controller.h>
 #include <pluginlib/class_list_macros.h>
 
+template <class T>
+void forward_command_controller::ForwardJointGroupCommandController<T>::starting(const ros::Time& time)
+{
+  // Start controller with current joint positions
+  commands_.resize(n_joints_, 0.0);
+  for(unsigned int i=0; i<joints_.size(); i++)
+  {
+    commands_[i]=joints_[i].getPosition();
+  }
+}
+
+
 PLUGINLIB_EXPORT_CLASS(position_controllers::JointGroupPositionController,controller_interface::ControllerBase)
--- a/position_controllers/src/joint_position_controller.cpp
+++ b/position_controllers/src/joint_position_controller.cpp
@@ -37,4 +37,11 @@
 #include <position_controllers/joint_position_controller.h>
 #include <pluginlib/class_list_macros.h>
 
+template <class T>
+void forward_command_controller::ForwardCommandController<T>::starting(const ros::Time& time)
+{
+  // Start controller with current joint position
+  command_ = joint_.getPosition();
+}
+
 PLUGINLIB_EXPORT_CLASS(position_controllers::JointPositionController,controller_interface::ControllerBase)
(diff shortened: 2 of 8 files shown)


fmessmer:
I'll try to find a solution for initializing the other position controllers with the current position as well.
I.e. `effort_controllers::JointPositionController` and `velocity_controllers::JointPositionController`

UPDATE: Ooops, just saw that those already use `joint_.getPosition();` in their `starting()`!

Also:
- Do we want a toggle that allows to chose whether positon controllers are initialized with current or 0.0?
- Do we want a parameter with which it's able to specify the initial position (as suggested in #104)?

fmessmer:
Ping? Any comments on this?
@benwr would this solve your issue #104?

position_controllers/src/joint_group_position_controller.cpp
```diff
@@ -38,4 +38,22 @@
 #include <position_controllers/joint_group_position_controller.h>
 #include <pluginlib/class_list_macros.h>
 
+template <class T>
+void forward_command_controller::ForwardJointGroupCommandController<T>::starting(const ros::Time& time)
+{
+  // Start controller with current joint positions
+  try{
```
adolfo-rt: Can this actually fail ,given that `init` should return false if the requested joints don't exist?.

I say this because `starting()` is a real-time safe method, and exception handling is not a real-time safe mechanism, in the sense that stack unwinding takes a non-deterministic amount of time.

position_controllers/src/joint_position_controller.cpp
```diff
@@ -37,4 +37,17 @@
 #include <position_controllers/joint_position_controller.h>
 #include <pluginlib/class_list_macros.h>
 
+template <class T>
+void forward_command_controller::ForwardCommandController<T>::starting(const ros::Time& time)
+{
+  // Start controller with current joint position
+  try{
```
adolfo-rt: Ditto exception handling

adolfo-rt:
>    Do we want a toggle that allows to chose whether positon controllers are initialized with current or 0.0?
>   Do we want a parameter with which it's able to specify the initial position (as suggested in #104)?

I'm inclined to say no to both, for simplicity's sake. Semantic zero adheres to the principle of least surprise. Still, I have no strong argument against having those parameters, if the need exists.

Also, take into account that upon starting, as soon as the first control command message is processed, it will be taken into account, so very few control cycles will take place in semantic zero mode, if a command publisher is already in place.

position_controllers/src/joint_group_position_controller.cpp
fmessmer: I guess you are right.
In case the `joint_` cannot return `getPosition()` it should have failed earlier already!
I removed the `try-catch` part

position_controllers/src/joint_position_controller.cpp
fmessmer: dito: removed!

fmessmer:
Alright, for our use-case we also do not need these parameters...
If need arises, the parameters could be added in another follow-up PR...

Shall I squash then?

adolfo-rt:
> Shall I squash then?

Yes, please.

I really appreciate these fixes that make the project more consistent.

fmessmer:
Squashed....should be good to go then...

PR #136, opened 2014-10-11
Diff drive open loop odom

bmagyar:
This PR is an update from the hydro-devel branch of ros_controllers from PAL Robotics.

> The purpose of this PR is to support open loop odometry when there are no encoders.
> It basically adds:
> - open_loop param
> - extends Odometry class to update using the last linear and angular velocities received by the controller, i.e. set the odometry twist directly with then and integrate the position (x, y)

Related PR:
https://github.com/pal-robotics/ros_controllers/pull/39

### DIFF
--- a/diff_drive_controller/CMakeLists.txt
+++ b/diff_drive_controller/CMakeLists.txt
@@ -61,5 +61,6 @@ if (CATKIN_ENABLE_TESTING)
     test/diff_drive_odom_tf.test
     test/diff_drive_odom_tf_test.cpp)
   target_link_libraries(diff_drive_odom_tf_test ${catkin_LIBRARIES})
+  add_rostest(test/diff_drive_open_loop.test)
 
 endif()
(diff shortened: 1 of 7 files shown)


adolfo-rt:
I only have a naming question. Is it clear by just stating `open_loop` that we are referring  to the odometry computation? (as opposed to a name like `open_loop_odometry`).

efernandez:
There's no other thing that can be open loop, apart from the odometry, so IMHO it's better to keep the name as `open_loop`.
Also, you can think of it as open loop control, rather than whether odometry is used or not.

adolfo-rt:
If you think the meaning is clear enough, then I give my +1. Once this PR is merged, don't forget to update the controller's wiki page: http://wiki.ros.org/diff_drive_controller

bmagyar:
+1

efernandez:
@adolfo-rt Ok, we'll take care of updating the wiki.
Thanks for your comments!

adolfo-rt:
The whole sequence of PRs related to the diff-drive controller seem to be in place (#136-#138). Starting the merge process...

Thanks for your continued efforts.

PR #137, opened 2014-10-11
diff drive remove angles

bmagyar:
This PR is an update from the hydro-devel branch of ros_controllers from PAL Robotics, and it depends on https://github.com/ros-controls/ros_controllers/pull/136.

It eliminates the dependency on `angles`.

Related PR:
https://github.com/pal-robotics/ros_controllers/pull/40


adolfo-rt:
+1

Issue #147, opened 2014-10-27
Making room for mecanum_drive_controller


arennuit:
Hi guys,

I am currently developping a mecanum_drive_controller which (as its name states) will take a twist as input and generate desired wheel velocities. From what I have seen this class will be rather close to diff_drive_controller. So I was thinking of factoring the common code out into a drive_controller class (from which diff_drive_controller and mecanum_drive_controller would derive). It does not look like a big issue, but I was wondering how you would welcome this proposition... If you consider its worth the effort, then I'll make a proposition in a fork of mine, otherwise no worries I'll do a completely separate package in a repo of mine.

Just let me know what you think.

Thanks,

Antoine.

adolfo-rt:
How much of `diff_drive_controller` do you think could be factored-out and reused?.

arennuit:
On a very rough estimate I would say 50%... and spread like this:
- diff_drive_controller: 50%
- odometry: 30%
- speed_limiter: 100%

bmagyar:
I would advise against further generalization of the existing diff_drive_controller code since it would introduce  more complexity to something that is otherwise extremely simple. 

Luckily though in C++ it is possible to inherit code from a parent class. I'd suggest an approach where you introduce a dependency to diff_drive_controller, inherit code from there and override the necessary parts in your controller. 

Feel free to add some virtuals to whereever it's necessary in the diff_drive_controller code.

adolfo-rt:
...or see if some controller components can be factored out in separate classes / functions, so _compose_  instead of _extend_.

efernandez:
@arennuit What do you mean a mecanum driver controller? Do you mean an omnidirectional base using mecanum wheels, so it can move sideways?

The `diff_drive_controller` simply computes the odometry and transforms robot-space velocities into wheel-space velocities for differential and skid steer platforms. IMO an omnidirectional (or even an Ackermann) platform could have a similar structure to the diff_drive_controller, but at the end the equations differ, so there's no so much you can use from the diff_drive_controller (apart from auto-imposing you a set of methods, classes and attributes). I think it's better that you implement it separately and then we can see whether it makes sense or not to factor something or create a class hierarchy.

arennuit:
@bmagyar, @adolfo-rt : no worries I do not plan to extend the diff_drive_controller but only to factor the common code out (as Adolfo is mentioning)

@efernandez: yes I mean a controller which computes mecanum wheels velocities given a desired mobile base twist and also computes the odometry

[...]

PR #149, opened 2014-10-29
Mecanum drive

arennuit:
This is a proposition to integrate a mecanum_drive_controller in ros_controllers. This controller computes the wheels velocities given a desired twist and computes the odometry given the observed wheels velocity. 

This PR relates to issue #147 :  Making room for mecanum_drive_controller.

There is no unit test yet.

Also it is not clear yet which twists are expected as input (desired twist) and output (odometry twist). This question is raised in issue #147 and in ROS answers: http://answers.ros.org/question/196233/geometry_msgstwist-references-navigation-stack-ros_control/.


arennuit:
This code still needs further testing, unit tests and review. I will try to do this when pressure cools down a bit :)

bmagyar:
You can find plenty of tests in diff_drive_controller/test, feel free to refactor them. I'm sure most of them will come very useful to you! 

((It might make sense after you have them to run a coverage tool to see which parts of the code are not being excersized during the tests.))

arennuit:
@bmagyar You are right, this is exactely how I plan to proceed. Only need to get out of my rush to consolidate all this. Thanks ;)

adolfo-rt:
Please let us know when the code stops being in flux and is in shape for a review. Adding tests would be a big plus.

arennuit:
As for the unit tests I agree with you and this is my intention. As for the review, code has been rather stable for a week or two now, but I am still undergoing a few hand tests before making things officially for review ;)

mikepurvis:
I'm interested in this, and may be able to contribute to tests and verification in the new year.

Regarding simulation, how do you expect this to work with Gazebo? Is a simulation of mecanum geometry sufficient to get accurate behaviour, or is it necessary to have a separate plugin which handles the properties of the wheels themselves?

Related: http://answers.gazebosim.org/question/4078/how-to-simulate-a-mecanum-wheel-in-gazebo/

arennuit:
Glad to see there is interest about this package ;)

Regarding the simulation, I recommend not to use gazebo for mecanum wheels. I spent a lot of time (I mean A LOT ;)) only to realize gazebo is not quite ready yet for the scenario.

Basically you have 2 choices for simulating mecanum wheels:
- you simulate the real mecanics with the many rollers
  - this leads to complex simulations, hard to get stable
- you emulate them using this trick: http://www.forum.coppeliarobotics.com/viewtopic.php?f=9&t=390
  - unfortunately you cannot do this in gazebo, see here why http://answers.gazebosim.org/question/7126/change-frame-of-a-linkjoint-without-moving-the/

The solution shown in the link you sent does not simulate the mecanum wheels at all: it is only a kinematic trick allowing to get good looking visuals, but unfortunately you do not get a proper mecanum wheels behavior.

Because of this we decided to use vrep rather than gazebo, and... well, this turned out to be an excellent decision. Vrep really rocks and does feel professional. I have coded a ros_control / vrep back-end example (equivalent to ros-indigo-gazebo-ros-control, but for vrep rather than gazebo), and it is available here: https://github.com/arennuit/vrep_ros_control_example.

Regarding your contribution, let me know when you feel ready, I may be able to help ;)

mikepurvis:
@arennuit We're now using your `mecanum_drive_controller` on our prototype [Ridgeback](http://www.clearpathrobotics.com/ridgeback/) units, and we're very happy with it. One thing to note is that it's definitely hard-coded for the rectangular case, and I wonder if the name of the controller should reflect that (ie, that it is not suitable for a platform with six mecanum wheels, or three, or whatever).

Also, have you coalesced on an ideal approach to simulating mecanum wheels in Gazebo? We've not really been satisfied with any of the existing options.

arennuit:
Hello Mike,

And thanks for your feedback. You are right, the fixed number of wheels the controller can handle is not clear from its name (and you certainly have to dig into the doc and code to know about it). I need to come back to the controller's code in a few weeks so this observation may lead to **a bit of renaming** ;)

Now as for simulating mecanum wheels inside gazebo, there are several ways to do it: 1) **simulate the real mechanics**: this has turned out to be very resource intensive and is not very stability, or 2) **use a trick** such as [described here](http://www.forum.coppeliarobotics.com/viewtopic.php?f=9&t=390#p1380): this is the method I decided to go for. I have found using 2) within gazebo required code modifications of gazebo really deeply rooted and too time consuming to meet our R&D constraints. The restriction inside gazebo is described [in this post](http://answers.gazebosim.org/question/7126/change-frame-of-a-linkjoint-without-moving-the/) and reported [in this issue](https://bitbucket.org/osrf/gazebo/issue/1286/change-frame-of-a-link-joint-without). The workaround I have found is to use vrep (after using it for a few months now, I can say this simulation software highly deserves attention, it really rocks) and the trick mentioned in 2). I have shared some code bridging vrep and ros_controllers [in github here](https://github.com/arennuit/vrep_ros_control_example) (this allows me to use the **mecanum_drive_controller** in a clean and straightforward way).
Now maybe you allow me to take advantage of your message to ask you another question. The mecanum_drive_controller takes a cartesian velocity as input (i.e. the command input in the mecanum_drive_controller is a twist). Now what we want to do is to have our mobile robot follow a cartesian path. We can sample our path in such a way that each sample contains the desired cartesian position, twist (and even acceleration but we do not use the acceleration) of the mobile robot - the desired cartesian path is obviously on a plane. In order to transform the desired path into a velocity command (i.e. the input of the **mecanum_drive_controller**) we use the following command law (this is C++ code):

_twist_command[i] = kf_[i] \* desiredVelocity_inBody[i] + kp_[i] \* errorPosition_inBody[i] + kv_[i] \* errorVelocity_inBody[i];_

that is the twist command is made up out of the desired twist + feedback on the position and velocity errors.

This command law turns out not to be as stable as I had hoped. Hence my question: what method do you use in order to have your robot follow a path precisely? Do you use a command law such as mine, or do you have another method?

PS: sorry GitHub does not seem to regard my markdown instructions...

mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -0,0 +1,516 @@
...
+  if (!getWheelRadius(model, model->getLink(wheel0_urdfJoint->child_link_name), wheel0_radius) ||
+      !getWheelRadius(model, model->getLink(wheel1_urdfJoint->child_link_name), wheel1_radius) ||
+      !getWheelRadius(model, model->getLink(wheel2_urdfJoint->child_link_name), wheel2_radius) ||
+      !getWheelRadius(model, model->getLink(wheel3_urdfJoint->child_link_name), wheel3_radius))
+  {
+    ROS_ERROR_STREAM_NAMED(name_, "Couldn't retrieve wheels' radius");
+    return false;
+  }
+
+  if (abs(wheel0_radius - wheel1_radius) > 1e-3 ||
```
efernandez: std::abs, I think with abs only it's truncating to int type

efernandez:
@arennuit Could you provide the reference for the kinematics equations you're using for the mecanum drive?
In principle, it seems to be based on [this presentation](http://www.chiefdelphi.com/media/papers/download/2614) by Mackenzie; although some signs differ on [your implementation](https://github.com/ros-controls/ros_controllers/pull/149/files#diff-e5745efd21bf8a422a81064b2b4a27efR262), the kinematic parameter `wheels_k_` seems to be 4 times the `a + b` on Mackenzie's model, and it seems you don't follow the same wheels enumeration as him, as in [this equations](https://github.com/ros-controls/ros_controllers/pull/149/files#diff-8e769a063e0a88b35f747c4400afd0faR99). According to it, the name mecanum drive looks appropriate to me.
Anyway, you're applying more equations... (at least in the odometry integration) so I guess you're using another reference.

mecanum_drive_controller/src/odometry.cpp
```diff
@@ -0,0 +1,147 @@
...
+  // Reset timestamp:
+  timestamp_ = time;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+bool Odometry::update(double wheel0_vel, double wheel1_vel, double wheel2_vel, double wheel3_vel, const ros::Time &time)
+{
+  /// We cannot estimate the speed with very small time intervals:
+  const double dt = (time - timestamp_).toSec();
+  if (dt < 0.0001)
```
efernandez, line 45: Since now the wheel velocities are actually velocities, you don't have to divide by `dt`, so you can get rid of this.

mecanum_drive_controller/include/mecanum_drive_controller/odometry.h
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * \brief Initialize the odometry
+   * \param time Current time
+   */
+  void init(const ros::Time &time);
+
+  /**
+   * \brief Updates the odometry class with latest wheels position
+   * \param wheel0_vel  Wheel velocity [rad]
```
efernandez: Now this is rad/s

mecanum_drive_controller/include/mecanum_drive_controller/odometry.h
```diff
@@ -0,0 +1,204 @@
...
+   * \param wheel2_vel  Wheel velocity [rad]
+   * \param wheel3_vel  Wheel velocity [rad]
+   * \param time      Current time
+   * \return true if the odometry is actually updated
+   */
+  bool update(double wheel0_vel, double wheel1_vel, double wheel2_vel, double wheel3_vel, const ros::Time &time);
+
+  /**
+   * \brief Updates the odometry class with latest velocity command
+   * \param linearX  Linear velocity [m/s]
```
efernandez: linearY missed

mecanum_drive_controller/include/mecanum_drive_controller/odometry.h
```diff
@@ -0,0 +1,204 @@
...
+  /**
+   * \brief Updates the odometry class with latest wheels position
+   * \param wheel0_vel  Wheel velocity [rad]
+   * \param wheel1_vel  Wheel velocity [rad]
+   * \param wheel2_vel  Wheel velocity [rad]
+   * \param wheel3_vel  Wheel velocity [rad]
+   * \param time      Current time
+   * \return true if the odometry is actually updated
+   */
+  bool update(double wheel0_vel, double wheel1_vel, double wheel2_vel, double wheel3_vel, const ros::Time &time);
```
efernandez, line 43: Consider using left/right_front/back_wheel_vel instead of only numbers; at least indicate where is each wheel supposed to be on the doxygen comments.

mecanum_drive_controller/src/odometry.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return false; // Interval too small to integrate with
+
+  timestamp_ = time;
+
+  /// Compute forward kinematics (i.e. compute mobile robot's body twist out of its wheels velocities):
+  /// NOTE: we use the IK of the mecanum wheels which we invert using a pseudo-inverse.
+  /// NOTE: in the diff drive the velocity is filtered out, but we prefer to return it raw and let the user perform
```
efernandez, line 53: Using a window of 1 for the rolling mean doesn't work?

mecanum_drive_controller/src/odometry.cpp
```diff
@@ -0,0 +1,147 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+void Odometry::setWheelsParams(double wheels_k, double wheels_radius)
+{
+  wheels_k_ = wheels_k;
+
+  wheels_radius_ = wheels_radius;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+void Odometry::integrateExact(double linearX, double linearY, double angular)
```
efernandez: For an holonomic robot, the rigid motion equations are applied directly, in contrast with the differential drive. That is:

(x, y) += R(yaw) \* (v_x, v_y) \* dt
yaw += w \* dt

where (v_x, v_y) and omega are the linear and angular velocities in the BODY frame.

If you want, you can rename the function to just integrate and get rid of all the integrate_fun_ boost::function stuff.

mecanum_drive_controller/test/diff_drive_test.cpp
```diff
@@ -0,0 +1,151 @@
+///////////////////////////////////////////////////////////////////////////////
```
efernandez, line 1: There is certainly a lot of work to migrate all these tests for the mecanum drive.
I'll suggest to provide some basic tests starting from this one.

Anyway, all the tests are rostests, which depend on gazebo. So, given the issues with simulating mecanum or omni wheels on gazebo, that looks like a big blocking issue, right?

arennuit: Hello @efernandez, I am away from the office during the next 3 days. But I'll come back to you on Monday with the equations and other answers.

arennuit:
Hello @efernandez, thanks for your feedback (sorry I have only been able to come back to the mecanum controller now. The equations I use come from [here](http://www.chiefdelphi.com/media/papers/download/2722). The reference frame is indeed different from the one used Mackenzie (which explains the signs difference).
Now I am not sure I understand why you say my _wheels_k_ is 4 times Mackenzie's _a+b_. _wheels_k_ is computed in file [mecanum_drive_controller.cpp](https://github.com/ros-controls/ros_controllers/pull/149/files#diff-e5745efd21bf8a422a81064b2b4a27efR394)
 and is a mean on the 4 wheels, this looks fine to me, no?
And to answer your last question, the reason why there is a frame transformation in the odometry is because the odometry computes a body velocity (i.e. a body twist) but a world position (well the position is actually not expressed in the world frame but in the /odom frame, which can be different).

mecanum_drive_controller/src/odometry.cpp
arennuit: You are right. I can put the filtering back in (if requested by popular demand ;)).

mecanum_drive_controller/src/odometry.cpp
arennuit, line 45: Not sure I understand this one...

mecanum_drive_controller/include/mecanum_drive_controller/odometry.h
arennuit: now fixed

mecanum_drive_controller/include/mecanum_drive_controller/odometry.h
arennuit: now fixed

mecanum_drive_controller/src/odometry.cpp
arennuit, line 53: You are right. I can put the filtering back in (if requested by popular demand ;)).

mecanum_drive_controller/test/diff_drive_test.cpp
arennuit, line 1: Indeed, I have not been able to simulate mecanum wheels in gazebo so I now use VRep. I am not sure how to handle this one... This is the reason why there is currently no unit test for the controller.
This is a blocking problem for me.

mecanum_drive_controller/include/mecanum_drive_controller/odometry.h
arennuit: now fixed

mecanum_drive_controller/include/mecanum_drive_controller/odometry.h
arennuit: now fixed

mecanum_drive_controller/src/mecanum_drive_controller.cpp
arennuit: now fixed

arennuit:
Also, have you experimented with the quality of the odometry? I have a large drift which I am not happy with: a few centimers on a 10s movement - could be due to the fact I use simulation at 40Hz only. Can you let me know if the results you obtain with the odometry are good enough in your case?

--- UPDATE ---

The drift was actually caused by the feedback using odometry measures (there is not much that can be done here: drift is a well know problem of odometry). So I have double checked and I am now confident that the IK and FK bits of code are fine.

arennuit:
I have thought about a workaround to get useful unit tests despite of not being able to simulate mecanum wheels in gazebo. We could actually design a robot using standard wheels (rather than mecanum wheels) and unit test the mecanum IK and FK (odometry) on this scene.
Running the scene and the unit test would not show that the controller is working (the behavior would obviously be nonesense) but this unit test could actually show that the behavior of the code does not change over time... which is actually the reason of being unit tests.
We know that the controller works, if we enforce its behavior does not change over time then we get what we want out of the unit tests, no?

mecanum_drive_controller/test/diff_drive_test.cpp
arennuit, line 1: Hello @efernandez, I have actually fixed the code according to your review. Thanks for your help ;)

mikepurvis:
Note that I'm working on trying to simulate these wheels in Gazebo using directional friction. I believe it's pretty close to functional, though there is some remaining oddball behaviour that I was hoping to get some assistance from the Gazebo devs on: http://answers.gazebosim.org/question/8693/mecanum-simulator-with-directional-friction/

adolfo-rt:
Can't you get some basic testing in place without a dynamic simulator?. Something like comparing controller outputs to known correct results?. Some white-box (implementation aware) testing could maybe exercise different known code paths, or validate different configurations (wheel placement, size, etc.). I typically like to unit test the guts (actual computation) in simple, very targeted tests, and then test the whole system (ROS interfaces) in a higher-level, but less exhaustive integration rostest.

arennuit:
@mikepurvis interesting, keep us informed on how this goes...

@adolfo-rt you are right the gazebo-based tests are more _functional test_ than _unit tests_. What you propose is real unit tests (atomic as unit tests should be).

LiohAu:
any chance to see this merged one day ? :)

mikepurvis:
I pulled the guts of it into `ridgeback_control`: https://github.com/ridgeback/ridgeback/tree/indigo-devel/ridgeback_control/include/mecanum_drive_controller

jonbinney:
I'd love to have this merged in; it sounds like lack of tests is the main barrier? @mikepurvis do you have any new thoughts on the feasibility of accurate omni-drive simulation in gazebo?

mikepurvis:
Re-pasting https://github.com/ridgeback/ridgeback/issues/23#issuecomment-415518189:

>I really wanted the directional friction solution to work, and sunk a nontrivial amount of time on it, but all that was ~3 years ago and under Gazebo 7. It's possible that the situation is different under Gazebo 8 or 9— if anyone's interested to pick up the work again, you can start here:
>
>https://github.com/ridgeback/ridgeback_simulator/blob/indigo-devel/mecanum_gazebo_plugin/src/mecanum_plugin.cpp
>
>https://github.com/ridgeback/ridgeback/blob/kinetic-devel/ridgeback_description/urdf/ridgeback.gazebo#L89-L107
>
>Basically just switch the plugins in the URDF over to the commented-out ones, launch the sim, and give it a series of twists that command it to move in X and Y. Generally I found that the "strafe" commands would seem to work correctly in either direction for a metre or two and then it would seemingly go unstable and spin out of control. I strongly suspect that the plugin was not correctly updating the directional friction as the wheel turned, but without a reliable way to visualize the friction vector in realtime, it seems like a dead end to try to fix it.

bmagyar:
Let's try this for Noetic, see ref above

Issue #157, opened 2015-01-22
Make ForwardCommandController and ForwardJointGroupCommandController thread-safe


adolfo-rt:
There is currently no synchronization between the non-rt thread that executes the subscriber callback and the rt thread that runs the controller update method. A `realtime_tools::RealtimeBuffer` could be used to fix this, as is done in other controllers.

The subject was originally brought up in [this ros-answers discussion](http://answers.ros.org/question/201297/proper-use-of-ros_controls-realtimebuffer/?answer=201521#post-id-201521).

adolfo-rt:
@ipa-fxm you should be aware of this, as I believe you use these controllers.

fmessmer:
Thanks for the hint...
Although we haven't experienced any problem with the current version, I guess #160 fixes this issue.

fmessmer:
#161 fixes it for the forward_command_controllers
However, similar race conditions might occur in other controllers e.g. effort_controllers/JointPositionController as they are using a shared member [here](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/effort_controllers/include/effort_controllers/joint_position_controller.h#L167)

PR #160, opened 2015-01-23
make fcc and fjgcc thread safe

fmessmer:
this is related to #157

### DIFF
--- a/forward_command_controller/include/forward_command_controller/forward_command_controller.h
+++ b/forward_command_controller/include/forward_command_controller/forward_command_controller.h
@@ -42,6 +42,7 @@
 #include <hardware_interface/joint_command_interface.h>
 #include <controller_interface/controller.h>
 #include <std_msgs/Float64.h>
+#include <realtime_tools/realtime_buffer.h>
 
 
 namespace forward_command_controller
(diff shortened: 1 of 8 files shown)


forward_command_controller/include/forward_command_controller/forward_command_controller.h
```diff
@@ -85,14 +86,24 @@ class ForwardCommandController: public controller_interface::Controller<T>
...
+  {
+    tmp_command_ = *(command_.readFromRT());
+    joint_.setCommand(tmp_command_);
+    
+  }
 
   hardware_interface::JointHandle joint_;
-  double command_;
+  realtime_tools::RealtimeBuffer<double> command_;
+  double tmp_command_; // pre-allocated memory that is re-used to set the realtime buffer
```
adolfo-rt, line 98: `tmp_command_` does not have to be a class member, as it will be allocated on the stack as an automatic variable. You could just use it in the `update(...)` method as:

``` c++
const double cmd = *(command_.readFromRT());
joint_.setCommand(cmd);
```

if the command was a more complex structure, involving variable-sized containers for instance, then preallocation would definitely make sense.

fmessmer, line 98: I thought about that, too, but then I decided to keep things similar, thus having the member variable for the ForwardCommandController as well as it can be found in the ForwardJointGroupCommandController and the effort_controllers::JointPositionController!

I was even wondering whether actually using a member variable may cause conflicts again between the callback and update function, i.e. whether we in general only should have local variables instead. Or is this resolved through the RealtimeBuffer?

@adolfo-rt: let me know how you would like to have it.

fmessmer:
Closing this PR as superseding PR #161 has been merged...

PR #161, opened 2015-01-28
thread-safe forward controllers

mathias-luedtke:
Fix for issue #157, supersedes #160 

ForwardCommandController was not thread-safe for 32bit systems, but  maybe for 64bit.
ForwardJointGroupCommandController was not thread-safe at all.
- thread-safe/realtime-safe topic handling
- check for empty list of joints
- properly resets to neutral zeros at restart (resize only works the first time)
- dependency on realtime_tools
- command(s)_ was renamed to command(s)_buffer_.

Please note: this version uses references to the realtime-data, this works only for single-consumer cases. They could be replaced by copies, but this introduces some overhead..

### DIFF
--- a/effort_controllers/src/joint_group_effort_controller.cpp
+++ b/effort_controllers/src/joint_group_effort_controller.cpp
@@ -42,7 +42,7 @@ template <class T>
 void forward_command_controller::ForwardJointGroupCommandController<T>::starting(const ros::Time& time)
 {
   // Start controller with 0.0 efforts
-  commands_.resize(n_joints_, 0.0);
+  commands_buffer_.readFromRT()->assign(n_joints_, 0.0);
 }
 
 
--- a/forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
+++ b/forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
@@ -46,6 +46,7 @@
 #include <hardware_interface/joint_command_interface.h>
 #include <controller_interface/controller.h>
 #include <std_msgs/Float64MultiArray.h>
+#include <realtime_tools/realtime_buffer.h>
 
 
 namespace forward_command_controller
@@ -72,7 +73,7 @@ template <class T>
 class ForwardJointGroupCommandController: public controller_interface::Controller<T>
 {
 public:
-  ForwardJointGroupCommandController() { commands_.clear(); }
+  ForwardJointGroupCommandController() {}
   ~ForwardJointGroupCommandController() {sub_command_.shutdown();}
 
   bool init(T* hw, ros::NodeHandle &n)
@@ -102,7 +103,9 @@ public:
         return false;
       }
     }
-    
+
+    commands_buffer_.writeFromNonRT(std::vector<double>(n_joints_, 0.0));
+
     sub_command_ = n.subscribe<std_msgs::Float64MultiArray>("command", 1, &ForwardJointGroupCommandController::commandCB, this);
     return true;
   }
@@ -110,13 +113,14 @@ public:
   void starting(const ros::Time& time);
   void update(const ros::Time& time, const ros::Duration& period) 
   {
+    std::vector<double> & commands = *commands_buffer_.readFromRT();
     for(unsigned int i=0; i<n_joints_; i++)
-    {  joints_[i].setCommand(commands_[i]);  }
+    {  joints_[i].setCommand(commands[i]);  }
   }
 
   std::vector< std::string > joint_names_;
   std::vector< hardware_interface::JointHandle > joints_;
-  std::vector< double > commands_;
+  realtime_tools::RealtimeBuffer<std::vector<double> > commands_buffer_;
   unsigned int n_joints_;
 
 private:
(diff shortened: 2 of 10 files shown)


forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
```diff
@@ -98,6 +103,7 @@ class ForwardJointGroupCommandController: public controller_interface::Controlle
         return false;
       }
     }
+    commands_buffer_.readFromRT()->resize(n_joints_, 0.0);
```
adolfo-rt: Use [`readFromNonRT()`](https://github.com/ros-controls/realtime_tools/blob/indigo-devel/include/realtime_tools/realtime_buffer.h#L109) here. The implementation will make sure that it waits until it gets the shared resource. It's OK to wait for a resource during `init`.

effort_controllers/src/joint_group_effort_controller.cpp
```diff
@@ -42,7 +42,7 @@ template <class T>
 void forward_command_controller::ForwardJointGroupCommandController<T>::starting(const ros::Time& time)
 {
   // Start controller with 0.0 efforts
-  commands_.resize(n_joints_, 0.0);
+  commands_buffer_.readFromRT()->assign(n_joints_, 0.0);
```
adolfo-rt, line 45: You get no protection on the shared resource by doing this, because:
- You first get a pointer to the shared resource on `readFromRT()`, that's OK.
- You then write into the shared resource with no synchronization primitives in place.

The thing here is that `realtime_tools::RealtimeBuffer` does not handle well writing from real-time.

You can likely get away with the trick, because:
- The controller update in the `controller_manager` is serialized, and
- If all other writers to the shared resource (this would be controller ROS callbacks) use `writeFromNonRT`, they [will not write](https://github.com/ros-controls/realtime_tools/blob/indigo-devel/include/realtime_tools/realtime_buffer.h#L125) into the pointer [you're manipulating](https://github.com/ros-controls/realtime_tools/blob/indigo-devel/include/realtime_tools/realtime_buffer.h#L106).

So although kludgy, it will actually work.

The alternatives would be to either use `writeFromNonRT` and wait for the lock, or not use `realtime_tools::RealtimeBuffer` at all.

position_controllers/src/joint_group_position_controller.cpp
```diff
@@ -42,10 +42,10 @@ template <class T>
 void forward_command_controller::ForwardJointGroupCommandController<T>::starting(const ros::Time& time)
 {
   // Start controller with current joint positions
-  commands_.resize(n_joints_, 0.0);
+  std::vector<double> & commands = *commands_buffer_.readFromRT();
```
adolfo-rt, line 45: Same thing I mention above about writing into a shared resource without synchronization primitives in place.

velocity_controllers/src/joint_group_velocity_controller.cpp
```diff
@@ -42,7 +42,7 @@ template <class T>
 void forward_command_controller::ForwardJointGroupCommandController<T>::starting(const ros::Time& time)
 {
   // Start controller with 0.0 velocities
-  commands_.resize(n_joints_, 0.0);
+  commands_buffer_.readFromRT()->assign(n_joints_, 0.0);
```
adolfo-rt, line 45: Same thing I mention above about writing into a shared resource without synchronization primitives in place.

forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
mathias-luedtke: If I use readFromNonRT here, it works because the creation of the subscriber after the resize, serializes the unprotected write access (and because starting is not concurrent to init).
If I use readFromtRT here, it works because init, starting and update are serialized through the controller_manager.
None of both ways provides a protective write. The RealtimeBuffer lock just protects the pointer swap.
Maybe writeFromNonRT (with the copy penalty) is the least error-prone solution for later modifications.
However, it still depends on the serialization of init, starting and update.

effort_controllers/src/joint_group_effort_controller.cpp
mathias-luedtke, line 45: I am aware of this shortcoming, but it works for the current solution, as you already pointed out..
Initially I didn't want to use the RealtimeBuffer, but it is currently the best (and tested) solution we have..

mathias-luedtke:
I have commented on the specific lines..
Since all your comments come down to the drawbacks of RealtimeBuffer (that I circumvented by the design of the  implementation), I highly suggest that you merge it anyway..
The old implementation leads to segmenation faults if the input topic was published before the controller was actually started (race condition with allocation/resize in commandCB and starting).

I have already thought about alternatives to RealtimeBuffer, I will open an issue in realtime_tools for discussion on this as soon I got more insights on the alternatives.

forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
adolfo-rt: You can first `readFromNonRT()`, set the new values and then  `writeFronNonRT()`. You'll wait twice for the lock, once in each call, but that's still OK.

In fact, in the above usecase you don't actually need to read the buffer at all. You can simply prepare a vector and call `writeFromNonRT()`.

The thing to avoid doing is call one of the read methods and manipulate the raw pointer afterwards.

adolfo-rt:
Please take a look at the inline comment on the init method. We can then merge.

mathias-luedtke:
Done :)

adolfo-rt:
Could you please squash the last three commits together, and leave 4cdc1e4 as-is?. This will simplify cherry-picking the fix to other branches.

adolfo-rt:
Great, thanks for making these controllers better!.

PR #163, opened 2015-02-19
Add option to obtain velocity directly from hardware

### LINKED ISSUE #159: diff_drive_controller doesn't use joint.velocity
In https://github.com/ros-controls/ros_controllers/blob/indigo-devel/diff_drive_controller/src/odometry.cpp#L78, diff_drive_controller currently differentiates joint.position to obtain velocity, differentiates again to obtain acceleration, and then obtains the final odometry twist component by integrating acceleration.

This results in filtered twist component on the outgoing odometry with a very slow response. Since it's possible to update joint.velocity directly from hardware, would it be possible to generate the twist message directly from this information, improving the response? Could either parametrize or detect from incoming data whether the velocity information is filled out.

paulbovbel:
Fix #159 

Option to use joint velocity from hardware to update odometry twist. Much faster response than rolling mean implementation. Joint position still used in either case to update odometry pose.

### DIFF
--- a/effort_controllers/src/joint_effort_controller.cpp
+++ b/effort_controllers/src/joint_effort_controller.cpp
@@ -41,7 +41,7 @@ template <class T>
 void forward_command_controller::ForwardCommandController<T>::starting(const ros::Time& time)
 {
   // Start controller with 0.0 effort
-  command_ = 0.0;
+  command_buffer_.writeFromNonRT(0.0);
 }
 
 
--- a/effort_controllers/src/joint_group_effort_controller.cpp
+++ b/effort_controllers/src/joint_group_effort_controller.cpp
@@ -42,7 +42,7 @@ template <class T>
 void forward_command_controller::ForwardJointGroupCommandController<T>::starting(const ros::Time& time)
 {
   // Start controller with 0.0 efforts
-  commands_.resize(n_joints_, 0.0);
+  commands_buffer_.readFromRT()->assign(n_joints_, 0.0);
 }
 
 
--- a/forward_command_controller/CMakeLists.txt
+++ b/forward_command_controller/CMakeLists.txt
@@ -2,11 +2,11 @@ cmake_minimum_required(VERSION 2.8.3)
 project(forward_command_controller)
 
 # Load catkin and all dependencies required for this package
-find_package(catkin REQUIRED COMPONENTS controller_interface hardware_interface std_msgs)
+find_package(catkin REQUIRED COMPONENTS controller_interface hardware_interface std_msgs realtime_tools)
 
 # Declare catkin package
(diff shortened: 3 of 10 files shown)


efernandez:
LGTM, but this needs a new test to evaluate the case when `estimate_velocity_from_position: false`.
However, I still wonder if this is the better place to do it, or maybe all this logic should go to the joint level (taking the velocity here, instead of estimating it).

@bmagyar what do you think?

paulbovbel:
If you mean that the robothw implementations should provide the velocity (whether estimated in software or read from hardware), I would agree (proposed this in #159).

paulbovbel:
An alternative could be using NaNs to indicate unused fields in the JointStateHandle, so that diff_drive_controller can decide whether to estimate velocity or not, however this may be getting too clever.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -178,6 +179,10 @@ namespace diff_drive_controller{
     controller_nh.param("enable_odom_tf", enable_odom_tf_, enable_odom_tf_);
     ROS_INFO_STREAM_NAMED(name_, "Publishing to tf is " << (enable_odom_tf_?"enabled":"disabled"));
 
+    controller_nh.param("estimate_velocity_from_position", estimate_velocity_from_position_, estimate_velocity_from_position_);
+    ROS_INFO_STREAM_NAMED(name_, "Estimating joint velocity from joint position is  " <<
+        (enable_odom_tf_?"enabled":"disabled"));
```
bmagyar: This is a bit redundant, I guess you meant `estimate_velocity_from_position_` here :)

bmagyar:
As Enrique mentioned above, we would appreciate a test as well that exercises the param `estimate_velocity_from_position`. Generally I'd add a test for 3 cases and have a very simple test for each and see if the controller works, doesn't work of fails how it is supposed to or not.
- `estimate_velocity_from_position: false`
- `estimate_velocity_from_position: true`
- `estimate_velocity_from_position: something_invalid`

...but in general, it looks good to me.

If you wish we could start a discussion on the velocity estimation topic but I'm afraid that would take more time. I'm more neutral on this one.

paulbovbel:
Typo fixed. I would appreciate any guidance on how you feel I could test this behaviour - `estimate_velocity_from_position` falls back to the default (`true`) if `something_invalid` is entered, and the outputs should be fairly similar in simulation in either the `true` or `false` case.

I'm fairly neutral on further changes to velocity estimation, other than that it should be optional, and it would be great if there was a way for hardware to mark 'unused' fields in JointState handles as `nan` or similar, since 0 is otherwise valid.

PR #169, opened 2015-03-29
[diff_drive_controller] Add wheel_separation and wheel_radius parameters

bmagyar:
@T045T, @efernandez


bmagyar:
@T045T, you have push access to the branch of this PR in my fork. :+1:

T045T:
@bmagyar I've written some tests, and rewritten the code so it's possible to specify just the separation or just the radius as a parameter. Can you review these changes?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -193,8 +193,31 @@ namespace diff_drive_controller{
...
     controller_nh.param("angular/z/min_acceleration"       , limiter_ang_.min_acceleration       , -limiter_ang_.max_acceleration      );
 
-    if (!setOdomParamsFromUrdf(root_nh, left_wheel_names[0], right_wheel_names[0]))
-      return false;
+    // If either parameter is not available, we need to look up the value in the URDF
+    bool lookup_wheel_separation = !controller_nh.getParam("wheel_separation", wheel_separation_);
+    bool lookup_wheel_radius = !controller_nh.getParam("wheel_radius", wheel_radius_);
+
+    // Short-circuit in case we don't need to look up anything, so we don't have to parse the URDF
+    if(lookup_wheel_separation or lookup_wheel_radius)
```
bmagyar: Excellent idea, but then why not short-circuit in the function? This would move the bloat a bit away from the relatively clean param parsing chunk of code.

T045T: Good point, I'll change that. Also, since `setOdomParamsFromUrdf` doesn't actually set the parameters any more, is it okay to rename it to `lookupOdomParamsInUrdf`?

bmagyar: I would say it does.
https://github.com/bmagyar/ros_controllers/blob/separation_and_diameter_params/diff_drive_controller/src/diff_drive_controller.cpp#L476

T045T: But the actual odometry values are now being set [here](https://github.com/bmagyar/ros_controllers/blob/separation_and_diameter_params/diff_drive_controller/src/diff_drive_controller.cpp#L217). This is obviously a minor point, I just want to make sure the function name doesn't confuse future developers :)

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -407,8 +427,16 @@ namespace diff_drive_controller{
 
   bool DiffDriveController::setOdomParamsFromUrdf(ros::NodeHandle& root_nh,
                              const std::string& left_wheel_name,
-                             const std::string& right_wheel_name)
+                             const std::string& right_wheel_name,
+                             bool lookup_wheel_separation,
+                             bool lookup_wheel_radius)
   {
+    if(not(lookup_wheel_separation or lookup_wheel_radius))
```
bmagyar: Even though I think these keywords are awesome, the rest of the project is not using them, so I'd ask you to remove them to keep the consistency of `ros_controllers`

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -193,8 +193,28 @@ namespace diff_drive_controller{
     controller_nh.param("angular/z/max_acceleration"       , limiter_ang_.max_acceleration       ,  limiter_ang_.max_acceleration      );
     controller_nh.param("angular/z/min_acceleration"       , limiter_ang_.min_acceleration       , -limiter_ang_.max_acceleration      );
 
-    if (!setOdomParamsFromUrdf(root_nh, left_wheel_names[0], right_wheel_names[0]))
+    // If either parameter is not available, we need to look up the value in the URDF
+    bool lookup_wheel_separation = !controller_nh.getParam("wheel_separation", wheel_separation_);
+    bool lookup_wheel_radius = !controller_nh.getParam("wheel_radius", wheel_radius_);
+
+
```
bmagyar: You have 2 empty lines here, one would be enough.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -421,46 +449,48 @@ namespace diff_drive_controller{
...
+    if(lookup_wheel_separation)
     {
-      ROS_ERROR_STREAM_NAMED(name_, right_wheel_name
-                             << " couldn't be retrieved from model description");
-      return false;
-    }
+      // Get wheel separation
+      if(!left_wheel_joint)
+        {
```
bmagyar: A little bit of misindentation here

T045T:
Thanks! I've fixed those three :)

bmagyar:
Nice work!
I made small pedantic notes and I ran all the tests and they are fine.

But... I have spotted that the urdf you added with spherical wheels is actually almost the same as this one:
https://github.com/bmagyar/ros_controllers/blob/indigo-devel/diff_drive_controller/test/diffbot_bad.xacro
At that moment this was considered a "bad" urdf, but not I believe it could be used without any issues. Actually the one you added has a more proper name. 
There is one test that checks for a non-standard urdf, so your test doesn't need to validate the "no param added and urdf is not the expected structure" case, since there is a test for that already. But! it would be good to make it more clear what bad urdf means in those cases.

T045T:
I looked at the bad URDF test as well, and in addition to the spherical wheels, it also has the wrong joint names.
The [controller config](https://github.com/bmagyar/ros_controllers/blob/indigo-devel/diff_drive_controller/test/diffbot_controllers.yaml) uses `wheel_0_joint` and `wheel_1_joint`, while the URDF defines `joint_w1` and `joint_w2`, so it can't be used for a passing test, sadly.

Looks like the wront joint names weren't actually what that test was supposed to check though (`diffbot_drive_wrong.test` does that), so maybe we can just remove the outdated "bad" URDF outright and replace `diffbot_drive_bad_urdf.test` with `diffbot_drive_radius_param_fail.test`?

bmagyar:
Good catch!
Give it a shot and if all unit tests are still ok, we should be done.

efernandez:
Nice work guys... I don't have the cycles to check this now, but it seems OK at first sight.

paulbovbel:
+1, tested with https://github.com/clearpathrobotics/urdf_tutorial/tree/ros-control-tutorials, works as expected

bmagyar:
+1 from me too.

@adolfo-rt ?

adolfo-rt:
Please update the package's ROS wiki page, so users can know about the new configuration alternatives.

Thanks for the patch, and for increasing test coverage :)

paulbovbel:
Would you please do a release with these changes, so that I can work diff_drive_controller into http://wiki.ros.org/urdf/Tutorials

T045T:
I've updated http://wiki.ros.org/diff_drive_controller with the new parameters.

bmagyar:
Awesome, thanks Nils!

Issue #167, opened 2015-03-27
diff_drive_controller assumes joint parent == origin


T045T:
I'm currently working on a URDF file for a differential drive robot that's build from several (potentially reusable) parts. This means that the drive wheel joints' `parent` element is not always `base_link`, or even in a different position at all.

In [diff_drive_controller.cpp](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/diff_drive_controller/src/diff_drive_controller.cpp#L440), `left_wheel_joint->parent_to_joint_origin_transform` is reported as the distance to origin. However, that is not true.

In my specific case, the parent link is in the same location, but oriented differently, resulting in a wheel separation of 0.

I have two ideas on how to address this (not counting a change to the URDF):
- Look up the wheel separation via `tf`
- add a `wheel_separation` parameter that overrides the lookup in the URDF

Please let me know what you think of these - I can implement either or both of them, but would like some feedback first.

bmagyar:
Hi,

First of all, which ROS distro are you using? Not that it makes any difference, but it is easier to get new stuff released into indigo.

Actually I have already implemented a solution for this using a parameter for 'wheel_separation' and also one for 'wheel_diameter' (on a robocup project for this year we reinvented the wheel in a way where cylinders are not good approximations anymore).

So the solution for this is done. I can prepare a pull request of it from my private repo, but I will need a test. Do you mind implementing it? I can give you access to my fork so you can push there directly. There are plenty of examples on tests in the appropriate folder.

As a final note, I find it hard and risky to add a tf listener to a realtime-safe module.

T045T:
I'm using indigo - sorry I didn't mention that before.

Sure, I can write a test for this :)

And you're right, I forgot about the realtime requirement - although in this case it would just be a one-time lookup of a (hopefully) static transform.

bmagyar:
#169

bmagyar:
@adolfo-rt  I believe we can close this one.

PR #171, opened 2015-04-02
Traverse URDF to base_frame to calculate wheel_separation

paulbovbel:
Previous behaviour was to examine only the joint connected the wheel link to its parent, which fails if wheel is not connected directly to base_link.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -86,6 +86,38 @@ static bool isCylinder(const boost::shared_ptr<const urdf::Link>& link)
 }
 
 /*
+ * \brief Get vector between initial and root link, searching updwards through URDF link tree
+ * \param [in] initial_link Origin link for vector
+ * \param [in] root_link Root link of model [m]
+ * \param [out] vector Vector from origin to root link [m]
+ * \return true if the vector was calculated, false otherwise
+ */
+static bool getLinkToRootVector(const boost::shared_ptr<const urdf::Link>& initial_link,
+                                const boost::shared_ptr<const urdf::Link>& root_link,
+                                urdf::Vector3& vector)
+{
+  boost::shared_ptr<const urdf::Link> link = initial_link;
+
+  while(link->name != root_link->name)
+  {
+    if(!link->getParent())
+    {
+      ROS_ERROR_STREAM("Link " << initial_link->name << " does not have link " << root_link->name << " as an ancestor");
+      return false;
+    }
+    if(!link->parent_joint)
+    {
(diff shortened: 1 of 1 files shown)


efernandez:
@paulbovbel The euclidean distance between vectors might be modified if the left wheel to `base_link` and right wheel to `base_link` segments aren't colinear.

IMHO, for these cases it might be better to use a parameter to specify the wheel separation.
@bmagyar @artivis

paulbovbel:
I can't think of any obvious configurations where wheels wouldn't be colinear, but it would certainly be helpful to be able to override wheel separation in general. 

This patch however would be a good start to having the controller work for URDFs where the wheels aren't attached directly to the root link. I noticed this because I'm working with the urdf tutorials (http://wiki.ros.org/urdf/Tutorials/Using%20Xacro%20to%20Clean%20Up%20a%20URDF%20File) to add a diff_drive_controller example, but the model has wheel->leg->base_link.

bmagyar:
+1 for not using tf. 
This solution though assumes that all your frames between wheel and base share the same orientation which is not enforced anywhere. Adding up the position elements into a final vector is only valid when this orientation constraint is true.

What do you think about this?
Issue: https://github.com/ros-controls/ros_controllers/issues/167
PR: #169

paulbovbel:
That's great, I didn't spot that thread before. For getting the wheel separation from URDF when it's not overriden via params, would you still be interested in the URDF traversal implemented in this pull? If so, I can rebase this on top of that pull once you merge it in.

EDIT: @bmagyar, I just understood what you meant about the 'same orientation', that would be an unpleasant surprise for a user. I can think about how to rework this to take any rotations in account, probably using the KDL-URDF parser.

paulbovbel:
So, are you interested in a more 'robust' wheel_sep autodetection method rebased onto #169 ? If so, then I can investigate using KDL to get the proper wheel separation that includes any orientation changes.

bmagyar:
Hi Paul,

Yes I do agree that removing this assumption would be beneficial, ad the KDL-URDF parser sounds good but on the other hand I'm wondering if it makes life easier for anyone apart from adding code to the controller. Personally I would prefer to keep the controller as clean as possible but I am generally open to  new features provided that there is a strong need for them and they acn be introduced in a backward-compatible way (as you see how many are coming in, and you don't see how many we have internally in PAL ;))

paulbovbel:
@bmagyar , thanks, I'd like to come back to this. Meanwhile, could you please release a new version of ros_controllers, so that #169 is included?

adolfo-rt:
> @bmagyar , thanks, I'd like to come back to this. Meanwhile, could you please release a new version of ros_controllers, so that #169 is included?

https://github.com/ros/rosdistro/pull/8324

bmagyar:
Thanks Adolfo!

PR #170, opened 2015-04-01
[Indigo] Reset to semantic zero in HardwareInterfaceAdapter for PositionJointInterface

fmessmer:
Semantic zero was set for VelocityJointInterface (i.e. 0.0) and EffortJointInterface (i.e. 0.0), but not for PositionJointInterface (i.e. currentPosition)

Issue #144, opened 2014-10-24
hold position behavior in face of steady state error


dcconner:
We're seeing an issue on robot where a joint is physically unable to reach desired position due to actuator force limitation.  This results in a steady state error.

Thus we get a constant control command= k_p \* error.

When the trajectory enters the setHoldPosition it uses getPosition (actual position) on https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L689
This takes error to zero, and reduces the servo command value causing a perturbation in the system.

For cases with potential steady state error (e.g. holding a heavy load), should we not hold to the previous desired position to provide constant control.

Only on controller (re)start should we initialize desired position equal to the actual position.

adolfo-rt:
The `setHoldPosition` method is called when:
- The controller starts
- An action goal is canceled
- An empty trajectory is sent through the _topic_ interface (the topic way of canceling)

I would argue that if you cancel execution, you expect joints to stop, and remain stopped. If we hold desired position and the error is large, the source of error might disappear at some point and cause the robot to jump to a far-away reference. This might happen a long time after we entered hold position mode, and could be very surprising and undesired. Example error sources that can disappear/change are a moving obstacle or the robot reorienting itself differently wrt gravity.

I'm still open to discussing the proposal, but it seems to me that this should be solved at a lower level.

adolfo-rt:
An alternative that comes to mind would be to reimplement `setHoldPosition` such that when  the `stop_time` parameter is zero (the default), the current behavior is preserved, but when greater than zero it computes a stopping trajectory from the _desired_ position and velocity  to the _actual_ position and zero velocity.

This might be an acceptable compromise.

dcconner:
Not sure the suggested compromise addresses the issue.  Anytime the output "desired" value tracks actual, the system with not hold if the code is depending on proportional error gain to hold the required effort.

I need to do some more digging, but that must wait a bit.

adolfo-rt:
My understanding of your problem is that you get a perturbation because the desired state discretely jumps to track the current state. The proposed alternative would create a trajectory that goes from the desired state to the current one in a specified time, making the transition continuous.

sachinchitta:
@dcconner - what is your underlying actuator interface? Are you specifying torques to hardware level controllers or position-velocity setpoints? You might need a feed-forward term to compensate for the situation you are seeing.

[...]

PR #176, opened 2015-05-06
Set desired in starting

sachinchitta:
I think desired state should be updated in starting to match the current state. (Also current state should first be initialized here since the joints are going to be read anyway in the first setHold()).

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -696,18 +696,12 @@ setHoldPosition(const ros::Time& time)
   const unsigned int n_joints = joints_.size();
   for (unsigned int i = 0; i < n_joints; ++i)
   {
-    //    hold_start_state_.position[i]     =  joints_[i].getPosition();
-    //    hold_start_state_.velocity[i]     =  joints_[i].getVelocity();
     hold_start_state_.position[i] = desired_state_.position[i];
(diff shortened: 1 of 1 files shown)


adolfo-rt:
I'd like to understand what the problem is, and I can't figure it out only by looking at the proposed solution. Could you please elaborate on this?.

I have two questions on the patch:
1. Why does `current_state_` need to be refreshed on `starting()`?. I don't see any harm in doing it, but I also don't see why it's necessary. The variable is not used in `starting()`, but in `update()`, where it currently gets refreshed.
2. Why do you need to set `desired_state_ = current_state_`. This is supposed to be done in `setHoldPosition(time_data.uptime)`, which takes place just after your patch.

Maybe there's something amiss on how `setHoldPosition()` is implemented, which is causing you trouble.

dcconner:
Our approach, related to #144 and #177, was to have the joints desired pose stored in HWInterface, and upon switching controllers set the desired to that prior desired value, and initialize integral terms in PID to give bumpless control.  Jumping to current  actual position for the desired on startup may not give you the desired behavior, but the actual behavior will be somewhat application specific.  Again, I don't think there is a single best solution, so having some easy way to change implementations of this low level behavior might be in order long term.

We just did our own thing in a custom fork, and didn't intend to change the standard behavior.

fmessmer:
I think I proposed something similar in #170!
The JointPositionInterface, definitely does not start in "semantic zero".

jschleicher:
Can be closed here.

The diff was part of the discussion in #297 and was merged there.

PR #177, opened 2015-05-06
Set hold using desired - only for discussion first

sachinchitta:
I am creating this PR mainly to discuss the way setHold operates. Here's how I came to this PR:

I am working with an industrial robot using a position command interface constantly streaming points. When I abort a trajectory that had already been sent, I found that the industrial controller would complain and shut down with an "acceleration command exceeded" error message. Digging a little deeper, I realized that the robot was lagging the given command. When the trajectory was aborted, setHold uses the actual robot position to generate a new segment that will stop the robot. However, this creates a discontinuity in the command to the robot (since actual was lagging command) creating a request that seems to require a large velocity/acceleration. Naturally, the robot controller complains and shuts down. 

I have seen this happen on other robots too (in a different context not using this particular controller) when they have trouble tracking and the command gets reset using actual instead of the previous command. The particular behavior, of course, depends on how the internal robot controllers act  - asking for a big jump in command triggers errors on some controllers. 

This PR contains my fix (but probably requires some changes to be realtime correct). I am using desired (instead of actual) to generate the new segment in setHold. This works for me right now but putting it up for discussion. Note this PR also depends on #176.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -696,18 +696,12 @@ setHoldPosition(const ros::Time& time)
   const unsigned int n_joints = joints_.size();
   for (unsigned int i = 0; i < n_joints; ++i)
   {
-    //    hold_start_state_.position[i]     =  joints_[i].getPosition();
-    //    hold_start_state_.velocity[i]     =  joints_[i].getVelocity();
     hold_start_state_.position[i] = desired_state_.position[i];
     hold_start_state_.velocity[i] = desired_state_.velocity[i];
-
     hold_start_state_.acceleration[i] =  0.0;
 
-    //    hold_end_state_.position[i]       =  joints_[i].getPosition();
-    //    hold_end_state_.velocity[i]       = -joints_[i].getVelocity();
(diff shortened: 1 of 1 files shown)


adolfo-rt:
Related discussion: https://github.com/ros-controls/ros_controllers/issues/144. In particular, read this comment and the following one: https://github.com/ros-controls/ros_controllers/issues/144#issuecomment-60375667

adolfo-rt:
Holding position at the desired state is an unsafe choice when the tracking error is not in steady-state and has a large magnitude. In steady-state it makes no matter if the error is large, the robot will hold position nicely. Conversely, if there's a large non-steady-state error, upon requesting to hold position the robot _might_ keep on moving, potentially traveling a large distance and at high speed. This is both unexpected and unsafe.

My suggestion is to add a `joint_limits_interface` or similar to your robot hardware abstraction. This way, it does not matter which controller is sending commands, or if they are too aggressive, you have the guarantee that the limits you want to enforce are respected (position, velocity, acceleration, effort, ...).

dcconner:
I commented on #144.  I don't agree with Adolfo's assertion that "in steady-state it makes no matter if the error is large, the robot will hold position nicely. "  If torque is proportional to error, then changing to current actual position will change torque and allow the joint to move in some applications.  I don't think there is a single best solution for all applications.

jschleicher:
Superseded by #297.

Can be closed here.

PR #178, opened 2015-05-12
Enable usage of namespaces for joint position controllers.

pfotzer:
In order to be able to use namespaces e.g. for multiple robots, the joint position controllers in effort_controllers and velocity_controllers need a possibility to load the robot_description from the parameter server located inside of a namespace. The best would be to also remove the hard coded "robot_description" in future.
The URDF model has to be loaded in <namespace>/robot_description on the parameter server.


adolfo-rt:
Looks like a duplicate of #98. It is currently possible to have multiple robots under robot-specific namespaces. Please take a look at #98 for recommended usage patterns.

PR #183, opened 2015-07-12
Sync with hydro-devel branch in PAL Robotics fork

efernandez:
This takes the missing commits from hydro-devel branch in PAL Robotics fork.

I've discarded many commits that were already merged upstream, recovered some old missed unit tests and fix a few conflicts for new features that weren't already merged. In brief, this PR adds:
1. Test for bad URDF
2. Test for NaN
3. `velocity_rolling_window_size` param
4. ~~`preserve_turning_radius` param~~
5. Limit jerk
6. Other minor fixes

Note that it's recommended to keep 4 and 5 features disabled, as they are quite new/experimental.

All tests passed.


efernandez:
@adolfo-rt @bmagyar

efernandez:
@adolfo-rt @bmagyar ... just pinging

bmagyar:
:+1:

adolfo-rt:
Thanks @efernandez ,the PR is looking very good this time.

My only comments are on the two experimental features:
- Experimental features should be clearly documented, and disabled by default.
- What makes them experimental?. What would be required to consider them stable?.
- Will there be a commitment from your part to make them stable?.

More generally, would it make sense to defer merging these two features until they can be considered stable?. I'd like to avoid merging a WIP feature that never ends up getting into shape.

efernandez:
@adolfo-rt 
- I've disabled the experimental preserve_turning_radius feature by default.
- It's experimental in the sense that in some corner cases it might produce side effects.
- I'll work on making it stable and manage all corner cases.

Regarding the jerk limits, they simply need proper tuning. I shouldn't have said it's experimental. I works as expected but would create oscillations if bad limits are provided. For that reason is disabled by default.

adolfo-rt:
OK for jerk limits needing tuning.

Last question: What are the side effects of `preserve_turning_radius`?.
- Should they not be there?. Does the implementation need to be revised?.
- Is it a parameter tuning issue? (doesn't seem like).

efernandez:
The side effects of `preserve_turning_radius` are basically that trying to keep the turning radius some limits might not be enforced in some cases (depends on the linear and angular speed). The implementation needs to be extended to support all the cases properly. They aren't trivial changes, but I'll work on that at the earliest.

adolfo-rt:
OK, so how about removing 1f45940 from the PR, and creating a separate PR it once it's ready for prime-time?. The other changes look good.

efernandez:
@adolfo-rt OK, I've just removed 1f45940 from the PR, and updated its description.

adolfo-rt:
Thank you so much for iterating with me, @efernandez. Merging.

I'll be glad to merge the preserve turning radius feature once it's good to go.

efernandez:
Thanks for the review @adolfo-rt 

Now I won't give too much urgency to the preserve turning radius feature, but I'll try to work on it during the next months.

PR #184, opened 2015-07-31
Allow flexible subclassing of fwd jnt group cmd

ghost:
Without changing functionality for the default implementation,
the following changes were made to allow subclassing:
- Changed private members to protected
- Added initCommandSub to make the subscribed topic and callback
  function configurable
- Virtualized protected functions to allow for subclasses to specify
  overriding subscriber and callback

To see example subclasses using this implementation see:
baxter_position_controller.h
https://gist.github.com/rethink-imcmahon/9e10320f5fe83ce390d8
baxter_position_controller.cpp
https://gist.github.com/rethink-imcmahon/8764ddd006180c0b58b8

### DIFF
--- a/forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
+++ b/forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
@@ -106,7 +106,7 @@ public:
 
     commands_buffer_.writeFromNonRT(std::vector<double>(n_joints_, 0.0));
 
-    sub_command_ = n.subscribe<std_msgs::Float64MultiArray>("command", 1, &ForwardJointGroupCommandController::commandCB, this);
+    initCommandSub(n);
     return true;
   }
 
@@ -123,9 +123,13 @@ public:
   realtime_tools::RealtimeBuffer<std::vector<double> > commands_buffer_;
   unsigned int n_joints_;
 
-private:
+protected:
   ros::Subscriber sub_command_;
-  void commandCB(const std_msgs::Float64MultiArrayConstPtr& msg) 
+  virtual void initCommandSub(ros::NodeHandle &n)
+  {
+    sub_command_ = n.subscribe<std_msgs::Float64MultiArray>("command", 1, &ForwardJointGroupCommandController::commandCB, this);
+  }
+  virtual void commandCB(const std_msgs::Float64MultiArrayConstPtr& msg)
   {
     if(msg->data.size()!=n_joints_)
     { 


rethink-imcmahon:
Tangentially related, I've rewritten some of Baxter's sim controllers to use the `ForwardJointGroupCommandController` because the original implementation was slow to initialize when loading the controllers alongside Gazebo. The likely cause, multiple URDF parses, was discussed in this thread - https://github.com/ros-controls/ros_control/issues/85

adolfo-rt:
You should make the destructor virtual, to allow properly deleting an instance of a derived class through a pointer to base class.

adolfo-rt:
> Tangentially related, I've rewritten some of Baxter's sim controllers to use the ForwardJointGroupCommandController because the original implementation was slow to initialize when loading the controllers alongside Gazebo. The likely cause, multiple URDF parses, was discussed in this thread - ros-controls/ros_control#85.

Using the single-joint controllers for multi-joint control is indeed not very efficient resource-wise. Multiple URDF parsing, more plugin instances, more ROS interfaces (hence more system resources like threads and sockets).

adolfo-rt:
@ipa-fxm, do you have any additional remarks?.

fmessmer:
Nope, no further remarks!
LGTM

mathias-luedtke:
This patch changes the non-RT timing, but that is not critical.
To be honest I am not fond of this solution, since a specialized class is specialized horizontally through inheritance (and function pointers). I would prefer a base class for it.

mathias-luedtke:
Sketch: Base class with virtual(!) init that handles the parsing and virtual(?) update that reads from the buffer. Derive for the different ROS topic types with proper subscriber member and callbacks.

mathias-luedtke:
I have now implemented a base class: https://github.com/ros-controls/ros_controllers/compare/ros-controls:indigo-devel...ipa-mdl:fjgcc_base
In order to use them in arbitrary derived classes I had to move the starting specializations into the header.

What do you think?

rethink-imcmahon:
Thanks for reviewing this, all. Sorry for the delay in response to this. I was out of town last week. 
@adolfo-rt good catch on my lack of virtual destructor (:facepalm:)
@ipa-mdl I like your new base class implementation. One of the major motivations for my initial pull request was to enable `commandCB` with parameters of a type other than `std_msgs::Float64MultiArray` if inheritance is invoked. Then whatever type was accepted in the `commandCB`, it becomes `commandCB`'s responsibility to convert its received message and stuff it into a `Float64MultiArray` for the underlying controllers. For that to work in your implementation, I think `commandCB` would need to be `virtual` and `protected` (along with `sub_command_`). Please correct me if I am misunderstanding, or you know of a cleaner way to accomplish this:
https://github.com/ros-controls/ros_controllers/compare/ros-controls:indigo-devel...ipa-mdl:fjgcc_base#diff-6be8cb644c12239d9b65abd33ebdc1d8R175
https://github.com/ros-controls/ros_controllers/compare/ros-controls:indigo-devel...ipa-mdl:fjgcc_base#diff-6be8cb644c12239d9b65abd33ebdc1d8R174
The `virtual bool init()` part looks great.

mathias-luedtke:
You do not need virtual function to bind to ROS topic subscribers, it does not work like function overloading (different argument types).
For using another topic type, you just have to inherit from ForwardJointGroupCommandControllerBase and implement your subscribers and callbacks like in ForwardJointGroupCommandController.
You could use actions or services instead of topics as well..

```
class MyPositionController: public ForwardJointGroupCommandControllerBase<hardware_interface::PositionJointInterface>
{
public:
  virtual ~MyPositionController() {my_sub_command_.shutdown();}

  virtual bool init(T* hw, ros::NodeHandle &n)
  {
    if(!ForwardJointGroupCommandControllerBase<hardware_interface::PositionJointInterface>::init(hw, n)){
        return false;
    }else{
        my_sub_command_ = n.subscribe<MyType>("another_command", 1, &MyPositionController::myCommandCB, this);
        return true;
    }
  }

private:
  ros::Subscriber my_sub_command_;
  void myCommandCB(const MyType& msg)
  {
    // fill commands buffer somehow
   commands_buffer_.writeFromNonRT(msg->data);
  }
};
```

rethink-imcmahon:
Oh! Interesting, I was misunderstanding / attempting to prevent the unnecessary base `void commandCB(const std_msgs::Float64MultiArrayConstPtr& msg)` from being created, but that doesn't seem like a big deal at all, now that I understand your code better.
Your implementation seems much cleaner than the one I proposed. I'll play around with it for replacing `baxter_simulator`'s controllers and get back to you.

mathias-luedtke:
`commandCB(const std_msgs::Float64MultiArrayConstPtr& msg)` is just not part of the base class :)
It is even possible to inherit from `ForwardJointGroupCommandController` and add a second interface.

rethink-imcmahon:
@ipa-mdl thanks for the implementation! I was able to inherit with the example program as you mentioned after specializing the template in the `init` parameters:

```
virtual bool init(hardware_interface::PositionJointInterface* hw, ros::NodeHandle &n)
```

Also, I had to add `inline` to the three `starting()` specialization definitions in `forward_joint_group_command_controller.h` because I include that file in three of my controllers:

```
template <> inline void ForwardJointGroupCommandControllerBase<hardware_interface::EffortJointInterface>::starting(const ros::Time& time)
template <> inline void ForwardJointGroupCommandControllerBase<hardware_interface::PositionJointInterface>::starting(const ros::Time& time)
template <> inline void ForwardJointGroupCommandControllerBase<hardware_interface::VelocityJointInterface>::starting(const ros::Time& time)
```

Aside from that, I have everything functioning well. Baxter is looking happy as a clam (a clam whose new controllers need tuning :)

happy_baxter_clam

mathias-luedtke:
Oh that one T slipped through..
Great to hear that it works :)
Happy tuning!

I have added the inline specifiers, as an alternative the functions could be moved into a library.
If it's okay for you I will open a PR.

rethink-imcmahon:
@ipa-mdl, sure! I'm happy to use your pull request in place of my (somewhat hacky) solution. Thanks for taking the time to implement it.

PR #185, opened 2015-08-03
F#182 partial trajectory mod

beatrizleon:
Enables the control of partial trajectories. In order to do this, a trajectory now is composed of a vector of vectors of segments (one vector of segments per each joint), so each joint can a vector with different number of segments. 

The changes of code are the minimum I could modify to make it work. If the changes are accepted, we can improve it. Specially, the definition of a segment, as it is only per joint, it does not need a permutation vector, and all its related methods.

All test are passing, but I had to modify some of them to the new implementation. If it is accepted, we can think of new test needed specifically to test partial trajectories.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
@@ -330,10 +330,21 @@ Trajectory initJointTrajectory(const trajectory_msgs::JointTrajectory&       msg
     }
   }
 
+
+
   // Initialize result trajectory: combination of:
   // - Useful segments of currently followed trajectory
   // - Useful segments of new trajectory (contained in ROS message)
   Trajectory result_traj = *(options.current_trajectory);
+
+  //Iterate to all segments to set the new goal handler
+  for (unsigned int joint_id=0; joint_id < joint_names.size();joint_id++)
+  {
+    for (unsigned int segment_id=0; segment_id < result_traj[joint_id].size(); segment_id++)
+    {
+      (result_traj[joint_id])[segment_id].setGoalHandle(options.rt_goal_handle);
+    }
+  }
 
   std::vector<unsigned int> permutation_vector_per_joint(1,0); //refactor this and remove it as it is not needed
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
@@ -246,30 +246,32 @@ Trajectory initJointTrajectory(const trajectory_msgs::JointTrajectory&       msg
   // If unspecified, a trivial map (no permutation) is computed
   const std::vector<std::string> joint_names = has_joint_names ? *(options.joint_names) : msg.joint_names;
 
-  ROS_INFO_STREAM("Joint names size" << joint_names.size());
-
   std::stringstream log_str;
-  log_str.str("");
-  log_str << "Joint names size:" ;
-  for (unsigned int i=0; i < joint_names.size();i++) log_str << joint_names[i] << "\t";
-  log_str << "\n";
-  ROS_INFO_STREAM(log_str.str());
 
-  log_str.str("");
-  log_str << "msg.joint_names size:" ;
-  for (unsigned int i=0; i < msg.joint_names.size();i++) log_str << msg.joint_names[i] << "\t";
-  log_str << "\n";
-  ROS_INFO_STREAM(log_str.str());
 
-  log_str.str("");
-  log_str << "msg.position size:" ;
-  for (unsigned int i=0; i < msg.points.size();i++)
-  {
-    for (unsigned int j=0; j < msg.points[i].positions.size();j++)
-      log_str << msg.points[i].positions[j] << "\t";
-    log_str << "\n";
-  }
-  ROS_INFO_STREAM(log_str.str());
+//  ROS_INFO_STREAM("Joint names size" << joint_names.size());
+//
+//  log_str.str("");
+//  log_str << "Joint names size:" ;
+//  for (unsigned int i=0; i < joint_names.size();i++) log_str << joint_names[i] << "\t";
+//  log_str << "\n";
+//  ROS_INFO_STREAM(log_str.str());
+//
+//  log_str.str("");
+//  log_str << "msg.joint_names size:" ;
+//  for (unsigned int i=0; i < msg.joint_names.size();i++) log_str << msg.joint_names[i] << "\t";
+//  log_str << "\n";
+//  ROS_INFO_STREAM(log_str.str());
+//
+//  log_str.str("");
+//  log_str << "msg.position size:" ;
+//  for (unsigned int i=0; i < msg.points.size();i++)
+//  {
+//    for (unsigned int j=0; j < msg.points[i].positions.size();j++)
+//      log_str << msg.points[i].positions[j] << "\t";
+//    log_str << "\n";
+//  }
+//  ROS_INFO_STREAM(log_str.str());
 
   std::vector<unsigned int> permutation_vector = internal::permutation(msg.joint_names,joint_names);
 
@@ -279,12 +281,12 @@ Trajectory initJointTrajectory(const trajectory_msgs::JointTrajectory&       msg
     return Trajectory();
   }
 
-  //std::stringstream log_str;
-  log_str.str("");
-  log_str << "permutation_vector:" ;
-  for (unsigned int i=0; i < permutation_vector.size();i++) log_str << permutation_vector[i] << "\t";
-  log_str << "\n";
-  ROS_INFO_STREAM(log_str.str());
+//  //std::stringstream log_str;
+//  log_str.str("");
+//  log_str << "permutation_vector:" ;
+//  for (unsigned int i=0; i < permutation_vector.size();i++) log_str << permutation_vector[i] << "\t";
+//  log_str << "\n";
+//  ROS_INFO_STREAM(log_str.str());
 
   // Tolerances to be used in all new segments
   SegmentTolerances<Scalar> tolerances = has_default_tolerances ?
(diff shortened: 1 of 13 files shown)


adolfo-rt:
I did a first pass reading the PR. I'll do a second, more detailed one soon and come back with any comments that might arise. I'd also like to do some experimental testing.

Thanks for putting this together.

beatrizleon:
Thanks @adolfo-rt for reviewing it. I found an error in my branch when a new trajectory was send with an empty vector of accelerations. It is fixed now.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
```diff
@@ -114,6 +113,12 @@ struct InitJointTrajectoryOptions
   ros::Time*                 other_time_base;
 };
 
+template <class Trajectory>
+bool IsNotEmpty(typename Trajectory::value_type trajPerJoint)
+{
+  return trajPerJoint.size()>0;
```
adolfo-rt: `return !trajPerJoint.empty();`

adolfo-rt:
@beatrizleon Update: I'm in the middle of a detailed code review. I'm about halfway done, and I'd like to give some feedback tomorrow.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
```diff
@@ -114,6 +113,12 @@ struct InitJointTrajectoryOptions
   ros::Time*                 other_time_base;
 };
 
+template <class Trajectory>
+bool IsNotEmpty(typename Trajectory::value_type trajPerJoint)
```
adolfo-rt: `isNotEmpty`

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
```diff
@@ -295,93 +313,156 @@ Trajectory initJointTrajectory(const trajectory_msgs::JointTrajectory&       msg
...
-      position_offset = wraparoundOffset(last_curr_state.position,
-                                         first_new_state.position,
-                                         *(options.angle_wraparound));
-      if (position_offset.empty())
+      for (unsigned int segment_id=0; segment_id < result_traj[joint_id].size(); segment_id++)
       {
-        ROS_ERROR("Cannot create trajectory from message. "
-                  "Vector specifying whether joints wrap around has an invalid size.");
-        return Trajectory();
+        (result_traj[joint_id])[segment_id].setGoalHandle(options.rt_goal_handle);
```
adolfo-rt: The new goal handle should not be applied to the trajectory currently being executed. Only to the segments of the new one, and to the bridge segment that continuously transitions between the current and new trajectories, which is done below.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -393,48 +347,100 @@ update(const ros::Time& time, const ros::Duration& period)
...
-  typename Trajectory::const_iterator segment_it = sample(curr_traj, time_data.uptime.toSec(), desired_state_);
-  if (curr_traj.end() == segment_it)
-  {
-    // Non-realtime safe, but should never happen under normal operation
-    ROS_ERROR_NAMED(name_,
-                    "Unexpected error: No trajectory defined at current time. Please contact the package maintainer.");
-    return;
-  }
+  std::vector<bool> successful_joint_traj;
+  successful_joint_traj.resize(joints_.size(), false);
```
adolfo-rt: Dynamic memory allocations are disallowed in the real-time path.

In controllers, computation takes place in two places: 
- `update` method: contains periodic, real-time computation.
- ROS API callbacks: contains asynchronous, non real-time computation.

When implementing the `update` method, one has to be very careful not to break determinism and real-time friendliness.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -393,48 +347,100 @@ update(const ros::Time& time, const ros::Duration& period)
...
 
   // Update current state and state error
   for (unsigned int i = 0; i < joints_.size(); ++i)
   {
+    typename Segment::State desired_joint_state_ = typename Segment::State(1);
```
adolfo-rt: Dynamic allocations are taking place here and in the line below as well.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
adolfo-rt: You can likely make the method real-time safe by pre-allocating resources on `init`, as the sizes are known at initialization time.

I forgot to say above that `starting` and `stopping` are also in the real-time path.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
```diff
@@ -66,9 +66,6 @@ inline std::vector<unsigned int> permutation(const T& t1, const T& t2)
 {
   typedef unsigned int SizeType;
 
-  // Arguments must have the same size
-  if (t1.size() != t2.size()) {return std::vector<SizeType>();}
```
adolfo-rt, line 70: If vectors are no longer to be of the same size, then we are no longer talking about a permutation (rearranging). What's being computed is a map between `t1` indices (implicitly encoded in return vector indices) to `t2` indices. Also, the size of `t1` cannot be larger than that of `t2`. That would be an early optimization.

The function name and documentation should be changed accordingly.

adolfo-rt:
I ran the tests and tested it in simulation. Seems to work as expected (yay).

**Make it opt-in**
One important thought:  Partial joint specification should be opt-in and disabled by default. Rationale:

When a new command specifying a subset of the joints is processed, the unspecified joints keep on executing their previous goal as if nothing happened. I see potential for robots doing unexpected things when different commands that don't know about each other interact. There's always been the risk of controller clients sending bogus commands and crashing the robot or some other unexpected behavior, but the interaction of conflicting commands within a controller was never an issue. I would consider this an advanced behavior that has to be enabled explicitly, as in _I know what I'm doing_.

**Review thoroughness**
Review-wise, I have to confess that the changes are so extensive, that even after having spent a few hours reviewing them, I don't get the feeling of having been thorough enough. It's good that tests are passing, although I don't feel very confident about the completeness of the tolerance checking tests (my fault). When I wrote the tests, that was the only part that was not exercised to the fullest.

To make the test suite more complete and exercise the new feature, I'm thinking of: while the controller executes a trajectory involving all joints (not holding position, but actually moving), send a new command (topic interface) or goal (action interface) that:
- Send trajectories with all waypoints in the past: Nothing gets executed.
- Send trajectories with some waypoints on the past, and some in the future: Some get executed.
- Send trajectories with all waypoints in the future. All get executed, in the future.

I would also like the implementation to be validated on a real-time system.

**More**

The structure and style of the patch can be improved for better code reading and brevity ,but I'll skip those comments for now.

beatrizleon:
@adolfo-rt  Thanks for the review. I am going through your comments now.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
beatrizleon: I thought about implementing it that way at first and have a saved list of active goals, however, I realized that the action server can handle only one goal at a time (http://wiki.ros.org/actionlib). As thechnically the other joints need to finish their trajectories, what I did was prempted the previous goal and creating a new goal (as it was originally), but now the new goal contains the trajectories for all joints, only that each joint has different number of segments according to the partial trajectories. Then the goal will be considered finished when all joints finish their corresponding trajectories. I have tested this behaviour with our hand and it is working as expected.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
beatrizleon, line 70: I agree. What name should be better? maybe "mapping" or "jointMapping"?

beatrizleon, line 70: Fixed it and added a test.

beatrizleon, line 70: I have also removed the permutation variable inside a state and segment as they are now a set for a specific joint.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
adolfo-rt: Still, why are you assigning the new goal handle to the previous trajectory segments?.

One thing that would simplify the review process is some sort of design explanation of what technical problems you've faced, and the proposed solutions. This could go in the PR description. This helps a lot to review large PRs.

A complementary strategy is to have a commit history that tackles each problem individually, so each commit can be reviewed almost independently. Sometimes this is not possible, because of the way individual technical problems are coupled.

beatrizleon: Agree. I fixed it to only assign it to segments after current time.

beatrizleon:
How are you thinking we could implement the opt-in option?

adolfo-rt:
> How are you thinking we could implement the opt-in option? 

With a ROS parameter, say `allow_partial_joint_specification`, or something shorter, but equally descriptive. One would have to explicitly set it to `True` to enable partial joint specification. To enforce the behavior, you would have to do something similar to what was done up to now:
- When using the action inteface. Keep the [check](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L545-L552) that rejected goals early on, but disable it if `allow_partial_joint_specification == True`
- For topic commands, the check that kicked-in was [this one](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h#L248-L252). Same story, keep the check, but disable it if `allow_partial_joint_specification == True`.

Note that these checks relied on the `permutation` function, which actually checked for permutations (not mappings). As an alternative, you can call your new mapping function and additionally check that the two input vectors have the same size.

beatrizleon:
Added the ros param. I named it: "allow_partial_joints_goal" but it is opened to suggestions :)

beatrizleon:
@adolfo-rt I added a set of tests (very similar to the existing ones) to test partial trajectories which included the ones you mentioned in your previous comment.

We do not have a real time system, then I would not be able to test that.

Appart from that I think I have addressed all your comments. Let me know what you think.

ugocupcic:
@adolfo-rt any news on this? We're kind of waiting for that functionality at Shadow :)

Cheers,

Ugo

sachinchitta:
@ugocupcic - @adolfo-rt did mention on the mailing list that he would be offline for 3 weeks starting Sept. 8.

ugocupcic:
Ah! OK thanks @sachinchitta :)

toliver:
@adolfo-rt any new on this? If there is anything else that we could improve to get it merged please let us know.

v4hn:
It has been nearly three month since @adolfo-rt was last active on github.
Is there anyone else who could look at that?
It was mentioned that this should be in the kinetic release in https://github.com/ros-controls/ros_control/issues/234 . Is there any reason why this should not be merged into indigo as well?
It would help me quite a lot if it were.

bmagyar:
I made the releases for Kinetic and if all goes fine (since it's a well discussed issue) I will merge it to kinetic-devel soon. I'm not sure though if these changes should go to Indigo&Jade as well, so my approach so far was to merge all new stuff to kinetic-devel, and wait for some of our sleeping agents to appear so we can discuss propagating them back to indigo&jade.

toliver:
We are going to try to rebase this branch and open a new PR against kinetic-devel. Hopefully this week or the next.

@bmagyar  please let us know if we should also keep this PR open just in case it's getting merged into indigo later on.

bmagyar:
Great Toni!

For the rebase may I ask you to also either squash the commits into one huge commit or into bigger chunks so that the "work in progress" type of commits are not reflected in the proposed changes? It's better if we track feature additions rather than the entire development processes.

Once the new PR is open, this can be closed. I am personally ok to push through all the new stuff for ROS Kinetic but if you'd like back-support for Indigo we will have to gather some more people and get a lot more feedback. If you wish to have the Kinetic features in older distros, feel free to open an issue and let's discuss them there.

bmagyar:
Closing since we merged it to Kinetic.

PR #187, opened 2015-08-18
Base class for ForwardJointGroupCommandController

mathias-luedtke:
As discussed in #184 this patch introduces a base class for ForwardJointGroupCommandController that allows subclassing for various ROS interface types.
- added virtual specifiers to outline the (implicitly) virtual member functions
- all base class data members are protected (instead of public)

I had to move the starting function definitions into the header, but migrated them to template specializations. If needed the implementation could be moved into a library.

### DIFF
--- a/forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
+++ b/forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
@@ -55,9 +55,7 @@ namespace forward_command_controller
 /**
  * \brief Forward command controller for a set of joints.
  *
- * This class forwards the command signal down to a set of joints.
- * Command signal and joint hardware interface are of the same type, e.g. effort commands for an effort-controlled
- * joint.
+ * This class updates foint handles from a RealtimeBuffer
  *
  * \tparam T Type implementing the JointCommandInterface.
  *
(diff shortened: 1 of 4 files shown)


forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
```diff
@@ -106,33 +99,87 @@ class ForwardJointGroupCommandController: public controller_interface::Controlle
...
   }
 
+protected:
   std::vector< std::string > joint_names_;
   std::vector< hardware_interface::JointHandle > joints_;
   realtime_tools::RealtimeBuffer<std::vector<double> > commands_buffer_;
   unsigned int n_joints_;
+};
+
+template <> inline void ForwardJointGroupCommandControllerBase<hardware_interface::EffortJointInterface>::starting(const ros::Time& time)
```
rethink-imcmahon, line 120: I think moving these three inline functions back into their cpp files makes sense:

```
effort_controllers/src/joint_group_effort_controller.cpp
position_controllers/src/joint_group_position_controller.cpp
velocity_controllers/src/joint_group_velocity_controller.cpp
```

The benefit here is that having to include their respective `.h` files protects us against prematurely specialization the `starting()` functions. For instance, I think we need 

```
effort_controllers::JointGroupPositionController
effort_controllers::JointGroupVelocityController
```

which would be the `group` analogs of
[effort_controllers::JointPositionController](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/effort_controllers/include/effort_controllers/joint_position_controller.h)
[effort_controllers::JointVelocityController](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/effort_controllers/include/effort_controllers/joint_velocity_controller.h)

both of which would inherit from `ForwardJointGroupCommandControllerBase<hardware_interface::EffortJointInterface>` but will likely require different `starting(const ros::Time& time)` implementations from `effort_controllers::JointGroupEffortController::starting()` (ie. the above line)

rethink-imcmahon:
Overall, a big +1 in favor of this addition. The `ForwardJointGroupCommandControllerBase` increases our flexibility in creating new forward group controllers.

forward_command_controller/include/forward_command_controller/forward_joint_group_command_controller.h
mathias-luedtke, line 120: Moving them into `forward_command_controller` allows to reuse them as well.
Most users will be fine with the default for `starting()`, if not they can reimplement it, it is (and has been) virtual.

Your examples

```
effort_controllers::JointGroupPositionController
effort_controllers::JointGroupVelocityController
```

are no forward controllers and should not use this base class.

bmagyar:
Apologies for the enormous delay on this one. @ipa-mdl I've retargeted the MR to melodic, could you please push a dummy commit trigger a new build setup on travis? I'm afraid restarting it will only get us more of the indigo pipeline.

PR #189, opened 2015-08-19
[rqt_joint_trajectory_controller] Many small improvements

adolfo-rt:
This PR addresses a number of small but annoying issues that were affecting `rqt_joint_trajectory_controller`. They should improve the plugin's usability and stability.

### DIFF
--- a/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_limits_urdf.py
+++ b/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_limits_urdf.py
@@ -33,12 +33,15 @@ def get_joint_limits(key='robot_description', use_smallest_joint_limits=True):
             if jtype == 'fixed':
                 continue
             name = child.getAttribute('name')
+            try:
+                limit = child.getElementsByTagName('limit')[0]
+            except:
+                continue
             if jtype == 'continuous':
                 minval = -pi
                 maxval = pi
             else:
                 try:
-                    limit = child.getElementsByTagName('limit')[0]
                     minval = float(limit.getAttribute('lower'))
                     maxval = float(limit.getAttribute('upper'))
                 except:
--- a/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_limits_urdf.py
+++ b/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_limits_urdf.py
@@ -49,8 +49,7 @@ def get_joint_limits(key='robot_description', use_smallest_joint_limits=True):
             try:
                 maxvel = float(limit.getAttribute('velocity'))
             except:
-                rospy.logwarn("Joint '%s' has no maximum velocity specification in URDF. Defaulting to 1.0.", name)
-                maxvel = 1.0
+                continue
             safety_tags = child.getElementsByTagName('safety_controller')
             if use_small and len(safety_tags) == 1:
                 tag = safety_tags[0]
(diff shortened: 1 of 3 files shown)


bmagyar:
:+1:

rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -33,12 +33,15 @@ def get_joint_limits(key='robot_description', use_smallest_joint_limits=True):
             if jtype == 'fixed':
                 continue
             name = child.getAttribute('name')
+            try:
+                limit = child.getElementsByTagName('limit')[0]
+            except:
+                continue
```
mathias-luedtke, line 39: This skips all continuous joints without limit tag (which is not required for them)

adolfo-rt, line 39: We need the velocity limit to perform velocity scaling (bottom-most slider in plugin). On a `<limit>` [URDF tag](http://wiki.ros.org/urdf/XML/joint), effort and velocity attributes are required, position limits are optional.

mathias-luedtke, line 39: That makes sense. A warning would be nice for non-fixed joints that were skipped.

adolfo-rt, line 39: You know I had that exact same though, and actually [pushed that change](https://github.com/ros-controls/ros_controllers/commit/3e5b9c034c5bc7b46ab411086288ac9a521cf8b1) before this PR. But it turns out that because we parse the entire URDF (we don't know which joints will end up being controlled by a JTC), this potentially generates lots of spam, from joints that might never be the business of this plugin.

I'm open to better suggestions, but my initial, naïve solution didn't work out well.

I'll see if I can find a non-spamming way of alerting about controllers that are being left out because of missing joint limits.

mathias-luedtke, line 39: I see two options:
1. just do not issue warnings for revolute and prismatic joints, might spam though
2. defer the skipping to https://github.com/ros-controls/ros_controllers/blob/rqt-jtc-improvements/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_trajectory_controller.py#L246

Maybe it is not worth the effort.

adolfo-rt, line 39: I admit that it might strike as odd to a user if a controller is running, but does not show up in the combo box.

Ideally, one would have to keep a list of all controllers that are running, but don't have a valid limits spec (hence don't show up in the combo), and print a warning _once_ every time one of these is found. I say once, because the plugin is [periodically checking for running controllers](https://github.com/ros-controls/ros_controllers/blob/rqt-jtc-improvements/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_trajectory_controller.py#L174-L177), so the combos are in sync with the system.

It's doable, just more work.

adolfo-rt, line 39: I've opened #194 to track this extra feature, as I don't have the resources to tackle it in this maintenance round.

PR #192, opened 2015-08-24
rqt_jtc: Clear controllers combo on cm change

### LINKED ISSUE #191: rqt_joint_trajectory_controller does not update joint list if controller names are identical
In the case two controller manager each in a different namespace exhibit each a controller with same name (but controlling different joints for instance left and right arm), the switching to the second controller on the second controller manager does not update the joint list and hence forbid to control the second controller's targets.

Work around could be to name the controller differently. But we already use namespace to run two separate robots (two control loops = two controller manager). 

```
/left/controller_mananger
/right/controller_manager 
```

running an `arm_joint_trajectory_controller`
calling them `left_arm_joint_trajectory_controller` would be redundand information and adding complexity to the deployment and usage.

How to reproduce. Have two controller managers on two namespaces and load a joint_trajectory_controller on each of them with the same name

open rqt_joint_trajectory_controller, 
select first controller_manager, list of controller is updated showing one controller.
select the controller, joint list is updated
select second controller_manager, list of controller remains the same (controller name being the same)
select the controller, joint list is NOT updated, probably because the on_change is not triggered on the controller combobox due to same name. 
Works if there are more than one controller in the controller manager. Or other controller managers that do not show the same controller name.

adolfo-rt:
Fixes #191.

Clear the list of running joint trajectory controllers when the
controller manager selection changes. This prevents potential conflicts when
multiple controller managers have controllers with the same names.

### DIFF
--- a/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_trajectory_controller.py
+++ b/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_trajectory_controller.py
@@ -237,6 +237,10 @@ class JointTrajectoryController(Plugin):
         self._cm_ns = cm_ns
         if cm_ns:
             self._list_controllers = ControllerLister(cm_ns)
+            # NOTE: Clear below is important, as different controller managers
+            # might have controllers with the same name but different
+            # configurations. Clearing forces controller re-discovery
+            self._widget.jtc_combo.clear()
             self._update_jtc_list()
         else:
             self._list_controllers = None


guihomework:
This commit fixes the issue. Hope it does not get annoying to others because it requires to reselect the controller from the cleared list but this is safer this way imho.

thanks

adolfo-rt:
> This commit fixes the issue. Hope it does not get annoying to others because it requires to reselect the controller from the cleared list but this is safer this way imho.

I consider the previous behavior a bug, so it's actually correct now. Thanks again.

PR #193, opened 2015-08-24
rqt_jtc: Add vertical scrollbar to joints list

### LINKED ISSUE #190: rqt_joint_trajectory_controller sliders access for a lot of joints
When a controller has a lot of joints (24 for the Shadow robot hand), the size of the sliders is too large and cannot be seen fully on a screen with HD resolution. 
Is it possible to include a slider to go up and down on the list when the lists gets out of screen ?

adolfo-rt:
Fixes #190.

### DIFF
--- a/rqt_joint_trajectory_controller/resource/joint_trajectory_controller.ui
+++ b/rqt_joint_trajectory_controller/resource/joint_trajectory_controller.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>518</width>
-    <height>741</height>
+    <width>336</width>
+    <height>317</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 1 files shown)


guihomework:
@adolfo-rt , thanks for the fix.  I can now access all the hand joints. 
Here is how it looks like (right side compared to before).
I just find strange the plugin does not occupy all the free space at the bottom.
comparison_rqt_jnt_traj_ctrl

adolfo-rt:
> I just find strange the plugin does not occupy all the free space at the bottom.

Good observation. I now removed a vertical spacer that was at the bottom of the plugin, which I hope resolves this last issue. By adding the vertical scroll area and changing the resizing policy of some elements, I believe the spacer is no longer necessary.

adolfo-rt:
Please try the latest changes directly from `indigo-devel`, as I'm merging.

guihomework:
Now it is good. 
rqt_jtc_vertical_scrollbar

Thanks for the quick fix on that useful tool.

PR #199, opened 2015-10-06
add check for multiple publishers on cmd_vel

Karsten1987:
prevent multiple subscriptions to the command topic in order to prevent any ambiguity of who is controlling the robot.
In case of multiple publishers the robot will brake.

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
@@ -138,6 +138,9 @@ namespace diff_drive_controller{
     /// Timeout to consider cmd_vel commands old:
     double cmd_vel_timeout_;
 
+    /// Whether to allow multiple publishers on cmd_vel topic or not:
+    bool allow_multiple_cmd_vel_publishers_;
+
     /// Frame to use for the robot base:
     std::string base_frame_id_;
 
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -115,6 +115,7 @@ namespace diff_drive_controller{
     , wheel_separation_multiplier_(1.0)
     , wheel_radius_multiplier_(1.0)
     , cmd_vel_timeout_(0.5)
+    , allow_multiple_cmd_vel_publishers_(true)
     , base_frame_id_("base_link")
     , enable_odom_tf_(true)
     , wheel_joints_size_(0)
@@ -180,6 +181,10 @@ namespace diff_drive_controller{
     controller_nh.param("cmd_vel_timeout", cmd_vel_timeout_, cmd_vel_timeout_);
     ROS_INFO_STREAM_NAMED(name_, "Velocity commands will be considered old if they are older than "
                           << cmd_vel_timeout_ << "s.");
+
+    controller_nh.param("allow_multiple_cmd_vel_publishers", allow_multiple_cmd_vel_publishers_, allow_multiple_cmd_vel_publishers_);
+    ROS_INFO_STREAM_NAMED(name_, "Allow mutiple cmd_vel publishers is "
+                          << (allow_multiple_cmd_vel_publishers_?"enabled":"disabled"));
 
     controller_nh.param("base_frame_id", base_frame_id_, base_frame_id_);
     ROS_INFO_STREAM_NAMED(name_, "Base frame_id set to " << base_frame_id_);
@@ -373,6 +378,15 @@ namespace diff_drive_controller{
   {
     if (isRunning())
     {
+      // check that we don't have multiple publishers on the command topic
+      if (!allow_multiple_cmd_vel_publishers_ && sub_command_.getNumPublishers() > 1)
+      {
+        ROS_ERROR_STREAM_THROTTLE_NAMED(1.0, name_, "Detected " << sub_command_.getNumPublishers()
+            << " publishers. Only 1 publisher is allowed. Going to brake.");
+        brake();
+        return;
+      }
+
       command_struct_.ang   = command.angular.z;
       command_struct_.lin   = command.linear.x;
       command_struct_.stamp = ros::Time::now();


bmagyar:
First I wanted to make a comment to use the actual controller name instead of "diff_drive_controller"  in the error message but it's a ROS_ERROR_NAMED so it's perfectly fine.

You may want to fix the english a bit: "publisher" -> "publishers".

Other than this, it looks good to me.
Should there be a param for this to enable the functionality that would be disabled by default to keep current behaviour? 

I'm tilting towards having it always on, and no param.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -373,6 +373,14 @@ namespace diff_drive_controller{
   {
     if (isRunning())
     {
+      // check that we don't have multiple publishers on the command topic
+      if (sub_command_.getNumPublishers() > 1)
+      {
+        ROS_ERROR_NAMED(name_, "Detected multiple publishers for diff_drive_controller. Going to brake.");
```
efernandez: I think this should be throttled, maybe at 1s period.

I'd also make the message a little more informative, but this is up to you.

`ROS_ERROR_STREAM_THROTTLE_NAMED(1.0, name_, "Detected " << sub_command_.getNumPublishers() << " publishers. Only 1 publisher is allowed. Going to brake.");`

efernandez:
LGTM, apart from my minor comment.

Regarding @bmagyar question about having an option, I'm also leaning to change the current behaviour, because I think in this case there's no 'current behaviour', just an unhandled situation.

The only thing that worries me is that the user might not know why the robot doesn't move unless it sees the terminal, but IMO this shows that the diff_drive_controller needs to provide diagnostics and currently doesn't. We might file an issue to implement this in the future. This PR is fine to go as it is.

Karsten1987:
Any news on this one?

@efernandez It's a valid point that the user wouldn't see the robot moving. But imaging that there are two publishers at the same time, this would either way lead to confusing results, no? There is basically no default behavior for this.

efernandez:
@Karsten1987 I agree. As I said, there was no behaviour defined in that case, so your PR is fine.
Thanks for considering my comment.

:+1: for merging this.

@Karsten1987 I don't have merge permissions, you have to bug someone else to have this merge ;)

adolfo-rt:
I'm unsure if a single-publisher policy should be enforced here. You could have multiple command publishers and have a higher level logic that enforces that only one is publishing at any time. With this PR, such usecases would have to change to adding/removing publishers so only one exists. Thoughts?.

efernandez:
@adolfo-rt has raised a good point. However I think that usually there's a single input to the controller and it's good to enforce that because it's safer.

In a practical case, I think it's more common to use a multiplexer for many inputs and select a single one every time, like http://wiki.ros.org/twist_mux, so the controller has a single input.

However, the other case is also possible, so I'd suggest to add a parameter, like `allow_multiple_cmd_vel_publishers` or `allow_multiple_publishers`. IMHO it should default to `False`, as it's safer, although that changes the current behavior in this case; if more than one publisher is detected we can mention this parameter on the log message.

We could also have `max_publishers` to enforce a given maximum number of publishers, but I think it looks overcomplicated.

bmagyar:
@adolfo-rt 's point is pretty good. A multiplexer is a good tool, but if someone wants to implement a similar effect with entangled components, then so be it. 

I hereby change my above opinion on the default of this param. Blocking when there are multiple publishers should be off by default, this also keeps our current practice of keeping the simple behaviour default, and extras can be turned on separately. 

As for name, I'd prefer `allow_multiple_cmd_vel_publishers`.

One thing that I think would be nice is to throw a warning with some throttle (1/2/3/4/5s?) even when the blocking is off.

mathias-luedtke:
>  A multiplexer is a good tool, but if someone wants to implement a similar effect with entangled components, then so be it.

In addition a multiplexer increases the latency, so there are reasons to avoid it in some cases.

If we really want to restrict the number of publishers connected to a **controller** then I would prefer a generic class that makes this feature available for other controllers as well.
With consistent parameter names and the corresponding logic.

> Blocking when there are multiple publishers should be off by default, this also keeps our current practice of keeping the simple behaviour default, and extras can be turned on separately.

+1

efernandez:
@ipa-mdl are you suggesting something like a sub-class of `ros::Subscriber` with a maximum number of publishers?
Note that this PR only rejects the messages if more than one publishers is publishing on the topic. It doesn't "restrict the number of publishers connected", as you're saying. For that I think we'd have to touch the logic in https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/src/libros/subscription.cpp#L794; correct me if I'm wrong, since I don't know that code too much.

BTW, it looks like checking the number of publishers requires to iterate over all the topics available on the system at that time: https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/src/libros/topic_manager.cpp#L893
I guess it shouldn't be a problem, but I was surprised a `map` isn't used.

@bmagyar If blocking is off, I think throttling a warning would end up being annoying.

IMHO for this particular case a `allow_multiple_cmd_vel_publishers` param + the check for the number of publishers looks like the most reasonable solution.

mathias-luedtke:
@efernandez:
That would be a nice quality of service use case for ROS2 ;)
I just meant to factor reject logic out of the controller. Perhaps a `RealtimeSubscriber` class would make sense, since this pattern is used in other controllers as well.

>  but I was surprised a map isn't used.

For small numbers a map might introduce more overhead than it gains performance..
http://www.boost.org/doc/libs/1_59_0/doc/html/container/non_standard_containers.html#container.non_standard_containers.flat_xxx

efernandez:
@ipa-mdl Yes, something like `RealtimeSubscriber` with a `unsigned int max_publishers_` attribute (where `0` could mean unbounded number of publishers) sounds good; although in this case the functionality isn't for RT, I can't think of another name... maybe `BoundedSubscriber`?!
But I'd be happy with this PR w/o any extension to the `RealtimeSubscriber`, as that will bring the risk of having this PR open for a long time.

Thanks for the link!
I was just pointing out that the search is `O(N)` in the worst case for the number of publishers; curiously, for the number of subscribers a LUT is used. Anyway, that was just a side comment.

mathias-luedtke:
@efernandez: I was thinking of bundling a Subscriber to a RealtimeBuffer.
But this is beyond the scope of this pull request. However, if we decide that bounded publisher features is reasonable for a controller, we should at least open an issue to request this feature for all topic-based controllers. But this no blocker for this PR :)

efernandez:
I :+1: for having the ability to limit (or emit a warning) the maximum number of publishers on a topic.
Not sure though if it should be specific to ROS controllers, it's probably something I'd like to have for any node/topic. Anyway, I think we should definitely create an issue for it, either here (considering that controllers are usually critical software) or in `ros_comm` (and let's see what others think about that).

Karsten1987:
I basically agree with everything said here, however IMHO this conversation drifted a bit from the initial PR. What are the concrete next steps here?

bmagyar:
I propose to merge this one and open an issue to address this possible
feature addition in the scope of each member of ros_controllers.

mathias-luedtke:
We can merge it if the new behaviour is turned per default.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -373,6 +373,15 @@ namespace diff_drive_controller{
   {
     if (isRunning())
     {
+      // check that we don't have multiple publishers on the command topic
+      if (sub_command_.getNumPublishers() > 1)
```
efernandez: @Karsten1987 Could you add the `allow_multiple_cmd_vel_publishers` param?

`if (!allow_multiple_cmd_vel_publishers_ && sub_command_.getNumPublishers() > 1)`

and in the constructor + `init` add the code to read the param `allow_multiple_cmd_vel_publishers`, which I think at the end we agreed on having it set to `true` by default, to keep the previous behavior.

Karsten1987:
I applied the changes requested by @efernandez, rebased and squashed the previous commits.

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
```diff
@@ -138,6 +138,9 @@ namespace diff_drive_controller{
     /// Timeout to consider cmd_vel commands old:
     double cmd_vel_timeout_;
 
+    /// Whether to allow multiple publishers on cmd_vel topic or not:
+    double allow_multiple_cmd_vel_publishers_;
```
mathias-luedtke: Should be bool? ;)

efernandez: Yes, that should be `bool`

efernandez:
Someone who can merge, this looks ready to be merged! :grinning:

bmagyar:
:+1: 

Thank you @Karsten1987 for the patch!

PR #209, opened 2016-02-13
Config init fix

mikepurvis:
@efernandez This fixes the bug we were looking at today; turns out the config structure isn't initialized to the defaults automatically— hilarious.


mikepurvis:
Whoops, wrong branch, disregard.

Issue #173, opened 2015-04-14
joint_trajectory_controller does not publish feedback via actionlib


mkoval:
The `FollowJointTrajectory` action can periodically return feedback about the actual state, desired state, and controller error. It would be nice if the `joint_trajectory_controller` published these messages.

This message is almost identical to the `control_msgs/JointTrajectoryControllerState` message that is already being published by the controller. I am willing to implement this feature myself and send a pull request, but I need some guidance on how to make this real-time safe.

I assume that I can't directly call `ServerGoalHandle::publishFeedback` because it is not real-time safe. It is also not wrapped in `RealtimeServerGoalHandle`. The most obvious answer is to add a `publishFeedback` method to `RealtimeServerGoalHandle::publishFeedback` and call it from `updateState` in the controller.

Is there any reason why `publishFeedback` isn't already implemented here?

adolfo-rt:
[Related post](https://groups.google.com/forum/#!topic/ros-sig-robot-control/pYAdBNIgsYs). Thanks for making it explicit as an issue.

I'll try to take a look at how to make the implementation in the coming days. In the meantime, you can take a look at how goal messages are published through the [RealtimeServerGoalHandle](https://github.com/ros-controls/realtime_tools/blob/indigo-devel/include/realtime_tools/realtime_server_goal_handle.h).

dcconner:
Note I'm also looking into ServerGoalHandle with #174.

One useful distinction with the published state that I might suggest is to make the feedback tied to changes in waypoint segments.  That way we are not just giving redundant information.

mkoval:
I've been swamped, so I haven't had a chance to look at `ServerGoalHandle` much.

> One useful distinction with the published state that I might suggest is to make the feedback tied to changes in waypoint segments. That way we are not just giving redundant information.

I would strongly prefer to get feedback at `state_publish_rate`. Waypoints are just a mechanism for specifying a `q(t)` spline. It would be undesirable to get different feedback messages from two trajectories that happen to have different parameterizations, but are otherwise identical.

It is frustrating that `joint_trajectory_controller` implements an `actionlib` interface, but I have to subscribe to an unrelated topic to get feedback. This also makes it difficult to associate feedback with a particular `JointTrajectoryControllerState` message with an actionlib goal.

miguelprada:
Has anyone actually made any progress on this?

[...]

PR #213, opened 2016-04-28
Write feedback for the RealtimeServerGoalHandle to publish.

miguelprada:
This builds upon ros-controls/realtime_tools#18 to make the `joint_trajectory_controller` publish action feedback on the actionlib topic, as discussed in ros-controls/ros_controllers#173.

It re-uses the non-realtime publishing capability of RealtimeServerGoalHandle to also deliver the action feedback using the same rate with which the action status is published (i.e. `action_monitor_rate`), instead of publishing at the same rate the state messages are published (i.e. `state_publish_rate`).


bmagyar:
I cannot merge this as is because it's not based on the latest kinetic-devel. I still intend to, so I may hand-pick the changes myself.

bmagyar:
Rebase done & merged on #219, closing this one

PR #214, opened 2016-04-30
Add antiwindup to get/setGains methods of jointPosition and jointVelocity controllers

paulbovbel:
See also: https://github.com/ros-controls/control_toolbox/pull/38


bmagyar:
ABI doesn't need to be compatible here since there's nothing released into Kinetic yet, though it may make sense to keep it the way it is now.
What do you think @paulbovbel ?

All tests seem fine to merge, I'm close to merging.

paulbovbel:
LGTM as is, preventing any API changes for kinetic unless users opt into antiwindup.

bmagyar:
:+1: 
Merging...

Issue #216, opened 2016-05-09
Create separate GitHub repositories for each package


alainsanguinetti:
Hello,

I am trying to use the joint_trajectory_controller in my project. This means that I have a fully set-up catkin workspace where I have my whole project, which is kind of a "ros release" that I want to deploy to several machines and that I track with git.
However, there is a problem in joint_trajectory_controller_impl.h, namely the definition of non-static non-inline functions in a header file, that I would like to fix and hopefully merge into the ROS official packages. 
My problem is, I don't know how I can simply clone that particular package, fix it and issue a pull request without having to pull the entire ros_controllers repository into a separate catkin folder, catkin_make install everything, switch workspaces, etc
I suggest that every package is a repository,(because every package is independent right ?) and that ros_controllers be the repository of the metapackage only. 

Would that make sense ?

bmagyar:
Hi,
I'm not sure I get your use-case, so help me out a bit more by starting with the usual things:
ROS version, operating system & version, way you set things up, error messages.
By the way:
If you need to use the joint_trajectory_controller, why don't you just install the binary package?
What's the exact issue with compiling controllers? I have done custom versions of a few at times and had no issues.
Carrying catkin-workspaces around can be done by using rosinstall files.

alainsanguinetti:
Hello, Thanks for your help and sorry for the poor quality of the PR, I'll do better next time!

The use case for the bug I am trying to fix is very simple: 
system is ROS Indigo on Ubuntu 14.04.4 64bits
the problem happens in this configuration:
- code1.h imports joint_trajectory_controller.h, definitions in code1.cpp
- code2.h imports code1.h, definitions in code2.cpp

when code1 and code2 get compiled to code1.o and code2.o, everything is OK, but link edition fails because the functions defined in joint_trajectory_controller_impl.h without inline or static keyword have been defined in different .o 

---

For the rationale behind providing packages as different repositories:

I would love to include in my project joint_trajectory_controller as a git submodule to be able of modifying it while still keeping track of improvements from the official codebase.
Also, I think it makes more sense to have one repository for every independently released package because since they are shipped separately, it should be possible to modify one independently of the other packages. 
Then, I don't see what does this repository map to in a catkin_workspace : it is not catkin_workspace(the source are not in a src/ folder) nor a package (not the structure of a package) so it is very difficult to work on it from an existing setup.
[...]

PR #217, opened 2016-05-09
Indigo devel

alainsanguinetti:
joint_trajectory_controller: 
Added "static inline" in front of functions defined in joint_trajectory_controller_impl.h
It caused the link edition to fail because of multiple definitions of those functions.

### DIFF
--- a/joint_trajectory_controller/CHANGELOG.rst
+++ b/joint_trajectory_controller/CHANGELOG.rst
@@ -2,11 +2,11 @@
 Changelog for package joint_trajectory_controller
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


bmagyar:
This PR is linked to the issues reported on #216. 

My general comments to this PR:
Most of your commits don't carry any addition, we are not going to merge changes such as "Testing user settings", merging things onto your own branch and updating package.xml and changelogs by hand, leave that to the maintainers ;)
There is a single commit that could be added to the repository, pull requests are expected to only carry the minimum amount of changes needed to patch an issue.
Though indigo is not EOL, at the moment we are only accepting changes to Jade and Kinetic. 
For this reason I'm going to close this PR, once we agreed that changes are necessary, I will ask you to open a new pull request against one of those branches. (Github doesn't allow to edit the target of a pull request so a new one needs to be created...)

Let's talk over at #216

Issue #221, opened 2016-05-26
Failing Travis tests 


bmagyar:
...which pass on the ROS buildfarm.
https://travis-ci.org/ros-controls/ros_controllers/builds/133042066#L7109

efernandez:
This is the error:

``` bash
rqt_plot: cannot connect to X server 
rqt_plot: cannot connect to X server 
[ERROR] [1464248901.774157876]: Joints on incoming goal don't match the controller joints.
[INFO] [WallTime: 1464248980.070233] Shutting down spawner. Stopping and unloading controllers...
testjoint_trajectory_controller_test ... ERROR!
ERROR: max time [80.0s] allotted for test [joint_trajectory_controller_test] of type [joint_trajectory_controller/joint_trajectory_controller_test]
  File "/usr/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/opt/ros/jade/lib/python2.7/dist-packages/rostest/runner.py", line 148, in fn
    self.test_parent.run_test(test)
  File "/opt/ros/jade/lib/python2.7/dist-packages/rostest/rostest_parent.py", line 132, in run_test
    return self.runner.run_test(test)
  File "/opt/ros/jade/lib/python2.7/dist-packages/roslaunch/launch.py", line 681, in run_test
    (test.time_limit, test.test_name, test.package, test.type))
```

which looks like an issue related with either a
1. Timeout (80s now)
2. No X display for rqt_plot

PR #222, opened 2016-05-26
Increase jtc test time limit to 120s (from 80s)

efernandez:
This addresses https://github.com/ros-controls/ros_controllers/issues/221 by adding `xvfb` to the Travis configuration `before_script` block, as explained in https://docs.travis-ci.com/user/gui-and-headless-browsers/

If the test still fails, I'll also increase the timeout.

FYI @bmagyar

### DIFF
--- a/joint_trajectory_controller/test/joint_trajectory_controller.test
+++ b/joint_trajectory_controller/test/joint_trajectory_controller.test
@@ -31,5 +31,5 @@
   <test test-name="joint_trajectory_controller_test"
         pkg="joint_trajectory_controller"
         type="joint_trajectory_controller_test"
-        time-limit="80.0"/>
+        time-limit="120.0"/>
 </launch>


efernandez:
It seems the documentation I was using is wrong:

``` bash
$ sh -e /etc/init.d/xvfb start
sh: 0: Can't open /etc/init.d/xvfb
```

efernandez:
Installing `xvfb` doesn't make any difference because it doesn't put anything in `/etc/init.d`.

Not sure why is that, but I'm going to comment it out and try increase the time limit instead.

efernandez:
Well, now it's another test that fails!!!

``` bash
[joint_state_controller.rosunit-joint_state_controller_test/publishOk][FAILURE]-
/home/travis/ros/ws_ros_controllers/src/ros_controllers/joint_state_controller/test/joint_state_controller_test.cpp:139
The difference between real_pub_rate and pub_rate is 3.4126115780579189, which exceeds 0.05 * pub_rate, where
real_pub_rate evaluates to 6.5873884219420811,
pub_rate evaluates to 10, and
0.05 * pub_rate evaluates to 0.5.
```

This looks clearly like a time dependant test, so I think the error will be intermittent until we make it more robust to the time variations. Tests that depend on time are highly discourage though.

I'm going to run the job test again.

@bmagyar Thoughts? Merge?

efernandez:
BTW, I still see

``` bash
rqt_plot: cannot connect to X server :99.0
```

So the problem with that one is NOT the X display, just the timeout. @bmagyar Are you OK with just removing the commits to add `xvfb` and just change the time limit?

efernandez:
Updated to test only with the 120 time limit. The `xvfb` didn't work at all, so I better remove it.

bmagyar:
Seems like we are happy now. Let's hope it will stay stable... maybe we should start thinking on non-time dependant tests

efernandez:
Yes. At least, time dependant tests should be unit tests, not rostests, to alleviate the intermittent failures.

But all happy for now... do you mind to cherry-pick this into `kinetic-devel`?
I'll take care of adding test support to `kinetic-devel` asap.

I think I also have to update all the repos with the same `.travis.yml` config, right? :(

bmagyar:
@efernandez make sure to use Xenial for Kinetic!
I can see `trusty` occur in many places in the config file despite that it's "macro-d" in the first few lines

PR #223, opened 2016-05-26
Increase jtc. test time limit to 120s

efernandez:
This cherry-picks https://github.com/ros-controls/ros_controllers/pull/222 into `kinetic-devel`.

@bmagyar for review

PR #211, opened 2016-04-19
Add support to check tests on Travis CI

efernandez:
First temptative for running tests on Travis CI for ROS Control.

I'll ping someone for review once it's working.

PR #224, opened 2016-05-26
Update .travis.yml for Kinetic

efernandez:
This fixes the Travis configuration:
- Now it builds the code with kinetic, not hydro
- Now it runs the tests

https://github.com/ros-controls/ros_controllers/pull/223 should be re-run after this gets merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,48 +1,81 @@
 # Travis Continuous Integration Configuration File For ROS Control Projects
-# Author: Dave Coleman
-language:
-  - cpp
-  - python
-python:
-  - "2.7"
+# Author: Dave Coleman, Enrique Fernandez
+sudo: required
+dist: xenial
+language: generic
 compiler:
   - gcc
+env:
+  global:
+    - ROS_DISTRO=kinetic
+
 notifications:
   email:
     recipients:
       - davetcoleman@gmail.com
-      - adolfo.rodriguez@pal-robotics.com
+      - adolfo.rodriguez@intermodalics.eu
+      - enrique.fernandez.perdomo@gmail.com
+      - bence.magyar.robotics@gmail.com
     on_success: change #[always|never|change] # default: change
     on_failure: change #[always|never|change] # default: always
-before_install: # Use this to prepare the system to install prerequisites or dependencies
-  # Define some config vars
-  - export ROS_DISTRO=hydro
+
+before_install:
   - export CI_SOURCE_PATH=$(pwd)
   - export REPOSITORY_NAME=${PWD##*/}
   - echo "Testing branch $TRAVIS_BRANCH of $REPOSITORY_NAME"
-  - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu precise main" > /etc/apt/sources.list.d/ros-latest.list'
+  # Add ROS repositories
+  - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -cs) main" >
+    /etc/apt/sources.list.d/ros-latest.list'
   - wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
   - sudo apt-get update -qq
-  - sudo apt-get install -qq -y python-catkin-pkg python-rosdep python-wstool ros-$ROS_DISTRO-catkin ros-$ROS_DISTRO-ros
-  # Setup rosdep
-  - sudo rosdep init
+  # Install and initialize rosdep
+  - sudo apt-get install -qq -y python-dev
+                                python-catkin-pkg
+                                python-rosdep
+                                python-wstool
+                                ros-$ROS_DISTRO-catkin
+                                ros-$ROS_DISTRO-ros
+                                python-rosinstall
+                                python-catkin-tools
+                                python-catkin-pkg
+                                python-rospkg
+                                python-vcstools
+  - sudo `which rosdep` init
   - rosdep update
-install: # Use this to install any prerequisites or dependencies necessary to run your build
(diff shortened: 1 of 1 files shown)


efernandez:
FYI @bmagyar 

This is an updated cherry-pick of https://github.com/ros-controls/ros_controllers/pull/211

If this passes, I'll then update all the `.travis.yml` for the other repos on ros-controls

efernandez:
This is failing here:

``` bash
$ sudo apt-get install -qq -y python-dev python-catkin-pkg python-rosdep python-wstool ros-$ROS_DISTRO-catkin ros-$ROS_DISTRO-ros python-rosinstall python-catkin-tools python-catkin-pkg python-rospkg python-vcstools
E: Unable to locate package ros-kinetic-catkin
E: Unable to locate package ros-kinetic-ros
```

@bmagyar Is that because those two are NOT ready for kinetic yet?

Maybe the `$(lsb_release -cs)` I used it's NOT well converted into `xenial`?

Full raw log: https://s3.amazonaws.com/archive.travis-ci.org/jobs/133222368/log.txt

bmagyar:
I would expect `catkin_tools` to pull in the right version of `catkin` so I think the install instruction for that could be omitted.

BTW I highly doubt that those 2 are missing... Went looking around and I don't see any trouble around them.

.travis.yml
```diff
@@ -1,48 +1,81 @@
 # Travis Continuous Integration Configuration File For ROS Control Projects
-# Author: Dave Coleman
-language:
-  - cpp
-  - python
-python:
-  - "2.7"
+# Author: Dave Coleman, Enrique Fernandez
+sudo: required
+dist: xenial
```
130s, line 4: Are you sure if Ubuntu 16.04 is already available on Travis CI?

From [one of the logs](https://s3.amazonaws.com/archive.travis-ci.org/jobs/133222368/log.txt), it seems the job is running on 12.04 to me:

```
Build dist: xenial
[34m[1mBuild image provisioning date and time[0m
Tue Dec 15 16:36:18 UTC 2015
[34m[1mOperating System Details[0m
Distributor ID: Ubuntu
Description:    Ubuntu 12.04.5 LTS
Release:    12.04
Codename:   precise
```

bmagyar, line 4: Thanks for the note! Probably we are not getting (I definitely don't!) how Travis works haha. 
I've found this:
https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

Perhaps we need to change the scripts to use docker on Travis and get a 1604 environment that way? If that's possible at all...

130s, line 4: I've opened a replacement option https://github.com/ros-controls/ros_controllers/pull/225

efernandez:
I clearly misunderstood this https://github.com/travis-ci/travis-ci/issues/5821

@130s Thanks for your contribution, but I have a few questions and no time now to investigate them. Maybe you can answer some of them:
1. In your PR, you use the industrial_ci repo. How are the recipes there to install ros-controls repos?
2. The .travis.yml of your PR looks like a copy from industrial_ci because the emails and other general configuration are completely different from the ones we use here. Could you update them?
3. The Travis jobs is also failing on your PR. I'll have to look why?
4. If we decide to depend on industrial_ci, are there any additional dependencies or repos tested other than ros-controls ones?
5. If we need to do some changes to the Docker recipes, how would we do that?

130s:
Since some of your questions are good ones and common enough, I [added my answers to FAQ page](https://github.com/ros-industrial/industrial_ci/pull/34/commits/81e811b280d31d5f3c40ea91e65b3a651b27a1cd). See a [rendered version here](https://github.com/130s/industrial_ci/blob/81e811b280d31d5f3c40ea91e65b3a651b27a1cd/README.rst#faq).

> 1 In your PR, you use the industrial_ci repo. How are the recipes there to install ros-controls repos?
> 3 The Travis jobs is also failing on your PR. I'll have to look why?

Hopefully the README above helps.

> 2 The .travis.yml of your PR looks like a copy from industrial_ci because the emails and other general configuration are completely different from the ones we use here. Could you update them?

Oops! Email recipients are updated.
But config-wise I don't see any ros-control specific in https://github.com/ros-controls/ros_controllers/pull/224/files.

> 5 If we need to do some changes to the Docker recipes, how would we do that?

Thanks for the feedback! [Added a request](https://github.com/ros-industrial/industrial_ci/issues/57).

> 4 If we decide to depend on industrial_ci, are there any additional dependencies or repos tested other than ros-controls ones?

Sorry I don't get what you mean.

efernandez:
@130s Thanks for your replies. Regarding question 4, I mean: does industrial_ci has some dependencies that could brake the system and probably they're not needed at all for running the ros-controls repos?

But generally, using industrial_ci looks a good idea to me, as it already provides a good configuration to run kinetic inside Travis CI Docker containers. Thanks for pointing it out! :smiley: 

I've updated the comments on #225

130s:
> Regarding question 4, I mean: does industrial_ci has some dependencies that could brake the system and probably they're not needed at all for running the ros-controls repos?

I must say there are a lot of dependency; Travis CI server itself, ROS binary repository, github, all apt repos that the depended packages are fetched from. For example [this ticketed issue is caused by apt server side](https://github.com/ros-industrial/industrial_ci/issues/53) that `industrial_ci` has no influence on.
(And exception handling currently implemented is far from being perfect...)

efernandez:
Since https://github.com/ros-controls/ros_controllers/pull/225 was merged and this PR didn't add Travis support to Travis properly because it needs a Docker container for Ubuntu xenial, I'm going to close it in favour of the other.

PR #227, opened 2016-06-15
Add velocity feedforward term to velocity HardwareInterfaceAdapter

miguelprada:
In the current implementation of the `VelocityJointInterface` specialization for `HardwareInterfaceAdapter`, a velocity command for the joint will only be generated (by the PID) in the presence of tracking error, therefore perfect tracking can never be achieved.

Adding a velocity feedforward term unloads much of the burden from the PID, which only needs to correct for the position and velocity drift that may appear if (when) the robot does not perfectly follow the generated commands. This should allow to improve the controller's performance even decreasing the PID gains significantly, therefore improving also the controller's stability.

I'm not 100% sure of the effects of this change in configurations where the PIDs have been tuned to carry all the load of tracking the velocity reference. I'm inclined to believe that nothing bad will happen, although some tests or some sound theoretical proof would be nice to have in this respect.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
@@ -182,6 +182,9 @@ public:
       }
     }
 
+    // Load velocity feedforward enable state from parameter server
+    controller_nh.param("use_velocity_ff", use_velocity_ff_, false);
+
     return true;
   }
 
@@ -215,7 +218,7 @@ public:
     // Update PIDs
     for (unsigned int i = 0; i < n_joints; ++i)
     {
-      const double command = desired_state.velocity[i] + pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
+      const double command = (desired_state.velocity[i] * use_velocity_ff_) + pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
       (*joint_handles_ptr_)[i].setCommand(command);
     }
   }
@@ -223,6 +226,8 @@ public:
 private:
   typedef boost::shared_ptr<control_toolbox::Pid> PidPtr;
   std::vector<PidPtr> pids_;
+
+  bool use_velocity_ff_;
 
   std::vector<hardware_interface::JointHandle>* joint_handles_ptr_;
 };
--- a/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
@@ -81,7 +81,7 @@ public:
  *   joints:
  *     - head_1_joint
  *     - head_2_joint
- *   
+ *
  *   constraints:
  *     goal_time: 0.6
  *     stopped_velocity_tolerance: 0.02
@@ -182,8 +182,12 @@ public:
       }
     }
 
-    // Load velocity feedforward enable state from parameter server
-    controller_nh.param("use_velocity_ff", use_velocity_ff_, false);
+    // Load velocity feedforward gains from parameter server
+    velocity_ff_.resize(joint_handles.size());
+    for (unsigned int i = 0; i < velocity_ff_.size(); ++i)
+    {
+      controller_nh.param(std::string("velocity_ff/") + joint_handles[i].getName(), velocity_ff_[i], 0.0);
+    }
 
     return true;
   }
@@ -218,7 +222,7 @@ public:
     // Update PIDs
     for (unsigned int i = 0; i < n_joints; ++i)
     {
-      const double command = (desired_state.velocity[i] * use_velocity_ff_) + pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
+      const double command = (desired_state.velocity[i] * velocity_ff_[i]) + pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
       (*joint_handles_ptr_)[i].setCommand(command);
     }
   }
@@ -227,7 +231,7 @@ private:
   typedef boost::shared_ptr<control_toolbox::Pid> PidPtr;
   std::vector<PidPtr> pids_;
 
-  bool use_velocity_ff_;
+  std::vector<double> velocity_ff_;
 
   std::vector<hardware_interface::JointHandle>* joint_handles_ptr_;
 };
@@ -282,6 +286,13 @@ public:
       }
     }
 
+    // Load velocity feedforward gains from parameter server
+    velocity_ff_.resize(joint_handles.size());
+    for (unsigned int i = 0; i < velocity_ff_.size(); ++i)
+    {
+      controller_nh.param(std::string("velocity_ff/") + joint_handles[i].getName(), velocity_ff_[i], 0.0);
+    }
+
     return true;
   }
 
@@ -301,7 +312,7 @@ public:
 
   void updateCommand(const ros::Time&     /*time*/,
                      const ros::Duration& period,
-                     const State&         /*desired_state*/,
+                     const State&         desired_state,
                      const State&         state_error)
   {
     const unsigned int n_joints = joint_handles_ptr_->size();
@@ -314,7 +325,7 @@ public:
     // Update PIDs
     for (unsigned int i = 0; i < n_joints; ++i)
     {
-      const double command = pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
+      const double command = (desired_state.velocity[i] * velocity_ff_[i]) + pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
       (*joint_handles_ptr_)[i].setCommand(command);
     }
   }
@@ -322,6 +333,8 @@ public:
 private:
   typedef boost::shared_ptr<control_toolbox::Pid> PidPtr;
   std::vector<PidPtr> pids_;
+
+  std::vector<double> velocity_ff_;
 
   std::vector<hardware_interface::JointHandle>* joint_handles_ptr_;
 };
--- a/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
@@ -133,33 +133,19 @@ private:
 };
 
 /**
- * \brief Adapter for an velocity-controlled hardware interface. Maps position and velocity errors to velocity commands
- * through a velocity PID loop.
+ * \brief Helper base class template for closed loop HardwareInterfaceAdapter implementations.
  *
- * The following is an example configuration of a controller that uses this adapter. Notice the \p gains entry:
- * \code
- * head_controller:
- *   type: "velocity_controllers/JointTrajectoryController"
- *   joints:
- *     - head_1_joint
- *     - head_2_joint
- *   gains:
- *     head_1_joint: {p: 200, d: 1, i: 5, i_clamp: 1}
- *     head_2_joint: {p: 200, d: 1, i: 5, i_clamp: 1}
- *   constraints:
- *     goal_time: 0.6
- *     stopped_velocity_tolerance: 0.02
- *     head_1_joint: {trajectory: 0.05, goal: 0.02}
- *     head_2_joint: {trajectory: 0.05, goal: 0.02}
- *   stop_trajectory_duration: 0.5
- *   state_publish_rate:  25
- * \endcode
+ * Adapters leveraging (specializing) this class will generate a command given the desired state and state error using a
+ * velocity feedforward term plus a corrective PID term.
+ *
+ * Use one of the available template specializations of this class (or create your own) to adapt the
+ * JointTrajectoryController to a specidfic hardware interface.
  */
 template <class State>
-class HardwareInterfaceAdapter<hardware_interface::VelocityJointInterface, State>
+class ClosedLoopHardwareInterfaceAdapter
 {
 public:
-  HardwareInterfaceAdapter() : joint_handles_ptr_(0) {}
+  ClosedLoopHardwareInterfaceAdapter() : joint_handles_ptr_(0) {}
 
   bool init(std::vector<hardware_interface::JointHandle>& joint_handles, ros::NodeHandle& controller_nh)
   {
@@ -196,7 +182,7 @@ public:
   {
     if (!joint_handles_ptr_) {return;}
 
-    // Reset PIDs, zero velocity commands
+    // Reset PIDs, zero commands
     for (unsigned int i = 0; i < pids_.size(); ++i)
     {
       pids_[i]->reset();
@@ -204,7 +190,7 @@ public:
     }
   }
 
-  void stopping(const ros::Time& time) {}
+  void stopping(const ros::Time& /*time*/) {}
 
   void updateCommand(const ros::Time&     /*time*/,
                      const ros::Duration& period,
@@ -237,19 +223,23 @@ private:
 };
 
 /**
- * \brief Adapter for an effort-controlled hardware interface. Maps position and velocity errors to effort commands
- * through a position PID loop.
+ * \brief Adapter for an velocity-controlled hardware interface. Maps position and velocity errors to velocity commands
+ * through a velocity PID loop.
  *
- * The following is an example configuration of a controller that uses this adapter. Notice the \p gains entry:
+ * The following is an example configuration of a controller that uses this adapter. Notice the \p gains and \p velocity_ff
+ * entries:
  * \code
  * head_controller:
- *   type: "effort_controllers/JointTrajectoryController"
+ *   type: "velocity_controllers/JointTrajectoryController"
  *   joints:
  *     - head_1_joint
  *     - head_2_joint
  *   gains:
  *     head_1_joint: {p: 200, d: 1, i: 5, i_clamp: 1}
  *     head_2_joint: {p: 200, d: 1, i: 5, i_clamp: 1}
+ *   velocity_ff:
+ *     head_1_joint: 1.0
+ *     head_2_joint: 1.0
  *   constraints:
  *     goal_time: 0.6
  *     stopped_velocity_tolerance: 0.02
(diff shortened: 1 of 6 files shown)


bmagyar:
Nice catch! First, could you please add some unit tests of expected outcomes? That is a good base to start a discussion on.

Also adding @ipa-fxm for further thoughts

miguelprada:
I'd be glad to, but I'm honestly not quite sure of what the tests should look like.

Only thing I can think of is making this feedforward term optional, using a parameter, and to include a test which compares the evolution of the velocity-based trajectory controller with and without the feedforward term. This would require adding a `VelocityJointInterface` to the `RRbot` used for testing, although it seems quite doable.

An additional benefit would be that adding a `VelocityJointInterface` to the `RRbot` would also allow to run the rest of the existing tests on the velocity-based controller, which, if I'm not wrong, are currently run only for the position-based trajectory controller.

What do you think? Any better ideas @bmagyar, @ipa-fxm, anyone?

miguelprada:
I just opened PR #228, duplicating the tests for the velocity-based joint trajectory controller, as a basis for testing this PR.

Any thoughts on how the tests for the (optional) feedforward term should look like?

And BTW, what's the nice thing to do with this PR? Rebase the only commit on top of #228 and reopen? Not quite comfortable with the rules of etiquette here yet :sweat_smile:

bmagyar:
Sorry Miguel for the delay. 
First: PRs can be rebased without the need to reopen them. You can do this by rebasing the branch you made the pull request from, and doing a push --force on it, overwriting the history. Make sure you have a backup if you don't feel comfortable doing this.

Duplicating all that test code may not be the best solution. I mean 1000 lines of duplication are kinda hard to maintain :) 

Any comments on this, @efernandez ?

I'm going to be quite busy this week but will try to get back next week, feel free to ping me.

efernandez:
I'm not an expert in control. My understanding is that with this the PID actually works on the state error, which is good.

Since I'm not sure of the side effects of this, I'd feel more comfortable if you could provide a param to enable/disable the feedforward term, and set it to false by default to keep previous behaviour; regarding the default value, I honestly don't have a strong opinion. Indeed, I really like that the PIDs can have smaller values and actually work on the state error, which I think it makes more sense.

miguelprada:
Sorry, guys, for the delay.

> Duplicating all that test code may not be the best solution. I mean 1000 lines of duplication are kinda hard to maintain :)

Totally agree with this, but I was a bit confused by how the `TEST_F` macros have harcoded values for the test fixture and figured at first that it was important to make them different across tests.

I just commited to #228 with what I hope is a better solution for the tests, which duplicates no code at all, and just requires duplicating the controller configuration and test launchfile. Let me know what you think, @bmagyar.

> Since I'm not sure of the side effects of this, I'd feel more comfortable if you could provide a param to enable/disable the feedforward term, and set it to false by default to keep previous behaviour; regarding the default value, I honestly don't have a strong opinion. Indeed, I really like that the PIDs can have smaller values and actually work on the state error, which I think it makes more sense.

I also thought about this and think it's a reasonable solution. Regarding the default value, I would be inclined to make it enabled by default at some point, but I understand if you want to leave that for future releases in order to avoid messing with working setups.

I'll give this a go, @efernandez.

miguelprada:
Hi guys, just changed the velocity feedforward term to be enabled or disabled based on a parameter named `use_velocity_ff`. The default case if the parameter is not found is to disable the feedforward term, although this is straightforward to change if it is decided to enable it by default.

I also rebased on top of #228, even though it hasn't been merged yet. I'm sorry if this causes any inconveniences.

Since #228 does not contain the velocity feedforward term, though, there's no specific test to validate it. Can you think of any additional test that can be added to validate this specific addition?

bmagyar:
Our common practice is to have one (perhaps one set of) explicit test(s) for each parameter we add. This makes sure that branch testing is somewhat achieved.

miguelprada:
I finally managed to add a dedicated test for the velocity feedforward optional term!

What I did is:
- drastically lower the PID gains for the `VelocityJointInterface`-based controller and enable the feedforward term in [rrbot_vel_controllers.yaml](https://github.com/miguelprada/ros_controllers/blob/c3e0988de79843d7b0f6bcf9320972ec061d4d3a/joint_trajectory_controller/test/rrbot_vel_controllers.yaml), which does not affect the outcome of the existing tests
- add a [new test](https://github.com/miguelprada/ros_controllers/blob/c3e0988de79843d7b0f6bcf9320972ec061d4d3a/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp#L985)<sup>1</sup> which executes the same trajectory twice, one with and one withouth `use_velocity_ff`, showing that the first one succeeds while the second one aborts due to a path tolerance violation
- [disable the test](https://github.com/miguelprada/ros_controllers/blob/c3e0988de79843d7b0f6bcf9320972ec061d4d3a/joint_trajectory_controller/test/joint_trajectory_controller.test#L39) for `PositionJointInterface`
- [fix a typo](https://github.com/miguelprada/ros_controllers/blob/c3e0988de79843d7b0f6bcf9320972ec061d4d3a/joint_trajectory_controller/test/joint_trajectory_controller_vel.test#L37)

Thoughts?

<sup>1</sup> If you wonder about the place where I inserted the new test, I put it there because I initially added it at the end and [this assertion](https://github.com/miguelprada/ros_controllers/blob/c3e0988de79843d7b0f6bcf9320972ec061d4d3a/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp#L993) was always failing. Moving it before the tolerance checking tests made it work flawlessly, so I guess there's some side effect from the latter two tests. I haven't had time to figure out what happens, though.

miguelprada:
I don't understand what's going on with the tests. They:
- just failed when triggered by the PR, but for the `PositionJointInterface` controller test [here](https://travis-ci.org/ros-controls/ros_controllers/jobs/232187994#L2077)
- failed as well (in different places) for my fork when I [rebased the branch](https://travis-ci.org/miguelprada/ros_controllers/builds/232187988) before the PR, a test whose results I didn't wait for because it...
- [passed](https://travis-ci.org/miguelprada/ros_controllers/builds/231752782) for the **very same commit** (`c3e0988`) when I was testing on another branch before rebasing `add_velocity_ff`
- they also consistently pass for me locally

Any ideas? Can anyone else please try?

mathias-luedtke:
The current test set is very sensitive to timing and processor load. That's why we have configured travis to run only one test at a time. However, this approach does not fix the timing issues, but reduces the number of occurrences. In addition I am not 100% sure that we really limit the number of jobs properly.

Right now I am runnign the tests of the current master and your PR locally with `run_ci`.

**Update:** Tests passed locally for multiple times.

miguelprada:
`ROS_DISTRO=kinetic ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu rosrun industrial_ci run_ci` failed for me 2 out of 10 times locally. :confused:

miguelprada:
Also, if this happens to be finally merged, is there any chance the changes are backported to indigo? Not sure about the compatibility policy that's followed around here.

joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
```diff
@@ -213,7 +216,7 @@ class HardwareInterfaceAdapter<hardware_interface::VelocityJointInterface, State
     // Update PIDs
     for (unsigned int i = 0; i < n_joints; ++i)
     {
-      const double command = pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
+      const double command = (desired_state.velocity[i] * use_velocity_ff_) + pids_[i]->computeCommand(state_error.position[i], state_error.velocity[i], period);
```
mathias-luedtke: This line multiplies a boolean with a double. value.
I would prefer a `velocity_ff`param, which defaults to `0.0`

mathias-luedtke:
Travis runs
```
 rosrun industrial_ci run_ci ROS_DISTRO=kinetic ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu CATKIN_PARALLEL_TEST_JOBS=-p1 ROS_PARALLEL_TEST_JOBS=-j1
```

joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
miguelprada: Is it the unconventional `double * bool` multiplication which you don't like or is it restricting the range of values to effectively `0` and `1`? If its the multiplication I can always reformulate in a conditional statement.

I ask because I can't think of any reasonable case where feedforward _gains_ other than `0` or `1` make sense. Can you?

bmagyar:
Thanks for your work on this @miguelprada . I can't join in the review discussion at the moment but I'd like to comment on the note about Indigo. Personally I'd like to eventually freeze Indigo and only allow bugfixes but we can discuss small and controlled features to be added if there's a big demand.

joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
mathias-luedtke: I am not sure if that really made sense, but we used 0.9 for the Care-O-bot 3 arms (pre-`ros_control` stack).
I don't see a reason to restrict it to either `0.0` or `1.0`, but I am no contoller expert.

miguelprada: Me neither. I had always assumed that the feedforward was either there or not, but let's assume whoever put the `0.9` in the COB knew what he was doing and make the parameter a double.

Then, what would you suggest for values greater than `1.0`? Should we raise some warning and clamp, just raise a warning or do nothing altogether?

miguelprada:
> Personally I'd like to eventually freeze Indigo and only allow bugfixes but we can discuss small and controlled features to be added if there's a big demand.

I'm not sure about the demand, but there's a specific reason I think it would be nice to have this in indigo.

The `ur_modern_driver`'s implementation performs much better when using the `VelocityJointInterface` than when using the `PositionJointInterface` (due to a non-negligible delay introduced by the latter). However, the behavior of the `speedj` URScript command (used by the `VelocityJointInterface`) has changed a [couple](https://github.com/ThomasTimm/ur_modern_driver/pull/57) of [times](https://github.com/ThomasTimm/ur_modern_driver/pull/100) in past UR software updates, introducing delays and making some users of this _driver_ experience stability issues ([myself included](https://github.com/ThomasTimm/ur_modern_driver/issues/55)). As a last note, `ur_modern_driver` is not yet available for Kinetic.

Adding this patch to the indigo branch would allow reducing the gains in the default configuration for the `joint_trajectory_controller` in that driver, potentially removing/reducing the instability issues.

miguelprada:
Just added a new commit making the parameter a `double`, as @ipa-mdl suggested.

As I was on it, I also added the optional feedforward available for the `EffortJointInterface`. In fact, it is in the latter case where I can see values different than `0` or `1` make any sense (e.g. to compensate for friction, based on velocity), but I guess it's better to have some consistency in the parameter API for the different interfaces.

Also, I thought it made more sense to be able to set a different value of the feedforward _gain_ for each joint, so I changed how the parameter is read as well.

One missing thing would be to add tests for the `EffortJointInterface`, similarly as #228, but it would be much more complex than that. I think it's beyond the scope of this PR, but let me know if you think otherwise.

Thoughts, anyone?

EDIT: I also rebased on top of current `kinetic-devel`. Hope that's fine.

miguelprada:
Friendly ping, @bmagyar, @ipa-mdl.

Not sure why job [#528.3](https://travis-ci.org/ros-controls/ros_controllers/jobs/305343529) is failing. It passes for me locally. Can anyone try to relaunch?

bmagyar:
Thumbs up for the rebase!
I've also relaunched the travis job.
Disregard the failed test on `four_wheel_steering_controller`, we have to revise those numbers to get more stable tests.

miguelprada:
And yet another failed test which I'm unable to reproduce locally :sweat: 

The failed test case seems to be different each time, though. Not sure what to do about this, to be honest.

mathias-luedtke:
> And yet another failed test which I'm unable to reproduce locally :sweat:

It's all about timing, hard to reproduce..

> Not sure what to do about this, to be honest.

I would suggest to ignore it for now, unless all jobs fail..

miguelprada:
Are you waiting for any additional input from me? Maybe I'm missing something, but I think all existing comments have been addressed.

It would be a nice Xmas present to finally get this merged so I can stop including `ros_controllers` source in all my workspaces :wink:

joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
```diff
@@ -277,6 +286,13 @@ class HardwareInterfaceAdapter<hardware_interface::EffortJointInterface, State>
       }
     }
 
+    // Load velocity feedforward gains from parameter server
```
mathias-luedtke: Can this be reused somehow?
It's the same as in line 185..
FF-enabled base class?

joint_trajectory_controller/test/joint_trajectory_controller.test
```diff
@@ -35,5 +35,6 @@
   <test test-name="joint_trajectory_controller_test"
         pkg="joint_trajectory_controller"
         type="joint_trajectory_controller_test"
-        time-limit="85.0"/>
+        time-limit="85.0"
+        args="--gtest_filter=-JointTrajectoryControllerTest.jointVelocityFeedForward"/>
```
mathias-luedtke: Why do you exclude this test?

mathias-luedtke: I guess this test file is for the position interface only?
Since the code gets built twice anyway, It could be an option to add macro switch to emphasize that this test won't get run for the 

At least a comment should be added.

miguelprada: > I guess this test file is for the position interface only?

Exactly.

> Since the code gets built twice anyway

Yeah, not very happy about this either. I'll see if I can come up with some way to only build them once.

And I'll add some comment anyways.

mathias-luedtke: >I'll see if I can come up with some way to only build them once.

Something like
`target_compile_definitions(joint_trajectory_controller_vel_test PRIVATE TEST_VELOCITY_FF=1)` would justify somehow to build it twice ;)

mathias-luedtke: IMHO building it twice is not too bad.

Building it just once would require (?) to:
1. move tests (without FF) into a library
2. expose text fixture class as header (in test directory)
2. add  source file for the main function
3. `add_rostest_gtest` with main, linked to lib
3. `add_rostest_gtest` with main + ff-test, linked to lib

bmagyar:
I won't have sleepless nights over building some tests twice :)
Just make sure we test all we need to. Me or Mathias can merge this once
finished.

joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
miguelprada: In fact, most code is duplicated between the `VelocityJointInterface` and the `EffortJointInterface` adapters. I'll refactor into a common base class for these specializations.

joint_trajectory_controller/test/joint_trajectory_controller.test
miguelprada: Finally went for the `#ifdef TEST_VELOCITY_FF` check. Is it more or less what you had in mind, @ipa-mdl ?

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -981,6 +981,9 @@ TEST_F(JointTrajectoryControllerTest, ignorePartiallyOldActionTraj)
 }
 
 // Velocity FF parameter ///////////////////////////////////////////////////////////////////////////////////////////////
+// This test will only be built and run for the VelocityJointInterface-based version of the JointTrajectoryController
+
+#ifdef TEST_VELOCITY_FF
```
mathias-luedtke: Please consider `#if TEST_VELOCITY_FF` to handle `TEST_VELOCITY_FF=0` as well.

mathias-luedtke:
Thanks for the refactoring!
Why did you name it `ClosedLoopHardwareInterfaceAdapter`? Can't position controllers be closed-loop as well?

Somehow the joint trajectory controller have become unstable with the last commit.

joint_trajectory_controller/test/joint_trajectory_controller.test
mathias-luedtke: > Finally went for the #ifdef TEST_VELOCITY_FF check. Is it more or less what you had in mind,

It's exactly what I had in mind :D

miguelprada:
> Why did you name it ClosedLoopHardwareInterfaceAdapter? Can't position controllers be closed-loop as well?

In those cases the adapter closes a loop by computing their commands based on the desired state and the state error. This is not the case for the rest of the adapter specializations.

> Somehow the joint trajectory controller have become unstable with the last commit.

Which one? The one modifying the tests or the refactor? I must admit I just run the tests locally and pushed when I saw that they passed :flushed: 

I'll play with it locally a bit.

mathias-luedtke:
> In those cases the adapter closes a loop by computing their commands based on the desired state and the state error. This is not the case for the rest of the adapter specializations.

I see, this makes sense!

> Which one? The one modifying the tests or the refactor?

79bf8d10e8b70ad88a02e611431faabef688e0c1. I have restarted [537.2](https://travis-ci.org/ros-controls/ros_controllers/builds/318036783). 

537.3 failed as well, but with [a different errror](https://travis-ci.org/ros-controls/ros_controllers/jobs/318036786#L8291)

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
miguelprada: Just noticed this comment. Updated.

mathias-luedtke:
Thanks for the update!

i have restarted all test jobs, which failed for unrelated errors :-/.

mathias-luedtke (approved):
Thanks a lot for your patience!
This is definitely ready for merge now.

I will wait some time (a couple of days?) for others to review it as well.

miguelprada:
No hurries. What's a couple of days compared with a year an a half I've been dragging this along :wink:

mathias-luedtke:
ho ho ho :santa:

PR #232, opened 2016-08-10
Included angles in dependencies

mryellow:
Was coming up as `angles/angles.h` not found on my `debian:jessie` system.


bmagyar:
Hi Mr Yellow,
Thanks for the fix! 

Cheers,
Mr Black.

bmagyar:
I cherry-picked your commit to jade-devel and kinetic-devel as well.

Issue #234, opened 2016-08-31
joint_position_controller build error


nxdefiant:
joint_position_controller fails to build on Debian testing (gcc 6.1.1, boost 1.61) with the following error:

```
/opt/ros/ros_catkin_ws/src/ros_controllers/effort_controllers/src/joint_position_controller.cpp: In member function 'virtual bool effort_controllers::JointPositionController::init(hardware_interface::EffortJointInterface*, ros::NodeHandle&)':
/opt/ros/ros_catkin_ws/src/ros_controllers/effort_controllers/src/joint_position_controller.cpp:88:41: error: no match for 'operator=' (operand types are 'boost::shared_ptr<const urdf::Joint>' and 'urdf::JointConstSharedPtr {aka std::shared_ptr<const urdf::Joint>}')
   joint_urdf_ = urdf.getJoint(joint_name);
```

bmagyar:
Which ROS distribution?
I've seen a similar issues these days when I was goofing around building
unreleased stuff from source.

It seems urdf updated their *ConstSharedPtr to use `std::shared_ptr`
instead of `boost::shared_ptr` but it hasn't been released yet.

I think I can quickly fix this tomorrow morning, or if you already have the
fix please open a pull request now and I merge it tomorrow.

nxdefiant:
ros is kinetic-devel, have not attempted to fix it yet.

bmagyar:
Btw I see the path you are using in the logs. Try not to build things in
/opt because it may mess things up big time. Have your sources in your home
and use catkin install with a path set to opt.

bmagyar:
Hey, 
Give a shot to the pull request I referred below! This is a non-invasive solution that should even keep current code untouched, I refactored the affected parts to use the name macros.

nxdefiant:
It compiles, thanks.

bmagyar:
:+1:

PR #235, opened 2016-09-01
Replace boost::shared_ptr<urdf::XY> with urdf::XYConstSharedPtr when exists

bmagyar:
As per noted by #234 , this may cause an issue in the future so instead of changing boost::shared_ptr instances here to std::shared_ptr I refactored the current code to use the macro-generated names. 

No changes in current ABI should be introduced by this change unless compiled with new urdf library.

@efernandez thoughts?

### DIFF
--- a/gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h
+++ b/gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h
@@ -45,6 +45,7 @@ boost::shared_ptr<urdf::Model> getUrdf(const ros::NodeHandle& nh, const std::str
 {
   boost::shared_ptr<urdf::Model> urdf(new urdf::Model);
 
+
   std::string urdf_str;
   // Check for robot_description in proper namespace
   if (nh.getParam(param_name, urdf_str))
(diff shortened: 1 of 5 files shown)


gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h
```diff
@@ -45,6 +45,7 @@ boost::shared_ptr<urdf::Model> getUrdf(const ros::NodeHandle& nh, const std::str
 {
   boost::shared_ptr<urdf::Model> urdf(new urdf::Model);
 
+
```
efernandez, line 48: remove this line

gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h
```diff
@@ -45,6 +45,7 @@ boost::shared_ptr<urdf::Model> getUrdf(const ros::NodeHandle& nh, const std::str
 {
   boost::shared_ptr<urdf::Model> urdf(new urdf::Model);
```
efernandez, line 46: `urdf::ModelSharedPtr` ?!

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -286,7 +285,7 @@ bool JointTrajectoryController<SegmentImpl, HardwareInterface>::init(HardwareInt
   boost::shared_ptr<urdf::Model> urdf = getUrdf(root_nh, "robot_description");
```
efernandez, line 285: `urdf::ModelSharedPtr` ?!

efernandez:
LGTM, just minor comments

gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h
bmagyar: As funny as that sounds, that type doesn't seem to exist.

efernandez: Fair enough. Merging, but consider a PR against urdf to add it.

Issue #233, opened 2016-08-31
Cryptic error message from joint_trajectory_controller when sending trajectories that occur completely in the past


BrettHemes-3M:
Originally brought up in [ros_industrial/kuka_experimental/#69](https://github.com/ros-industrial/kuka_experimental/issues/69), the `joint_trajectory_controller` reports a rather useless error when a trajectory is sent that occurs completely in the past.
(i.e., trajectory.header.stamp + trajectory.points.back().time_from_start  <  ros::Time::now())

The error reported from calling SimpleActionClient::sendGoal(...) is as follows:

> [ERROR] [1472046592.407557364]: Unexpected exception caught when initializing trajectory from ROS message data.

Seems to be from line 518 [here](http://docs.ros.org/indigo/api/joint_trajectory_controller/html/c++/joint__trajectory__controller__impl_8h_source.html#l00518).  After catching and reporting the error, the action server reports success and the robot remains motionless.

After some thought I guess this is technically correct behavior due to the fact that the goal trajectory is in essence complete, but a more informative error message (perhaps from a check to see if the trajectory occurs in the present) might save some confusion in the future.  Thoughts?

Additionally, there is a disclaimer in the documentation for [initJointTrajectory](http://docs.ros.org/hydro/api/joint_trajectory_controller/html/c++/namespacejoint__trajectory__controller.html#a8b55b5e436dc0e49ad024d2f3bd7568c) that states:

> This function (joint_trajectory_controller::initJointTrajectory) does not throw any exceptions by itself, but the segment constructor might. In such a case, this method should be wrapped inside a try block. 

... but it is a bit buried.

bmagyar:
Hi Brett,

Thanks for the feedback!

I've taken a quick look and couldn't find (upon first check) that the Segment constructor ( [here](http://docs.ros.org/indigo/api/joint_trajectory_controller/html/c++/joint__trajectory__segment_8h_source.html) ) would throw anything other than `std::invalid_argument` exceptions.

Could you find out where this exception comes from? One way would be to remove the `catch(...)` part and get a stacktrace with our good friend `gdb`.

The solution for this should be to add explicit handling for these cases, maybe the explicit check for the condition you wrote above.

@adolfo-rt any thoughts?

BrettHemes-3M:
The backtrace isn't as straightforward as I expected... I have included it below:

```
#0  0x00007ffff61bfcc9 in __GI_raise (sig=sig@entry=6)
[...]

PR #236, opened 2016-09-01
Add test that sends trajectory entirely in past

bmagyar:
This recreates the crash raised in #233 as a new test.


bmagyar:
@BrettHemes could you please confirm that this test covers your issue from the past?

BrettHemes-3M:
@bmagyar yes, this test covers the issue I was having.

bmagyar:
Thank you!

PR #245, opened 2016-12-05
Supply nh to urdf::Model

### LINKED ISSUE #244: Can't initialize effort_controllers/JointPositionController with non-global robot_description parameter
Ref: http://answers.ros.org/question/249138/cant-initialize-effort_controllersjointpositioncontroller-with-non-global-robot_description-parameter

Here's an outline of my problem:

1. I have multiple robots in Gazebo with different URDF descriptions. Each robot is pushed into a namespace (/robot1/...), and each frame id is pre-pended with a prefix (robot1_....). Consequently, I have no robot_description param in the global namespace.
2. libgazebo_ros_control (running inside Gazebo) initializes ControllerManager. Gazebo runs in the global namespace.
3. `effort_controllers/JointPositionController` (running inside Gazebo) calls [`urdf::Model::initParam`](https://github.com/ros/robot_model/blob/kinetic-devel/urdf/src/model.cpp#L89) which does not respect the controller's constructed namespace, and executes a [`searchParam`](https://github.com/ros/robot_model/blob/kinetic-devel/urdf/src/model.cpp#L96) in Gazebo's global namespace, which fails.

I can't figure out how to tell the controller to read the robot_description from `robot1/robot_description`. 

Furthermore, It's unclear to me why each controller is trying to read the URDF independently in the first place. Doesn't it make more sense to use the controller manager.

If this issues needs a patch, and I can provide one. Would an acceptable solution be that I duplicate Model's search inside a controller, since the controller has the correct NodeHandle?

piyushk:
Merge after ros/robot_model#168

Closes #244

### DIFF
--- a/effort_controllers/src/joint_position_controller.cpp
+++ b/effort_controllers/src/joint_position_controller.cpp
@@ -80,7 +80,7 @@ bool JointPositionController::init(hardware_interface::EffortJointInterface *rob
 
   // Get URDF info about joint
   urdf::Model urdf;
-  if (!urdf.initParam("robot_description"))
+  if (!urdf.initParamWithNodeHandle("robot_description", n))
   {
     ROS_ERROR("Failed to parse urdf file");
     return false;
--- a/velocity_controllers/src/joint_position_controller.cpp
+++ b/velocity_controllers/src/joint_position_controller.cpp
@@ -81,7 +81,7 @@ bool JointPositionController::init(hardware_interface::VelocityJointInterface *r
 
   // Get URDF info about joint
   urdf::Model urdf;
-  if (!urdf.initParam("robot_description"))
+  if (!urdf.initParamWithNodeHandle("robot_description", n))
   {
     ROS_ERROR("Failed to parse urdf file");
     return false;


piyushk:
@bmagyar I've updated this PR as per the discussion in ros/robot_model#168

bmagyar:
Great! Thanks for updating. Please keep in mind that it'll take a short while for that PR changes to be released, I can only merge this one after that.

piyushk:
Closing and reopening to trigger travis rebuild, since I don't have write permissions on this repository.

piyushk:
@bmagyar: It looks like the PR is ready for review/merge. I was hoping to have the fix out before the next sync. Could you take a look again?

bmagyar:
:+1: from me 

@progtologist , what do you think? do you have anything to test this on directly?

progtologist:
:thumbsup: from me as well.
@bmagyar I don't have something ready right now and since syncing is right around the corner, merge this and I will make a unit test for testing this asap.

bmagyar (approved):

bmagyar:
Deal. I'll take your word for it, but also open an issue as a kind reminder ;)

PR #246, opened 2016-12-07
Add urdf compatibility header

bmagyar:
Well...this is sort of a long shot but my tests with a Wily docker container revealed that a release of the current state with the new urdf headers would break on Wily the same way as it did with `gazebo_ros_packages`. 

https://github.com/ros-controls/ros_control/issues/258

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -42,6 +42,8 @@
 
 #include <urdf_parser/urdf_parser.h>
 
+#include <urdf/urdfdom_compatibility.h>
+
 #include <boost/assign.hpp>
 
 #include <diff_drive_controller/diff_drive_controller.h>


efernandez:
Is the same valid for Xenial? Then I can give it a try on my docker container.

How are you testing? Just `catkin build`?

bmagyar:
Yeah, it's a compilation issue only with Wily. Our good friend Travis should ensure it builds on Xenial.

mryellow:
Confirmed this looks to fix my build error with Debian Jessie source install.

https://github.com/ros/robot_model/issues/171

Not sure if it's the intended way to go about this though, assumed that nothing would need changing and the compatibility header would be included by the parser. https://github.com/ros/robot_model/blob/1.12.6/urdf_parser_plugin/include/urdf_parser_plugin/parser.h#L40

mathias-luedtke:
> Not sure if it's the intended way to go about this though, assumed that nothing would need changing and the compatibility header would be included by the parser. 

It is not included in https://github.com/ros/urdfdom/blob/master/urdf_parser/include/urdf_parser/urdf_parser.h

Update: but included in https://github.com/ros/robot_model/blob/kinetic-devel/urdf/include/urdf/model.h#L43

Update 2: but not in https://github.com/ros/urdfdom_headers/blob/master/urdf_model/include/urdf_model/model.h

Long story short: Given this confusing header situation, this fix is the best bet.

bmagyar:
Had to wait a while for this... I'll restart the travis jobs to see where we stand with this ATM. Once we're all green, this is a merger.

mathias-luedtke:
>  it's a compilation issue only with Wily.

with https://github.com/ros-industrial/industrial_ci/pull/116 you could test on wily by setting `UBUNTU_OS_CODE_NAME`

bmagyar:
Thanks @ipa-mdl , I've tested it with that, all seems fine now!
It's a bit more convenient than my previous efforts doing it all by hand :)

PR #247, opened 2016-12-14
Skip trajectories with an empty set of points.

miguelprada:
Upon receiving a trajectory goal with an empty set of points, `updateTrajectoryCommand` will detect it and set a hold trajectory, but [return `true`](https://github.com/ros-controls/ros_controllers/blob/99d55fc2a6a24612e5cbe2bcd8145d78e4cddbfd/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L513) nevertheless, causing `goalCB` to accept it. However the controller then fails to succeed or abort, since [this check](https://github.com/ros-controls/ros_controllers/blob/99d55fc2a6a24612e5cbe2bcd8145d78e4cddbfd/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L389) always fails (the segment has no associated goal handle, being part of the hold trajectory) and any client waiting for a result will be stuck.

It's my impression that avoiding to set the hold trajectory and just returning false is the right thing to do (just like a [few lines before](https://github.com/ros-controls/ros_controllers/blob/99d55fc2a6a24612e5cbe2bcd8145d78e4cddbfd/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L489)), but I'm not quite sure of the potential side-effects.

What do you think?

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -159,12 +159,16 @@ template <class SegmentImpl, class HardwareInterface>
 inline void JointTrajectoryController<SegmentImpl, HardwareInterface>::
 trajectoryCommandCB(const JointTrajectoryConstPtr& msg)
 {
-  const bool update_ok = updateTrajectoryCommand(msg, RealtimeGoalHandlePtr());
-  if (update_ok || (msg && msg->points.empty()))
+  if (msg && msg->points.empty())
   {
     ROS_DEBUG_NAMED(name_, "Empty trajectory command, stopping.");
     setHoldPosition(time_data_.readFromRT()->uptime);
     preemptActiveGoal();
+  }
+  else
+  {  
+    const bool update_ok = updateTrajectoryCommand(msg, RealtimeGoalHandlePtr());
+    if (update_ok) {preemptActiveGoal();}
   }
 }
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -508,9 +508,8 @@ updateTrajectoryCommand(const JointTrajectoryConstPtr& msg, RealtimeGoalHandlePt
   // Hold current position if trajectory is empty
   if (msg->points.empty())
   {
-    setHoldPosition(time_data->uptime);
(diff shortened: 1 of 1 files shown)


miguelprada:
I just realised (because of the travis errors) that this change breaks the expected behavior of the controller cancelling the currently active goal on receiving an empty trajectory via topic. I'll try to figure out some other way of avoiding this error.

bmagyar:
Doesn't `setHoldPosition` and `return false` implement this? 

...but yeah your point about tests on Travis is right, this is an expected behaviour. Returning false wouldn't hurt though I think...

miguelprada:
Setting the hold trajectory would effectively cancel any currently running action goal but returning false would cause it not to be preempted ([here](https://github.com/ros-controls/ros_controllers/blob/99d55fc2a6a24612e5cbe2bcd8145d78e4cddbfd/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L163)). :confused:

miguelprada:
After thinking about this a bit more I think that the change in the last commit makes the most sense.

There, `updateTrajectoryCommand` only returns `true` if the given trajectory has actually been accepted for execution, which makes it semantically more coherent IMO.

The expected behavior of cancelling any running trajectory upon receiving an empty trajectory via the command topic is implemented on the command topic callback itself.

Thoughts?

miguelprada:
Just realised the implementation was completely wrong. Should be ok now.

Test still fail, but don't think it's because of this PR. Test against the _main_ ROS repo passes but the other two fail. The one against `ros-shadow-fixed` because of some test failing for `diff_drive_controller`, and the one fetching dependencies from source seems to produce too long an output and is terminated.

mathias-luedtke:
This patch:
1. discards goals without points in action.
2. holds movement if topic message does not contains points.

Why should topic and action behave differently?
We should not change the current behaviour (empty goal stops the motion).

>  However the controller then fails to succeed or abort, since this check always fails (the segment has no associated goal handle, being part of the hold trajectory) and any client waiting for a result will be stuck.

IMHO we need to fix the [check](https://github.com/ros-controls/ros_controllers/blob/99d55fc2a6a24612e5cbe2bcd8145d78e4cddbfd/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L389), e.g. by setting the goal handle of the hold trajectory.

> Doesn't setHoldPosition and return false implement this?

If the goal is not accepted, it should have no impact on the controller.

miguelprada:
I think #271 addresses the same problem as this, but following the suggestions of @ipa-mdl. Would you have a look @bmagyar?

bmagyar:
Closing in favour of #271

Issue #251, opened 2017-01-11
seq is 0 in differential_drive controller


Timple:
If we run the differential_drive controller in open_loop mode, the seq in the headers stays 0. Although the timestamp in the header is correct.

In the code, I don't see the sequency in the header being set anywhere, is this a bug?

bmagyar:
I think this is just something none of us ever needed. If you have a need
for this, please feel free to provide a pull request implementing it.

Timple:
Well, it is in the header definition that it should be sequential. We have some tf problems with our robot and this sequence being zero is what stood out. 

It might not be the issue, but I will make a pull request nonetheless next week to fulfill the specification of the header.

bmagyar:
I cannot hurt, but keep in mind that this controller is being used on many
robots in production already, has tons of hours of test time. Seq is too
simple of a banana peel to slip on :)

Tf issues usually arise when time is out of sync, perhaps you have a multi
computer system where you'd need to periodically sync clocks?

On 12 Jan 2017 08:24, "Tim Clephas" <notifications@github.com> wrote:

Well, it is in the header definition that it should be sequential. We have
some tf problems with our robot and this sequence being zero is what stood
out.

It might not be the issue, but I will make a pull request nonetheless next
week to fulfill the specification of the header.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/ros-controls/ros_controllers/issues/251#issuecomment-272102551>,
or mute the thread

mathias-luedtke:
> We have some tf problems with our robot and this sequence being zero is what stood out.

I [doubt](https://github.com/ros/geometry2/search?q=seq) that tf uses the sequence number..

romainreignier:
It does not hurt to have it but it seems that `seq` is kind of deprecated:
https://discourse.ros.org/t/is-the-sequence-number-in-header-message-deprecated/1528
http://answers.ros.org/question/55126/why-does-ros-overwrite-my-sequence-number/?answer=55132#post-id-55132

bmagyar:
Yeah, I'd prefer to not add support for deprecated stuff so late. I'll close both this and the referenced PR. Thanks for bringing this up though!

PR #253, opened 2017-01-30
increasing seq in differential_drive controller topic headers

Timple:
This pull request solves https://github.com/ros-controls/ros_controllers/issues/251

However, it was not the source of the problems we where having. Feel free to accept or reject based on your judgement.

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
@@ -101,6 +101,7 @@ namespace diff_drive_controller{
     /// Odometry related:
     ros::Duration publish_period_;
     ros::Time last_state_publish_time_;
+    int seq_;
     bool open_loop_;
 
     /// Hardware handles:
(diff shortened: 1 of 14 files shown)


bmagyar:
Could I ask you to squash those commits? I'd merge it then

Timple:
Done!

bmagyar:
Ok, I only see `seq_ += seq_;` or similar. This'd double the value of the sequ variable, which is probably not the intended behaviour but since it starts off with 0 as initial value it stays there, doesn't it?

Timple:
That is absolutely right. It slipped through when I was trying to clean up stuff.

bmagyar:
+1 on first look, I'll try to test it tonight

bmagyar:
I'm now a bit puzzled whether we should actually go this way.
Give this a read:
https://discourse.ros.org/t/is-the-sequence-number-in-header-message-deprecated/1528

What is your opinion? I'd prefer not to add changes for this then.

mathias-luedtke:
>  I'd prefer not to add changes for this then.

+1

PR #255, opened 2017-03-06
JointPositionController can now load the robot_description from root_node_handle ns

shisha101:
Please refer to Issue https://github.com/ros-controls/ros_controllers/issues/256

### DIFF
--- a/effort_controllers/include/effort_controllers/joint_position_controller.h
+++ b/effort_controllers/include/effort_controllers/joint_position_controller.h
@@ -98,14 +98,17 @@ public:
    *
    * \param robot The specific hardware interface used by this controller.
    *
-   * \param n A NodeHandle in the namespace from which the controller
+   * \param root_nh A NodeHandle in the root of the controller manager namespace.
+   * This is where the ROS interfaces are setup (publishers, subscribers, services).
+   *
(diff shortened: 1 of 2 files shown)


graiola:
Hello,
I think this could be a possible solution to issue #244 but as it's implemented it would break the interface.
A fix would be to use an overloaded method or a method with default argument.

@bmagyar what do you think?

bmagyar:
Good point. I think we have to also organise the effort of supporting non-global `robot_description` parameters in all of `ros_controllers` including a list of todos, extra tests to keep this feature in place and guidelines for people implementing new controllers to make sure they have this feature too.

bmagyar:
@shisha101 please make sure that your changes are API and ABI compatible with the current version as we cannot break these in Indigo.

shisha101:
Will do. I'll post back when it's ready.

bmagyar:
Hello,

I believe we should push this support through the same way (&same code) as it was done on #245. We https://github.com/ros/robot_model/pull/193 will allow for that.

As such I'm closing this PR.

shisha101:
Looks good, thanks for the heads up :)

emekBaris:
@bmagyar, this feature is missing in melodic-devel.

PR #250, opened 2016-12-23
Allow diff_drive_controller to use spheres as well as cylinders

mallanmba:
Allow diff_drive_controller to use spheres as well as cylinders for wheel collision geometry. Cylinders are not well behaved on Gazebo/ODE heightfields, using spheres works around the issue.

PR #259, opened 2017-03-13
Mallan/kinetic allow sphere wheels in diff drive

mallanmba:
repeat of pull request #250 on kinetic-devel


graiola:
Hello,

Thank you for sharing your code with us.
After checking your pr I noticed that in the .test files you use xacro.py instead of xacro but the former is deprecated. Could you change it?
Also, could you rename the test files related to the cylinder wheels to have consistent naming?

Thanks

bmagyar:
Bump. @mallanmba we could get this in to the next Kinetic and Lunar release. Those changes won't take you a lot of time to implement ;)

mallanmba:
Sorry! Dropped the ball on this. I'm happy to make the change for xacro.py,
but I'm unclear about the consistent naming - what are you looking for?
Thanks,
-Mark

graiola:
Hello!

The idea would be to change some files name inside the test folder:

- diffbot.xacro -> diffbot_cylinder_wheels.xacro

- wheel.xacro -> cylinder_wheel.xacro

Moreover, I think it would be nice to have a symmetric test for cylinder as for diff_drive_radius_sphere.test 

Note:
in diff_drive_radius_sphere.test the test-name should be different from the one in diff_drive_radius_param.test

@bmagyar , do you have any other advice?

bmagyar:
I'm not sure that `diff_drive_radius_sphere.test` should have different tests as it it practically running the same test on a different configuration.

tfoote:
Closing this as #278 included and extended the work of @mallanmba in this PR and has been merged.

bmagyar:
Thanks!

PR #262, opened 2017-03-28
Update joint_position_controller.cpp for resolving namespace

liuzjmike:
Update JointPositionController::init() to use urdf::Model::initWithNodeHandle so that robot_description can be correctly found within the namespace


liuzjmike:
Just realized that it's the same as #245

PR #264, opened 2017-04-06
fix behavior for multi-turn revolute joints

codebot:
Previously, the behavior of revolute joints was unexpected when
the joint limits extended beyond [-pi, pi] due to unwrapping
happening in a few places in the angles package, which was called
from the joint position controller. This PR checks to see if the
range of motion extends beyond [-pi, pi] and, if so, goes directly
to the (clamped) target, rather than trying to unwrap the current
or target angles to find the shortest path.

### DIFF
--- a/effort_controllers/src/joint_position_controller.cpp
+++ b/effort_controllers/src/joint_position_controller.cpp
@@ -181,12 +181,25 @@ void JointPositionController::update(const ros::Time& time, const ros::Duration&
   // Compute position error
   if (joint_urdf_->type == urdf::Joint::REVOLUTE)
   {
-   angles::shortest_angular_distance_with_limits(
-      current_position,
-      command_position,
-      joint_urdf_->limits->lower,
-      joint_urdf_->limits->upper,
-      error);
+    if (joint_urdf_->limits->lower >= -M_PI &&
+        joint_urdf_->limits->upper <=  M_PI) {
+      // this path only works when the joint travel is
+      // within [-pi,pi]
+      angles::shortest_angular_distance_with_limits(
+        current_position,
+        command_position,
+        joint_urdf_->limits->lower,
+        joint_urdf_->limits->upper,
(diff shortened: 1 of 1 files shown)


graiola:
Hello,

Could you show me a situation where the behavior is unexpected?
I am testing the code with and without the pr, but I don't see any difference.
Is the issue related to a different version of angles package? (I am testing with kinetic)

Thanks

graiola (changes requested):
ping @codebot

IanTheEngineer:
@codebot @graiola I've experienced the same issue within the default_robot_hw_sim interface from gazebo_ros_control:
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_ros_control/src/default_robot_hw_sim.cpp#L338-L342
Any measured angle from Gazebo that is outside of -pi to pi is forced to be within that range. This is a problem for Sawyer, as the seventh joint has a range -4.15 to 4.15 rad. I can handle fixing Sawyer, but we should address this for the larger community. 

Rather than fixing it here, shouldn't we just modify the angles [`shortest_angular_distance_with_limits`](https://github.com/ros/angles/blob/master/angles/include/angles/angles.h#L215) to handle larger joint limits properly?

bmagyar:
We love test driven development here. (I also like Sawyer ;))

Could you please add a test that replicates this? An alternative version of
rrbot could work.

IanTheEngineer:
@bmagyar hmm, on further investigation, I'm experiencing the joint limits being reported directly from Gazebo7 as being truncated to a reduced range. That doesn't mean that the shortest_angular_distance_with_limits is correct, but rather that this whole thing requires more investigation. I'll add more color (and possibly unit tests) to this as I grok the underlying issue better.

bmagyar:
I think we addressed this in #477 and #466 , please reopen if not.

PR #265, opened 2017-04-12
Remove rqt_plot test_depend

bmagyar:
Let's see how this works out.

It feels silly to be blocked by `rqt_common_plugins` when we only use `rqt_plot` to visualise tests.

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -56,7 +56,6 @@ if(CATKIN_ENABLE_TESTING)
       actionlib
       controller_manager
       xacro
-      rqt_plot
   )
 
   find_package(rostest REQUIRED)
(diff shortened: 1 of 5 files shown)


bmagyar:
Monitoring plots now don't show up by default, so tests pass on travis but can be enabled by passing in `display_plots:="true"` where available.
Example:
```
rostest joint_trajectory_controller joint_trajectory_controller_vel.test display_plots:="true"
```

Can I get a thumbs up & merge if ok please @ipa-mdl ?

mathias-luedtke (approved):
I don't really like to omit the `test_depend`, but I see the benefits.

joint_trajectory_controller/CMakeLists.txt
```diff
@@ -56,7 +56,6 @@ if(CATKIN_ENABLE_TESTING)
       actionlib
       controller_manager
       xacro
-      rqt_plot
```
mathias-luedtke, line 59: This didn't belong here anyway ;)

bmagyar:
Thanks!

PR #267, opened 2017-04-27
Publish executed velocity if publish_cmd

artivis:
Optionally publish the velocity command to be executed (`geometry_msgs/TwistStamped`).

Disabled by default.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -245,7 +245,10 @@ namespace diff_drive_controller{
 
     setOdomPubFields(root_nh, controller_nh);
 
-    cmd_vel_pub_.reset(new realtime_tools::RealtimePublisher<geometry_msgs::TwistStamped>(controller_nh, "cmd_vel_out", 100));
+    if (publish_cmd_)
+    {
+      cmd_vel_pub_.reset(new realtime_tools::RealtimePublisher<geometry_msgs::TwistStamped>(controller_nh, "cmd_vel_out", 100));
+    }
 
     // Get the joint object to use in the realtime loop
     for (int i = 0; i < wheel_joints_size_; ++i)
@@ -344,7 +347,7 @@ namespace diff_drive_controller{
     last0_cmd_ = curr_cmd;
 
     // Publish limited velocity:
-    if (publish_cmd_ && cmd_vel_pub_->trylock())
+    if (publish_cmd_ && cmd_vel_pub_ && cmd_vel_pub_->trylock())
     {
       cmd_vel_pub_->msg_.header.stamp = time;
       cmd_vel_pub_->msg_.twist.linear.x = curr_cmd.lin;
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -120,6 +120,7 @@ namespace diff_drive_controller{
     , odom_frame_id_("odom")
     , enable_odom_tf_(true)
     , wheel_joints_size_(0)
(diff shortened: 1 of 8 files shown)


artivis:
@bmagyar Hello Sir. Would you be kind enough to have a look ?

bmagyar:
I like it :+1: 
We should add to the wiki that this is to monitor what effect limiters may have on the control input. (Correct me if I'm wrong.)
Can we have a test please?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -241,6 +245,8 @@ namespace diff_drive_controller{
 
     setOdomPubFields(root_nh, controller_nh);
 
+    cmd_vel_pub_.reset(new realtime_tools::RealtimePublisher<geometry_msgs::TwistStamped>(controller_nh, "cmd_vel_out", 100));
```
artivis: Actually I guess I could even allocate only `if (publish_cmd_)` and change l.347  
`if (publish_cmd_ && cmd_vel_pub_->trylock())`  
by  
`if (publish_cmd_ && cmd_vel_pub_ && cmd_vel_pub_->trylock())`
@bmagyar Any opinion ?

artivis:
Here you go.  
Added a test to check that by default the velocity is not published, and one that checks that it is published if the param is set.

bmagyar (dismissed):
LGTM :+1:

bmagyar:
[diff_drive_controller.rosunit-diff_drive_default_cmd_vel_out_test/testDefaultCmdVelOutTopic][FAILURE]
/home/catkin_ws/src/ros_controllers/diff_drive_controller/test/diff_drive_default_cmd_vel_out_test.cpp:41
Value of: isPublishingCmdVelOut()
  Actual: true
Expected: false
--------------------------------------------------------------------------------


[diff_drive_controller.rosunit-diff_drive_default_cmd_vel_out_test/testDefaultCmdVelOutTopic][FAILURE]
/home/catkin_ws/src/ros_controllers/diff_drive_controller/test/diff_drive_default_cmd_vel_out_test.cpp:54
Value of: isPublishingCmdVelOut()
  Actual: true
Expected: false
--------------------------------------------------------------------------------


SUMMARY
 * RESULT: FAIL
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 1

graiola:
I had the same failure.

Could it be related to a sync problem between publisher and subscriber?

artivis (commented):
ping

artivis:
Ah crap it's my fault.  
So the point is that I made a comment in the code and developed the tests accordingly to it...   
Please have a look at the comment and let me know if I should update the code or else adapt the tests.

diff_drive_controller/src/diff_drive_controller.cpp
artivis: @bmagyar @graiola So as of now the controller do instanciate the publisher, therefore the topic. This modification would allow the controller to instantiate iff `publish_cmd_` is `true`.
It is the reason the `default` test fails now.  
The question is then, do I include the modification or fix the test ? What do you think ?

bmagyar: Yeah, not allocating is probably better. Allocating it would create a topic that never gets published on.

artivis: :+1: I fix that right away.

artivis:
Instantiate `cmd_vel_pub_` iff `publish_cmd_`.  
That fixes default test.

bmagyar:
Fiddling around to force travis to refresh job

graiola (approved):
Travis is happy :+1:

PR #270, opened 2017-04-28
Disable angular jerk limit test

bmagyar:
This is a cherry-pick from Kinetic that let's us get continuous integration up&running on Indigo too.


graiola (approved):
Ok!

PR #271, opened 2017-05-05
Set hold trajectory goal handle when empty trajectory action is received.

miguelprada:
Previously, an empty trajectory received through the action interface would set hold trajectory and accept the action goal, but the action would never be terminated, leaving clients hanging.

In #247 it was suggested to amend this by skipping the empty trajectory altogether and rejecting the action, while keeping the expected behavior (stop the running trajectory) for empty trajectories received through the topic interface. However, @ipa-mdl [suggested](https://github.com/ros-controls/ros_controllers/pull/247#issuecomment-298638630) that this created a discrepancy in the behaviors of the topic and action interfaces.

This PR tries to address this concern by implementing the same behavior in both interfaces, and fixing the hanging client issue by setting the goal handle associated to the hold trajectory to that of the empty trajectory goal.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -244,7 +244,7 @@ private:
    * current position and zero velocity.
    * \note This method is realtime-safe.
    */
-  void setHoldPosition(const ros::Time& time);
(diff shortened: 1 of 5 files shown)


mathias-luedtke:
Great! This is exactly what I had in mind :+1: 
The code looks good, but I don't have time right now to review the test code.

mathias-luedtke (approved):
Good job!
I have even run the tests without the patches and they failed, just as expected :)

bmagyar:
I don't have anything to add...nice work @miguelprada !

PR #273, opened 2017-05-10
Exposed the joint trajectory controller verbose setting as a parameter.

bponsler:
Expose the joint trajectory controller verbose setting as a parameter

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -183,7 +183,17 @@ JointTrajectoryController<SegmentImpl, HardwareInterface>::
 JointTrajectoryController()
   : verbose_(false), // Set to true during debugging
     hold_trajectory_ptr_(new Trajectory)
-{}
+{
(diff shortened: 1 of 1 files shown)


bmagyar:
This is something I think should be documented well. The verbose parameter there is for advanced use as it breaks real-time safety by using ROS logging services. I would not "open it to the public" as a parameter. In it's current state there has to be a strong need to go debugging the controller and the one seeking it will find it. 

What do you think? 

What do you think, gentlemen? @graiola @ipa-mdl

bponsler:
That seems reasonable, totally fine if you want to pass on this.

bmagyar:
Maybe you could still change this PR so that instead of opening it as a param, you add a comment in the code to explain that the verbose flag possibly breaks realtime-safety.

bponsler:
I'm fine with adding a comment, but I would suggest also displaying a runtime warning to the user whenever the verbose flag is enabled to help prevent someone from accidentally leaving the flag enabled.

Thoughts?

bmagyar (approved):

bmagyar:
There is a failing test on travis due a timeout but this code it only affecting a print. Merging...

Issue #274, opened 2017-05-13
Incorrect trajectory produced when value wraps from -PI to PI


bponsler:
I've noticed several cases where the joint trajectory controller produces an incorrect trajectory resulting in a joint not reaching the desired end position (although the JTC thinks that it has reached the goal position and succeeds). This occurs for a continuous joint when the incoming trajectory starts at a wrapped position value (pi) while the current joint state (position being held) remains unwrapped (e.g., -pi).

I believe this may be related to issue #263 

Attached are:
    - trajectory_wrap_test.cpp -- test case highlighting the issue (this was used to produce the data for the graph)
    - trajectory_issue.png -- a graph of the data with the error
    - plot_error.py -- python script used to generate the graph

![trajectory_issue](https://cloud.githubusercontent.com/assets/1627139/26027847/37e5e120-37ca-11e7-9514-a5a81fc360eb.png)

Looking at the attached graph, the joint is being held at a position of -3.14159. Then a new trajectory comes in which starts its initial point at 3.141325538883086 due to wrapping. The thin red line displays the series of desired position values computed by the JTC. There is a clear hump at the beginning where the hold trajectory and the new trajectory are joined. The joint moves the wrong way (positive), then starts to come back down. This hump causes the final desired position to flatten out at -5.86344 which is roughly 24 degrees from the desired end position (0 radians from the trajectory, or -2 * PI). The JTC thinks it is reaching the correct position, but has not rotated far enough.

The thin green line in the graph is the set of desired positions from the JTC produced when I comment out the following line from joint_trajectory_segment.h:

    dist = next_position[i] > prev_position[i] ? std::abs(dist) : -std::abs(dist);

I commented this out because I believe that the if statement surrounding it might be incorrect. It reads:

    if (std::abs(dist) - M_PI < 1e-9)

I _believe_ the intent is to say that if the distance is equal to pi within the tolerance, however, this will return true for any value of dist between [pi, -pi]. Perhaps I misunderstand the intent of this?

I think this line should be:

    if (std::abs(dist - M_PI) < 1e-9)

Once I removed the line handling the "singularity" the produced trajectory matches the desired trajectory and reaches the correct final position.

[trajectory_wrap_bug.zip](https://github.com/ros-controls/ros_controllers/files/998830/trajectory_wrap_bug.zip)

Issue #263, opened 2017-04-05
joint_trajectory_controller - wraparoundOffset


progtologist:
Perhaps I have misunderstood the usage of the wraparoundOffset(...) function, but I think the math is wrong in it. The simple version of the math is:
```cpp
double dist = angles::shortest_angular_distance(previous, next);
if (std::abs(dist) - M_PI < 1e-9)
{
    dist = next > previous ? std::abs(dist) : -std::abs(dist);
}
offset = (previous + dist) - next;
```
So if one queries for a rotation starting from -90 degrees to 180 degrees, the offset would then be equal to -180.
Therefore the commanded "next" position would evaluate to:
```cpp
next.position = next.position + offset;
```
0 degrees, which is not correct.

Am I missing something here?

Issue #275, opened 2017-05-13
State error calculation not accounting for wrapping joint positions


bponsler:
The following line in joint_trajectory_controller_impl.h is not properly handling wrapping joint angles:

    state_error_.position[i] = desired_state_.position[i] - current_state_.position[i];

If the current joint position changes from -PI to PI this calculation will produce a large difference value which causes the joint to move very fast toward that new value.

I believe the following would properly account for this issue:

    state_error_.position[i] = angles::shortest_angular_distance(current_state_.position[i] - desired_state_.position[i]);

graiola:
Hi Brett,

Thanks for the detailed report! (#274)
Could I ask you to open a pull request with the code you've attached to this issue, feel free to add a test to our current test-suite. We should see how the rest of the tests behave.

PR #276, opened 2017-05-15
Address issue #263, joint_trajectory_controller - wraparoundOffset

graiola:
This PR should address both #263 and #274.
@progtologist , @bponsler Could you try it and see if it solves your issue? Thanks.
Note that this is for kinetic, but you should be able to cherry-pick it for indigo.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_segment.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_segment.h
@@ -230,7 +230,7 @@ Scalar wraparoundJointOffset(const Scalar& prev_position,
     Scalar dist = angles::shortest_angular_distance(prev_position, next_position);
 
     // Deal with singularity at M_PI shortest distance
-    if (std::abs(dist) - M_PI < 1e-9)
+    if (std::abs(std::abs(dist) - M_PI) < 1e-9)
     {
       dist = next_position > prev_position ? std::abs(dist) : -std::abs(dist);
     }
--- a/joint_trajectory_controller/test/joint_trajectory_segment_test.cpp
+++ b/joint_trajectory_controller/test/joint_trajectory_segment_test.cpp
@@ -87,6 +87,14 @@ TEST(WraparoundOffsetTest, WrappingPositions)
       EXPECT_NEAR(two_pi, wraparound_offset1, EPS);
       EXPECT_NEAR(0.0, wraparound_offset2, EPS);
     }
+
+    // Special case, from M_PI/2 to -M_PI
+    double wraparound_offset1 = wraparoundJointOffset(M_PI/2, -M_PI, true);
+    EXPECT_NEAR(two_pi, wraparound_offset1, EPS);
+    // From -M_PI to M_PI/2
+    wraparound_offset1 = wraparoundJointOffset(-M_PI, M_PI/2, true);
+    EXPECT_NEAR(-two_pi, wraparound_offset1, EPS);
+
   }
 
   // Both wrapping joints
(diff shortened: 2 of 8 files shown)


joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_segment.h
```diff
@@ -222,15 +222,15 @@ Scalar wraparoundJointOffset(const Scalar& prev_position,
                              const Scalar& next_position,
                              const bool&   angle_wraparound)
 {
-  // Return value
+   // Return value
```
bmagyar: There's no reason to change this line

joint_trajectory_controller/test/joint_trajectory_segment_test.cpp
```diff
@@ -108,6 +113,7 @@ TEST(WraparoundOffsetTest, WrappingPositions)
       EXPECT_NEAR(two_pi, wraparound_offset1, EPS);
       EXPECT_NEAR(2.0 * two_pi, wraparound_offset2, EPS);
     }
+
```
bmagyar: There's no reason to change this line

joint_trajectory_controller/test/joint_trajectory_segment_test.cpp
```diff
@@ -382,3 +388,4 @@ int main(int argc, char** argv)
   return RUN_ALL_TESTS();
 }
 
+
```
bmagyar: There's no reason to change this line

joint_trajectory_controller/test/joint_trajectory_segment_test.cpp
```diff
@@ -87,6 +87,11 @@ TEST(WraparoundOffsetTest, WrappingPositions)
       EXPECT_NEAR(two_pi, wraparound_offset1, EPS);
       EXPECT_NEAR(0.0, wraparound_offset2, EPS);
     }
+
+    // Special case
+    double wraparound_offset1 = wraparoundJointOffset(M_PI/2, -M_PI, true);
+    EXPECT_NEAR(two_pi, wraparound_offset1, EPS);
+
```
mathias-luedtke: What about checking `wraparoundJointOffset(-M_PI/2, M_PI, true);` as well?

bponsler:
Thanks for doing this @graiola! I wasn't going to be able to pull this together until later today.

I tested these changes on indigo and they solve the issue I was seeing before. You can see in the below graph that the desired state (red line) no longer has the upward bump seen in #274.

screenshot7

graiola:
@bponsler  no problem! I am happy that the pr solved the problem ;)
@ipa-mdl could you run some tests with your robot/s to validate the fix? Thanks.

mathias-luedtke:
> @ipa-mdl could you run some tests with your robot/s to validate the fix? Thanks.

@graiola: I am sorry, I cannot test it with hardware. I don't have any right now and we don't use wrap-around configs.

graiola:
@progtologist could you give it a shot? Thanks.

progtologist:
@graiola I cannot test it with hardware right now but it runs correctly in simulation so I see no reason why it shouldn't work with hardware as well! Great job!

bmagyar (changes requested):
Ping

bmagyar (approved):

bmagyar:
We are almost there! ;) 

Could you tell me what's the difference between the two "Added tests for issue #275" commits? They look the same to me. 
Could you remove one of them and the "merge branch blabla" commit as well?

graiola:
I removed the rqt_plot because it was causing problems with travis.

bmagyar:
Great, thanks!

PR #277, opened 2017-05-16
Fix for issue #275

bponsler:
Fix for issue #275 (State error calculation not accounting for wrapping joint positions).

Created an additional test case to test a robot with wrapping continuous joint angles.

### DIFF
--- a/joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
+++ b/joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
@@ -1,0 +1,279 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2013, PAL Robotics S.L.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the name of PAL Robotics S.L. nor the names of its
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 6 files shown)


bponsler:
This is for indigo since I don't have a quick way to compile/test for kinetic.

mathias-luedtke:
I have restarted [one job](https://travis-ci.org/ros-controls/ros_controllers/jobs/232655977) because of an unrelated error:
```
[diff_drive_controller.rosunit-diff_drive_odom_frame_test/testOdomTopic][FAILURE]
/root/catkin_ws/src/ros_controllers/diff_drive_controller/test/diff_drive_odom_frame_test.cpp:73
Value of: "new_odom"
Expected: odom_msg.header.frame_id.c_str()
Which is: ""
```

joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
```diff
@@ -0,0 +1,279 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2013, PAL Robotics S.L.
```
bmagyar, line 2: You may want to adjust the name here

graiola:
I try to run the fix on kinetic to see how it behaves...


... tests are fine on kinetic.

bmagyar:
Are you running this on Indigo? Do you have trouble on Indigo because of this? 
I'm a bit reluctant to add new stuff to Indigo that potentially changes behaviour but this can be considered bug fixing.

bmagyar (changes requested):

joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
bponsler: Who do you want this copyright attributed to? I'll also adjust the year to 2017 unless you disagree.

graiola:
Thanks for the pr @bponsler ! Could you please split the commit between the fix and the test? In this way we can cherry-pick the test on kinetic.

bponsler:
Sure thing. Let me know if you want me to do anything else

graiola:
Thank you Brett!

joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
bmagyar: Up to you. If it was on your own time, you can leave it like that or if it's within a company you may want to add them there. It's a test, hence it doesn't matter much.

bponsler: I'm fine with it as is, thanks for the suggestion though!

bmagyar (approved):

bmagyar:
Thanks Brett!

Issue #55, opened 2013-10-03
We need mobile base controllers


adolfo-rt:
All,

To make the ros_controll(ers) project appealing to a larger user base, we need to offer functionality needed by that user base, and controllers are a big part of that. Currently, the list of available general-purpose controllers is still quite limited, and we have a big zero on mobile base / vehicle controllers.

I've convinced a colleague to make a **diff-drive** controller that should land sometime soon, but I wonder about the **interest** and **generality** of other implementations. 
### Interest

Does it make sense to have mobile base controllers inside the (possibly high-frequency) `controller_manager` loop, or is it enough to control individual wheels with low-level single-joint controllers?. Many navigation scenarios have relatively slow dynamics, but I can imagine many scenarios (self-driving cars!) having requirements for faster dynamics.

Further, having a single ros_control plugin that takes in higher level navigation commands results in simpler deployments than single_joint controllers + an additional node for translating navigation commands to joint commands.
### Generality

Implementations like the **diff-drive** generalize well to multiple robot platforms. Does the same apply to an **Ackermann-type** controller, are they easily configurable (@jim-rothrock, @jack-oquin)?.  How about **omnidirectional** bases?, these seem more complex to generalize, is it the case?. Are there other controller types that would be of general interest?.

jack-oquin:
My personal experience with Ackermann navigation is limited to a single no-longer-operational vehicle.

From [reviews and discussions](http://wiki.ros.org/Ackermann%20Group/January%2029%2C%202012) conducted by the [ROS Ackermann Group](http://wiki.ros.org/Ackermann%20Group), I believe the [ackermann_msgs](http://wiki.ros.org/ackermann_msgs) definitions are sufficiently general for any four-wheeled vehicle steering only with the front wheels. Some three-wheel vehicles would probably work, too. 

I am encouraged to see the recent announcement of [wunderkammer-laboratory/ackermann_vehicle](https://github.com/wunderkammer-laboratory/ackermann_vehicle), a new Gazebo-based simulation for Ackermann drive vehicles.

Integrating Ackermann steering into the standard [ros-planning/navigation](https://github.com/ros-planning/navigation) packages appears challenging. I doubt that planning in the configuration space of an Ackermann vehicle fits neatly into their current implementation.

jack-oquin:
Is there a difference between **omnidirectional** and **holonomic**? 

I think ROS navigation already handles that for PR2 support.

jim-rothrock:
[...]

PR #280, opened 2017-06-16
Add four_wheel_steering_controller

vincentrou:
This PR depends on this new package : https://github.com/ros-controls/urdf_geometry_parser/pull/2

ref #55


four_wheel_steering_controller/scripts/teleop_four_wheel_steering_joy.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
+import rospy
+from math import pi
+from four_wheel_steering_msgs.msg import FourWheelSteering
+from sensor_msgs.msg import Joy
+
+class TeleopFourWheelSteeringJoy():
```
bmagyar: could you please remove this from this PR? let's try to minimise the scope

four_wheel_steering_controller/include/four_wheel_steering_controller/speed_limiter.h
```diff
@@ -0,0 +1,137 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2013, PAL Robotics, S.L.
```
bmagyar, line 4: this is a lot of copied code from the `diff_drive_controller`, could you use that one instead?

four_wheel_steering_controller/src/odometry.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  void Odometry::init(const ros::Time& time)
+  {
+    // Reset accumulators and timestamp:
+    resetAccumulators();
+    timestamp_ = time;
+  }
+
+
+  bool Odometry::update(const double &fl_speed, const double &fr_speed,
```
bmagyar: param setup and update functions are different, but integration and interface is the same as `diff_drive_controller`. Maybe we can make a super class?

four_wheel_steering_controller/test/config/view_odom.rviz
```diff
@@ -0,0 +1,158 @@
+Panels:
```
bmagyar: please don't put this in

bmagyar (changes requested):

four_wheel_steering_controller/include/four_wheel_steering_controller/speed_limiter.h
vincentrou: I agree with you but is it fine to depend on the diff_drive_controller ? Or is it better to create another package with speed_limiter and odometry library ?

four_wheel_steering_controller/src/odometry.cpp
vincentrou: I agree but maybe in a new package ? (as said above)

MoriKen254:
Hello. I also worked on steering drive controller stuff and opened pull request (#289) but they suggested I should join here to avoid duplication. Please check information about my current project on #289. We provided sample demonstration application and documentation as well.

I also want to contribute to this pull request for simplification, extension, documentation and so on, if you don't mind.

Now I'm checking `four_wheel_steering_controller` codes to compare with my `steer_drive_controller` and trying to run it in my PC. I believe both projects share similar concept but I found an important difference. 

---

 For instance, the super class of mine is different from yours: my `steer_drive_controller` inherits `MultiInterfaceController` (available from Jade) which initially handles multiple interfaces such as both position and velocity ones, while `four_wheel_steering_controller` does `ControllerBase` and looks resolve multiple interfaces initialization in itself.

So I believe using `MultiInterfaceController` way will avoid code duplication and make the codes simpler.

Like this, I might be able to contribute to something on this request from my struggles on  `steer_drive_controller`.

---

By the way, I also agree with an idea to create a super class (such as `wheels_drive_controller` ?) to avoid copying codes from `diff_drive_controller`. Maybe we should solve this one first. Please let me work on some modifications you haven't started yet.

Thank you.

vincentrou:
I agree the use of `MultiInterfaceController` will simplify the code. But I developed it for indigo and kinetic.
I think this package will be released just for kinetic so we can use the `MultiInterfaceController`.

MoriKen254:
@vincentrou 

Thank you for your reply! That makes sense if it's compatible with indigo. As you mentioned, if you just release it for kinetic `MultiInterfaceController` would be good.

By the way, I've read your codes and found a difference from mine. I posted a detail on #289 so please have a look at it if you are interested.

vincentrou:
As I mentioned in the PR #289 , I don't think it is relevant to merge it with this one.
@bmagyar is it possible to go forward with this PR ?

four_wheel_steering_controller/include/four_wheel_steering_controller/speed_limiter.h
bmagyar: Maybe better not...I'm a bit hesitant on this: methodically that's the right choice but from a release and maintenance point of view it's better if things are not that entangled. 

@ipa-mdl what do you think?

bmagyar:
I've restarted the tests. Bear in mind that reviewing this amount of code will take a bit of time.

four_wheel_steering_controller/src/four_wheel_steering_controller.cpp
```diff
@@ -220,26 +220,26 @@ namespace four_wheel_steering_controller{
...
         return false;
       else
-        controller_nh.setParam("wheel_radius",wheel_radius_);
+        root_nh.setParam("wheel_radius",wheel_radius_);
     if(lookup_wheel_base)
       if(!uvk.getDistanceBetweenJoints(front_wheel_names[0], rear_wheel_names[0], wheel_base_))
         return false;
       else
-        controller_nh.setParam("wheel_base",wheel_base_);
+        root_nh.setParam("wheel_base",wheel_base_);
```
bmagyar: I would remove this commit...`root_nh` here points to global param space, right? 
Those parameters should _especially_ be in a namespaced environment

four_wheel_steering_controller/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(four_wheel_steering_controller)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+
+#Add custom (non compiling) targets so launch scripts and python files show up in QT Creator's project view.
+file(GLOB_RECURSE EXTRA_FILES */*)
```
bmagyar: Please remove these changes, we shouldn't include qtcreator hacks in the cmake of released packages.

four_wheel_steering_controller/src/four_wheel_steering_controller.cpp
vincentrou: The thing is that I want to use those parameters in other nodes (so they don't have to parse again the urdf).
When the parameters are in the global space, other nodes do not need to know the controller node name to find them...
Maybe I could use service instead.

four_wheel_steering_controller/CMakeLists.txt
vincentrou: agreed

four_wheel_steering_controller/src/four_wheel_steering_controller.cpp
bmagyar: The param server is exactly for this kind of stuff. And there's nothing wrong with having to know the name of the controller node, that's not something you'd change anyways. I usually use `mobile_base_controller` as name for that on most robots regardless of what exact type of base controller that is.

bmagyar: Note with researcher hat on: Keep in mind that services for your experiments are very bad as they are not transparent to anything. At least rosparam can be dumped to a file for processing later.

bmagyar:
There's a failing tests that remains: 
```
[four_wheel_steering_controller.rosunit-four_wheel_steering_controller_4ws_cmd_test/testCrab][FAILURE]
/root/catkin_ws/src/ros_controllers/four_wheel_steering_controller/test/src/four_wheel_steering_4ws_cmd_test.cpp:84
The difference between sqrt(dx*dx + dy*dy) and cmd_vel.speed*travel_time is 0.01000576054909641, which exceeds POSITION_TOLERANCE, where sqrt(dx*dx + dy*dy) evaluates to 1.5899942632927615, cmd_vel.speed*travel_time evaluates to 1.6000000238418579, and POSITION_TOLERANCE evaluates to 0.01.
```

Just add some eps to account for this and I think we can merge.

vincentrou:
The tests are passing now for four_wheel_steering_controller.
But there is one error with the diff_drive_controller on some builds.
Maybe it is because I do not update my branch...

vincentrou:
Is there any other change requested for this PR ?

bmagyar (approved):

bmagyar:
It is all ok now, thanks for being so patient and stick all the way through!
Merging...

Maik93:
Hi all,
just cloned `ros_controllers` repo on a fresh Kinetic installation (cloning and switching to `kinetic_devel` branch). When I run `catkin_make` I encounter this error:

> Could not find a package configuration file provided by "four_wheel_steering_msgs" with any of the following names:
> four_wheel_steering_msgsConfig.cmake
> four_wheel_steering_msgs-config.cmake
> ...

Inside `four_wheel_steering` directory, there's no msgs, so I don't know what to do; by now I've removed the entire directory (I don't need this package at the moment), and everything works smoothly, I'm just here to tell you there's something incomplete in this pull request, I believe.

Don't know if open a new issue too, you tell me.

bmagyar:
You simply forgot to install the dependencies. There is nothing incomplete in this pull request.
To make your life easier next time use `rosdep install --from-paths src --ignore-src`.

Maik93:
Happy to hear that everything's ok then. I can confirm that it works now (even though there are some problem due to manifest's encoding if using python 2.7, as pointed out [here](https://github.com/ros-infrastructure/rospkg/issues/154)).

An alternative command to yours would have been this: `rosdep install four_wheel_steering_controller`.

Thanks again for your help and sorry for the misleading issue.

PR #282, opened 2017-06-27
Add travis state of supported releases

bmagyar:


graiola (commented):
Is there a reason why the build status for kinetic is repeated? Did you mean "lunar"? :)

bmagyar:
As of now Kinetic and Lunar releases are the same but didn't want to leave Lunar off, it'd feel as if we are not supporting it

graiola (approved):

PR #283, opened 2017-06-28
Avoid nan command to be processed

vincentrou:


bmagyar:
Ouch, great find!
Could you please also add a test that fails if this feature is not implemented?

vincentrou:
When I send a twist with a NaN in the linear.x, the robot goes backward full speed in gazebo and in real.
But I do not reproduce it with the test class... (published odom always at zero)

bmagyar:
This will need a rebase and that test. Sorry but I'd really prefer not to add any fixes that are not backed up by a test.

bmagyar:
I tried to run the test without the fix and I can confirm that this test doesn't really catch the issue. 

Could you please try setting the param `publish_cmd` and see what gets through there?

For more: https://github.com/ros-controls/ros_controllers/blob/kinetic-devel/diff_drive_controller/src/diff_drive_controller.cpp#L292

bmagyar:
Also we've got a conflict. Could you rebase the PR on the latest `kinetic-devel` please?

bmagyar:
Ping @vincentrou

vincentrou:
I rebase the PR on kinect in december but I will not have time to do more on this.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -451,6 +451,11 @@ namespace diff_drive_controller{
         return;
       }
 
+      if(std::isnan(command.angular.z) || std::isnan(command.linear.x))
```
artivis, line 454: How about using `!std::isfinite` instead of `std::isnan` so that you could catch +/- Inf too ?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -451,6 +451,11 @@ namespace diff_drive_controller{
...
+      {
+        ROS_WARN("Received NaN in geometry_msgs::Twist. Ignoring command.");
```
bmagyar, line 456: This should probably be throttled to avoid 200Hz prints if someone spams the controllers

diff_drive_controller/test/diff_drive_nan_test.cpp
```diff
@@ -78,6 +78,43 @@ TEST_F(DiffDriveControllerTest, testNaN)
...
+
+  // send NaN
+  for(int i=0; i<100;++i)
+  {
+    geometry_msgs::Twist cmd_vel;
+    cmd_vel.linear.x = NAN;
+    cmd_vel.angular.z = 0.0;
+    publish(cmd_vel);
+    geometry_msgs::TwistStamped odom_msg = getLastCmdVelOut();
+    ROS_INFO_STREAM("cmd_vel_out.twist.linear.x "<<odom_msg.twist.linear.x);
```
bmagyar, line 103: should check here for NaN instead of printing

diff_drive_controller/src/diff_drive_controller.cpp
bmagyar: sehr gut

martiniil:
I picked these commits for melodic. However I see a problem: The robot gets stuck during execution of the previous test case `testNaN` here
```
  // stop robot (will generate NaN)
  stop();
```
Due to
```
        const double lp = left_wheel_joints_[i].getPosition();
        const double rp = right_wheel_joints_[i].getPosition();
        if (std::isnan(lp) || std::isnan(rp))
          return;
```
the commands are no longer updated even after `start()`. I am not sure what the intended behaviour (of the controller) is here.

martiniil:
> I tried to run the test without the fix and I can confirm that this test doesn't really catch the issue.
> 
> Could you please try setting the param `publish_cmd` and see what gets through there?
> 
> For more: https://github.com/ros-controls/ros_controllers/blob/kinetic-devel/diff_drive_controller/src/diff_drive_controller.cpp#L292

The test passes since the command is never updated during this test case. See my comment above.

bmagyar:
Closing this in favour of #469

PR #288, opened 2017-07-05
[diff_drive_controller] odom_frame_id parametrized

asoriano1:
The odom_frame_id was fixed to "odom". The code has been modified a little in order to have the opportunity to change it by parameter.


bmagyar:
It is not encouraged to straight copy code around. Read [here for more](http://alblue.bandlem.com/2011/06/git-tip-of-week-cherry-picking.html) .

These are not your changes but you copied them from the `indigo-devel` branch but without the tests.
I've done the same a few days ago, this PR is a duplicate of #286 . We need a bit more time to consolidate the build there and we are good to go.

PR #293, opened 2017-08-19
export odom params if not on server

artivis:
Simply export the odometry parameters (wheel separation, radius and respective multipliers), if they are not already on the param server.
@bmagyar @efernandez Gentle ping for an easy one.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -234,6 +234,24 @@ namespace diff_drive_controller{
       return false;
     }
 
+    // Export parameters to param server if not available
+    if (lookup_wheel_separation)
+    {
+      controller_nh.setParam("wheel_separation", wheel_separation_);
+    }
+    if (lookup_wheel_radius)
+    {
+      controller_nh.setParam("wheel_radius", wheel_radius_);
+    }
+    if (!controller_nh.hasParam("wheel_separation_multiplier"))
+    {
+      controller_nh.setParam("wheel_separation_multiplier", wheel_separation_multiplier_);
+    }
+    if (!controller_nh.hasParam("wheel_radius_multiplier"))
+    {
+      controller_nh.setParam("wheel_radius_multiplier", wheel_radius_multiplier_);
+    }
+
     // Regardless of how we got the separation and radius, use them
     // to set the odometry parameters
     const double ws = wheel_separation_multiplier_ * wheel_separation_;


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -234,6 +234,24 @@ namespace diff_drive_controller{
       return false;
     }
 
+    // Export parameters to param server if not available
+    if (lookup_wheel_separation)
+    {
+      controller_nh.setParam("wheel_separation", wheel_separation_);
```
efernandez, line 240: I'm a little worry about the side effects of setting params from here.

For example, if you re-run the controller, it'd take the params set here, which might be unexpected by some users if they're playing with the lookup values instead.

bmagyar:
While I understand the motivation,  I'm a bit uncertain too. Values that
were explicitly pushed to the param server should persist but this may
allow for quiet accidents to happen.

Shouldn't we add a diagnostics message where the controller could publish
it's current parameters? That feels like a better way to display controller
internals.


On Aug 21, 2017 5:28 PM, "Enrique Fernández Perdomo" <
notifications@github.com> wrote:

*@efernandez* commented on this pull request.
------------------------------

In diff_drive_controller/src/diff_drive_controller.cpp
<https://github.com/ros-controls/ros_controllers/pull/293#discussion_r134276807>
:

> @@ -234,6 +234,24 @@ namespace diff_drive_controller{
       return false;
     }

+    // Export parameters to param server if not available
+    if (lookup_wheel_separation)
+    {
+      controller_nh.setParam("wheel_separation", wheel_separation_);

I'm a little worry about the side effects of setting params from here.

For example, if you re-run the controller, it'd take the params set here,
which might be unexpected by some users if they're playing with the lookup
values instead.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/ros-controls/ros_controllers/pull/293#pullrequestreview-57550016>,
or mute the thread

diff_drive_controller/src/diff_drive_controller.cpp
artivis: You are right. 
Maybe they could be set under a sub-namespace, say `/mobile_base_controller/runtime/those_parameters`.
Any better idea to export those params ?

artivis:
A diagnostic message could do the job I guess.

bmagyar:
I like the name space idea!

artivis:
Shall I update the PR with a sub-namespace ??
With `/runtime` ??

efernandez:
I can figure out your motivation behind this, but could you explain why you need this?
I wonder if you could just retrieve those values from somewhere else.

A diagnostic message as a diagnostic_updater is not a valid solution ATM because it's not real time, and if you publish only once, it'd become STALE.

That being said, if you still need this (which I guess you do b/c of the configuration automatically discovered w/o params), I'd like to consider the option of a diagnostic_updater that publishes the configuration once, from the constructor and later publishes something (a dummy thing for now) in a way that doesn't get STALE but also doesn't block the real-time update loop. That would be the cleaner way to do this IMHO, but I'd like to hear your thoughts on this.

efernandez:
For example, one solution could be a separate non-RT thread calling update on the diagnostic updater. For now, that should be fine, and will allow us to extend things in the future to provide diagnostics.
We could probably `readFromNonRT` (https://github.com/ros-controls/realtime_tools/blob/kinetic-devel/include/realtime_tools/realtime_buffer.h#L121) and that would work.

artivis:
The goal of this PR is obviously to be able to retrieve the parameters the diff_drive is configured with, at **any time**.
These parameters being static for the lifetime of the controller, I don't think there is a need for them to be published in a RT fashion. And I'm not sure to understand how can they become stale ? Do you mean that they would remain on the param server after the controller destruction ? This particular case can be handle at destruction with a `deleteParam` I guess.

bmagyar:
Generally what you'd like is to inspect what goes on inside the controller at any time. The motivation is figuring out how an experiment or test went. I think this is relevant to both research and industrial use. 
For this so far we have two solutions.

The question is which features something like this needs. Does it need to persist through a rosbag too?

artivis:
I do see the point of this being ' rosbagable ', however, a RT publisher spamming parameters that never change seems like an overkill ...
With these params being set on the param server, you still can dump it all in a yaml file...

efernandez:
> The goal of this PR is obviously to be able to retrieve the parameters the diff_drive is configured with, at any time.
The fact you want them "at any time" is important here.
My suggestion assume you can get them only at the startup, published on a first diagnostic message.

> These parameters being static for the lifetime of the controller, I don't think there is a need for them to be published in a RT fashion.
They don't need to be published in RT, since the diagnostic message will be updated on a non-RT method, I think.

> And I'm not sure to understand how can they become stale ?
This is just a concern if we finally use diagnostics.

> Do you mean that they would remain on the param server after the controller destruction ? This particular case can be handle at destruction with a deleteParam I guess.
Yes, that's what I mean, but I don't like touching the param server this way. It could confuse the user.

What about exposing them in the dynamc_reconfigure params? I'm not sure if that is already upstream, but I've added dynamic params to the diff drive controller either here or in the CPR fork.

If you don't like that idea, setting + deleting (if initially set) is what I'd do if I don't have another option.

artivis:
Indeed I don't like much the idea of the dynamic_reconfigure...  
Does it actually update the diff-drive parameter 'online' if they change through the dyn-reconf ? I'll have a look at the CRP fork...  
I guess the proper mechanism for this task would be a rosservice with a `getParam` or such.

bmagyar:
I'd prefer the param over dynamic reconfigure. There's a set of mechanisms
there that's better not disturbed for other things it was meant for.

artivis:
Closing in favor to #320.

PR #297, opened 2017-09-11
TrajectoryController: Use desired state to calculate hold trajectory

agutenkunst:
The current state is used for calculating the hold-trajectory, that results in a jump in the desired position. The time delay originating from the canopen communication leads to a desired position that is already in the past.

Below the effect can be clearly seen by comparing the trajectories (plotted from the CAN traces).

kinetic-vs-fix_06

It can be clearly seen that the desired trajectory switches the sign of the velocity. This results in a jerk of the respective drive. If the speed is high and/or the stop_time is reduced this effect increases.
I would also argue that this behavior causes mechanical stress in the robot.
 
Any opinions?

_Noticed with Schunk [LWA4p](https://github.com/ipa320/schunk_modular_robotics/tree/indigo_dev/schunk_description/urdf/lwa4p)_

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -748,6 +748,7 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
   typename Segment::State hold_start_state_ = typename Segment::State(1);
   typename Segment::State hold_end_state_ = typename Segment::State(1);
   const unsigned int n_joints = joints_.size();
+  const typename Segment::Time start_time  = time.toSec();
 
   if(stop_trajectory_duration_ == 0.0)
   {
@@ -757,8 +758,8 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
       hold_start_state_.position[0]     =  joints_[i].getPosition();
       hold_start_state_.velocity[0]     =  0.0;
       hold_start_state_.acceleration[0] =  0.0;
-      (*hold_trajectory_ptr_)[i].front().init(time.toSec(),  hold_start_state_,
-                                                           time.toSec(), hold_start_state_);
+      (*hold_trajectory_ptr_)[i].front().init(start_time,  hold_start_state_,
+                                              start_time, hold_start_state_);
       // Set goal handle for the segment
       (*hold_trajectory_ptr_)[i].front().setGoalHandle(gh);
     }
@@ -771,7 +772,6 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
     // - Create segment that goes from current state to above zero velocity state, in the desired time
     // NOTE: The symmetry assumption from the second point above might not hold for all possible segment types
 
-    const typename Segment::Time start_time  = time.toSec();
     const typename Segment::Time end_time    = time.toSec() + stop_trajectory_duration_;
     const typename Segment::Time end_time_2x = time.toSec() + 2.0 * stop_trajectory_duration_;
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -743,47 +743,64 @@ template <class SegmentImpl, class HardwareInterface>
 void JointTrajectoryController<SegmentImpl, HardwareInterface>::
 setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
 {
-  // Settle position in a fixed time. We do the following:
-  // - Create segment that goes from current (pos,vel) to (pos,-vel) in 2x the desired stop time
-  // - Assuming segment symmetry, sample segment at its midpoint (desired stop time). It should have zero velocity
-  // - Create segment that goes from current state to above zero velocity state, in the desired time
-  // NOTE: The symmetry assumption from the second point above might not hold for all possible segment types
-
   assert(joint_names_.size() == hold_trajectory_ptr_->size());
 
   typename Segment::State hold_start_state_ = typename Segment::State(1);
   typename Segment::State hold_end_state_ = typename Segment::State(1);
-
-  const typename Segment::Time start_time  = time.toSec();
-  const typename Segment::Time end_time    = time.toSec() + stop_trajectory_duration_;
-  const typename Segment::Time end_time_2x = time.toSec() + 2.0 * stop_trajectory_duration_;
-
-  // Create segment that goes from current (pos,vel) to (pos,-vel)
   const unsigned int n_joints = joints_.size();
-  for (unsigned int i = 0; i < n_joints; ++i)
+
+  if(stop_trajectory_duration_ == 0.0)
   {
-    // If there is a time delay in the system it is better to use calculate the hold trajectory starting from the
-    // desired position. Otherwise there would be a jerk in the motion.
-    hold_start_state_.position[0]     =  desired_state_.position[i];
-    hold_start_state_.velocity[0]     =  desired_state_.velocity[i];
-    hold_start_state_.acceleration[0] =  0.0;
+    // stop at current actual position
+    for (unsigned int i = 0; i < n_joints; ++i)
+    {
+      hold_start_state_.position[0]     =  joints_[i].getPosition();
+      hold_start_state_.velocity[0]     =  0.0;
+      hold_start_state_.acceleration[0] =  0.0;
+      (*hold_trajectory_ptr_)[i].front().init(time.toSec(),  hold_start_state_,
+                                                           time.toSec(), hold_start_state_);
+      // Set goal handle for the segment
+      (*hold_trajectory_ptr_)[i].front().setGoalHandle(gh);
+    }
+  }
+  else
+  {
+    // Settle position in a fixed time. We do the following:
+    // - Create segment that goes from current (pos,vel) to (pos,-vel) in 2x the desired stop time
+    // - Assuming segment symmetry, sample segment at its midpoint (desired stop time). It should have zero velocity
+    // - Create segment that goes from current state to above zero velocity state, in the desired time
+    // NOTE: The symmetry assumption from the second point above might not hold for all possible segment types
 
-    hold_end_state_.position[0]       =  desired_state_.position[i];
-    hold_end_state_.velocity[0]       = -desired_state_.velocity[i];
-    hold_end_state_.acceleration[0]   =  0.0;
+    const typename Segment::Time start_time  = time.toSec();
+    const typename Segment::Time end_time    = time.toSec() + stop_trajectory_duration_;
+    const typename Segment::Time end_time_2x = time.toSec() + 2.0 * stop_trajectory_duration_;
 
-    (*hold_trajectory_ptr_)[i].front().init(start_time,  hold_start_state_,
-                                                             end_time_2x, hold_end_state_);
+    // Create segment that goes from current (pos,vel) to (pos,-vel)
+    for (unsigned int i = 0; i < n_joints; ++i)
+    {
+      // If there is a time delay in the system it is better to calculate the hold trajectory starting from the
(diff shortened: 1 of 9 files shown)


bmagyar:
By looking at the plot and some experience with robots using this I think you are onto something. 

We do have however a policy to add tests to every issue we encounter and then the fix for it. This makes sure that a newly discovered problem will stay fixed.

Could you please reproduce the issue within a test? Without your fix the test should fail. You can measure the jerk you mentioned.

mathias-luedtke:
I can confirm these jump movements in certain cases with the Schunk arms, but I haven't found out why.

In `ros_canopen` the state should be 2 cycles (-> 20ms) ahead of the command.
(command(t) gets sent at the same time as state(t+1) gets requested, the goal should be reached at state(t+2).)

I think the fix is correct. The last command should be kept in order to prevent jumps.

To further understand the effect and to provide a proper test:
What raw data are you plotting (and how)?
Can you plot as data points (e.g. small circles) or provide a bag file?
What's the input of the controller?
What's in the `state` topic (the CAN data might be affected by the limit interfaces)?

jschleicher:
I've added a testcase. Please Review and comment on the code / test.

New Plots from the testcase run with
`
roslaunch joint_trajectory_controller joint_trajectory_controller.test display_plots:=True
rosrun joint_trajectory_controller joint_trajectory_controller_test
`

testcase_jerk_on_stop_pos
testcase_jerk_on_stop_vel

This PR is the same as #176 and #177; we only searched the issues and thus opened a duplicate PR, sorry for that.

mathias-luedtke:
Reviewing the old discussions using the desired state might [lead to problems as well](https://github.com/ros-controls/ros_controllers/pull/177#issuecomment-99436441).
This is even worse because `setHoldPosition` gets called if the tracking error is too high.

As a compromise the hold target could get capped by the max. allowed tracking error.
Not sure if this really makes sense..

> I've added a testcase. Please Review and comment on the code / test.

The test looks good, but I haven't tried it yet.

The `delay` topic just enables or disables the delay, so this should get reflected in a  `std_msgs/Bool` topic or a `std_srvs//SetBool`service.
As an alternative the `UInt8` value could specify the delay (= buffer size).

jschleicher:
@ipa-mdl I'm not aware of all use-cases, where ros_controllers is used. But sending out desired values to an underlying HW-interface (can be PID or any other control), I don't see any point in jumping from a `desired` position with error to the `actual` position (what is the step response of the underlying controller, if the `desired` position changes abruptly?).

The main point of this PR is, that while a joint is moving, the `actual` position has a delay and is thus behind the first point of the calculated stop trajectory. A jump backwards cannot be the desired behaviour.

Adolfo-rt suggests in a [comment](https://github.com/ros-controls/ros_controllers/issues/144#issuecomment-60383143), to preserve the old behaviour, if `stop_time` is zero. For me that could be an appropriate compromise. Please give me an example, where that is needed.

jschleicher:
I've changed the test case accordingly.
> The delay topic just enables or disables the delay, so this should get reflected in a std_msgs/Bool topic or a std_srvs//SetBoolservice.

jschleicher:
@bmagyar What do you think? Could this be merged?

Should I preserve the old behaviour for stop_time=0?

> Adolfo-rt suggests in a comment, to preserve the old behaviour, if stop_time is zero. For me that could be an appropriate compromise

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -755,12 +762,14 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
   const unsigned int n_joints = joints_.size();
   for (unsigned int i = 0; i < n_joints; ++i)
   {
-    hold_start_state_.position[0]     =  joints_[i].getPosition();
-    hold_start_state_.velocity[0]     =  joints_[i].getVelocity();
+    // If there is a time delay in the system it is better to use calculate the hold trajectory starting from the
```
bmagyar: "better __to use calculate__ the hold trajectory"

bmagyar (changes requested):

bmagyar:
@jschleicher Yes, please preserve that behaviour. 

Also, I know this request is pretty blunt, but can you think of more ways to add tests for your fix? One could be to ensure that old behaviour is preserved.

jschleicher:
Yes, I'll add that.

Thanks for the review.

mathias-luedtke:
~~The tests are blocked..
Please rebase on top of the current `kinetic-devel` branch and resolve the merge conflicts.~~
Thanks ;)

jschleicher:
I've added the requested test case, please review again.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -138,6 +138,13 @@ starting(const ros::Time& time)
   time_data.uptime = ros::Time(0.0);
   time_data_.initRT(time_data);
 
+  // Initialize the desired_state with the current state on startup
```
mathias-luedtke, line 141: This does not change the behavior, it gets overwritten in `update` anyway.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -736,45 +743,64 @@ template <class SegmentImpl, class HardwareInterface>
...
-    hold_start_state_.position[0]     =  joints_[i].getPosition();
-    hold_start_state_.velocity[0]     =  joints_[i].getVelocity();
-    hold_start_state_.acceleration[0] =  0.0;
+    // stop at current actual position
+    for (unsigned int i = 0; i < n_joints; ++i)
+    {
+      hold_start_state_.position[0]     =  joints_[i].getPosition();
+      hold_start_state_.velocity[0]     =  0.0;
+      hold_start_state_.acceleration[0] =  0.0;
+      (*hold_trajectory_ptr_)[i].front().init(time.toSec(),  hold_start_state_,
```
mathias-luedtke: nitpick: `start_time` could be used as well.

mathias-luedtke (commented):
After reading through the dicussions in https://github.com/ros-controls/ros_controllers/issues/144 again, the consensus was that the `stop_trajectory_duration` option is not ideal..

I think adding a parameter/flag might be better to not break existing behavior.

However, the code and the tests look good.

mathias-luedtke:
The desired state is not aligned to the hardware resolution (the current state should be), so commanding the desired state might result in small alternating motions.

Perhaps we should command the current state (joint-wise) if commanding the desired state would change the direction of motion.
Update: This would need a stateful hold with recomputing/adapting the hold trajectory in every cycle step.

mathias-luedtke:
The `time` parameter of `setHoldPosition` holds `time_data_.uptime` (time of last update). However, the `desired_state` will be reached at `time_data_.uptime`+ `time_data_.period`.

So with regard to the non-RT timing, it might make sense to set a 2 segment trajectory:
1. `current_state `@ `time_data_.uptime`
2. `desired_state `@ `time_data_.uptime + time_data_.period`
3. `stop_state `@ `time_data_.uptime + time_data_.period + stop_trajectory_duration`

Not sure if this really makes a difference..
However, this ensures that there is a valid sement for the next update cycle.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
jschleicher: exactly: It gets overwritten on the next update(). But `starting` calls `setHoldPosition` three lines later and reads the desired_state (if `stop_trajectory_duration_ != 0`), so we have to initialize it beforehand.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
jschleicher: yep.

mathias-luedtke:
As discussed with @jschleicher directly, this PR fixes a bug and the new behavior should be the default. If legacy users rely on the old behavior, it can be enabled by setting `stop_trajectory_duration` to zero (which is the default anyway). @bmagyar, do you agree?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
mathias-luedtke: Yes, I have just written this for clarification/justfication during the review :)

jschleicher:
> The time parameter of setHoldPosition holds time_data_.uptime (time of last update). However, the desired_state will be reached

From the plots and the `update` function I'd say `desired_state` is written at `time_data.uptime`:
https://github.com/ros-controls/ros_controllers/blob/e5a6c5a5f9f84756b6601961a5b03b7dffc5118c/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L469

So the hold trajectory is (using this PR) calculated starting from `desired_state@time_data_.uptime`. The first published value on the stop trajectory is calculated on the next update-call `@time_data_.uptime + time_data_.period` automatically.

joint_trajectory_controller/test/joint_trajectory_controller.test
```diff
@@ -35,5 +35,5 @@
   <test test-name="joint_trajectory_controller_test"
         pkg="joint_trajectory_controller"
         type="joint_trajectory_controller_test"
-        time-limit="85.0"/>
+        time-limit="95.0"/>
```
bmagyar: Is this necessary?

joint_trajectory_controller/test/joint_trajectory_controller_vel.test
```diff
@@ -34,6 +34,6 @@
   <!-- Controller test -->
   <test test-name="joint_trajectory_controller_vel_test"
         pkg="joint_trajectory_controller"
-        type="joint_trajectory_controller_test"
-        time-limit="85.0"/>
+        type="joint_trajectory_controller_vel_test"
+        time-limit="120.0"/>
```
bmagyar, line 38: Is this necessary?

bmagyar:
Guys, I think you are very close (or there already) to the source of #144 . I looked at it now, I think the approach is good and I've changed my stance on the question of the flag: we should put this by default.

BTW I experience this jerk every time I switch from the gravity compensation controller on `Tiago` to the `joint_trajectory_contorller` (the arm doesn't have time to fall, only a wee change on encoders).

As this is more of a major change, allow me to come back to this again later this week.

joint_trajectory_controller/test/joint_trajectory_controller_vel.test
jschleicher: This is "rebase" on develop, picked the change from there to avoid merge conflicts.

joint_trajectory_controller/test/joint_trajectory_controller.test
jschleicher: Yes, the additional test takes some more seconds. I think, some "sleep" in the old tests aren't really needed, I'll try to optimize a bit to stay at 85 seconds here.

bmagyar (approved):

bmagyar:
@ipa-mdl you may merge when ready.

mathias-luedtke:
@jschleicher, regarding the timing: 
The update function will compute the spline interpolation from the trajectory's starting state (at last uptime) to the final state (at last uptime + stop trajectory duration). So if the stop trajectoy duration is smaller than the cycle time, the spline interpolation will be skipped.
This does not matter for the postion interface, but it might be used for for the velocity and effort interface.

I guess for most users this should not be an issue because the stop trajectory duration should be big enough anyway for practical reasons.

mathias-luedtke (approved):
Please let me know if you want change anything w.r.t. the timing.
However, I am ready to merge this as-is.

Are you okay with squash-merging?
If not, please rebase accordingly, especially the last commit.

jschleicher:
Yes, please squash-merge.
Thank you.

mathias-luedtke:
Finally :tada: 
Thanks for your patience and the many iterations!

bmagyar:
Thanks guys for the effort!

PR #298, opened 2017-09-30
Update joint_trajectory_controller_impl.h for issue #48

onionsflying:
fix the issue #48

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -166,6 +166,32 @@ trajectoryCommandCB(const JointTrajectoryConstPtr& msg)
 
 template <class SegmentImpl, class HardwareInterface>
 inline void JointTrajectoryController<SegmentImpl, HardwareInterface>::
+actionResultCB(const ActionResultConstPtr& msg)
+{
+  if(msg->status.status == actionlib_msgs::GoalStatus::ABORTED)
+  {
+    // Controller uptime
+    const ros::Time uptime = time_data_.readFromRT()->uptime;
+
+    // Enter hold current position mode
+    setHoldPosition(uptime);
+
+    if(msg->result.error_code == control_msgs::FollowJointTrajectoryResult::PATH_TOLERANCE_VIOLATED)
+    {
+      ROS_WARN_NAMED(name_, "Stopping current movement because path tolerances are violated.");
+      return;
+    }
+    if(msg->result.error_code == control_msgs::FollowJointTrajectoryResult::GOAL_TOLERANCE_VIOLATED)
+    {
+      ROS_WARN_NAMED(name_, "Stopping current movement because goal tolerances are violated.");
+      return;
+    }
+    ROS_WARN_NAMED(name_, "Stopping current movement because the action is aborted");
+  }
+}
+
+template <class SegmentImpl, class HardwareInterface>
+inline void JointTrajectoryController<SegmentImpl, HardwareInterface>::
 preemptActiveGoal()
 {
   RealtimeGoalHandlePtr current_active_goal(rt_active_goal_);
@@ -192,17 +218,6 @@ checkPathTolerances(const typename Segment::State& state_error,
     control_msgs::FollowJointTrajectoryResult::PATH_TOLERANCE_VIOLATED;
     rt_segment_goal->setAborted(rt_segment_goal->preallocated_result_);
     rt_active_goal_.reset();
-    
-    // Sleep until the goal is aborted
-    ros::Duration action_monitor_period_2x=action_monitor_period_.operator *(2);
-    action_monitor_period_2x.sleep();
-
-    // Controller uptime
-    const ros::Time uptime = time_data_.readFromRT()->uptime;
-    
-    // Enter hold current position mode
-    setHoldPosition(uptime);
-    ROS_WARN_NAMED(name_, "Stopping current movement because path tolerances are violated.");
   }
 }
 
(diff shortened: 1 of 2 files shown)


mathias-luedtke:
This is basically the same patch @ipa-fxm  did in https://github.com/ipa-fxm/ros_controllers/commit/7b5a88a89010fcbda1cc9819871d8cb3332b61e9, but it [does not work properly](https://github.com/ros-controls/ros_controllers/issues/48#issuecomment-81122306).

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -193,11 +193,16 @@ checkPathTolerances(const typename Segment::State& state_error,
     rt_segment_goal->setAborted(rt_segment_goal->preallocated_result_);
     rt_active_goal_.reset();
     
+    // Sleep until the goal is aborted
+    ros::Duration action_monitor_period_2x=action_monitor_period_.operator *(2);
+    action_monitor_period_2x.sleep();
```
mathias-luedtke: sleeping in the RT update cycle is a no-go.

onionsflying: ok, understand. It turns out that the [update](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L374) loop after calling [setHoldPosition](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L671) can destroy the [goal_handle_timer_](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L567). So I make update-thread sleep for awhile until the timer run at least one more time for aborting the action. There should be many solutions without sleeping.

onionsflying:
I think [setHoldPosition](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L671) should be called outside the RT update. It consumes about 0.3 ms, which is a little long for RT.

onionsflying:
Hi, @ipa-mdl ! Would you like to review the new commit? If you have any suggestion, please tell me.

mathias-luedtke:
I am not convinced that this is a proper fix, it might even introduce a race condition with multi-threaded spinners.

> I think setHoldPosition should be called outside the RT update. It consumes about 0.3 ms, which is a little long for RT.

How did you measure this?
And why are 0.3ms too long for RT?

> If you have any suggestion, please tell me.

I think it would be great to have a unit test for 0d3de78ef8b19b54beb4a6b007c7e179d57aa5e0.
Their might be a bug somewehe underneath that leads to problem with lost goals.

onionsflying:
I just added some codes e.g. ROS_DEBUG_NAMED(name, "msg, %f", ros::Time::now().toSec()) into [joint_trajectory_controller_imp.h](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h) and [realtime_server_goal_handle.h](https://github.com/ros-controls/realtime_tools/blob/indigo-devel/include/realtime_tools/realtime_server_goal_handle.h) . Then I could get the informations that I mentioned. It's obviously not a strict test. I thought a RT cycle should take less than 1ms. And 0.3ms could be a big part, that might make the whole cycle take more than 1ms. So I wanna insist on my thoughts, since I couldn't imagine the race condition. [cancelCB(GoalHandle gh)](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L583) should be a method without race condition. Have you seen some differences between actionResultCB in 0a7f330 and [cancelCB](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L583) from the perspective of race condition?

I agree that there might be a bug somewhere. Thus I'm not able to explain why the timer [goal_handle_timer_](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L567) stops running abnormally when we patched the code in a way like 0d3de78 . (I know when the timer stops, because I added a line _ROS_DEBUG_NAMED("realtime_tools", "update runNonRealtime");_ in [runNonRealtime(const ros::TimerEvent &te)](https://github.com/ros-controls/realtime_tools/blob/indigo-devel/include/realtime_tools/realtime_server_goal_handle.h#L97) .)

mathias-luedtke:
> I couldn't imagine the race condition

Your code stops the motions without checking the goal matching the aborted one.
However, I don't think that adding a subscriber is a proper fix.

> I just added some codes e.g. ROS_DEBUG_NAMED(name, "msg, %f", ros::Time::now().toSec()) into 

This not a very precise measurement..

>  I thought a RT cycle should take less than 1ms. And 0.3ms could be a big part, that might make the whole cycle take more than 1ms.

In our context RT is about determinism and non-blocking behavior.
The actual cycle time is bound to the application.

> I agree that there might be a bug somewhere. Thus I'm not able to explain why the timer goal_handle_timer_ stops running abnormally when we patched the code in a way like 0d3de78

That's why it would be great to have a unit test for this case.
It might even help to do proper time measurements.

bmagyar:
Does this PR still hold relevance? Can I close it? Should we retarget/rework for melodic?

PR #304, opened 2017-11-27
Enable forwarding of the acceleration values from the trajectory

mmoerdijk:
The joint_trajectory_controller implementation does currently not support the forwarding of the acceleration values of the trajectory, although this information is available. This simple fix enables using the acceleration data from the trajectory for use with the PosVelAccJointHandle.


mathias-luedtke (approved):
LGTM

This will add the acceleration info to the state and feedback.

bmagyar:
:+1:

PR #310, opened 2018-01-13
Modifying joint_state_controller to publish under optional namespace parameter.

rkeatin3:
There are some use cases in which publishing joint states under a namespace can be useful. For example, when using a joint_state_publisher to aggregate separate joint state topics, the ability to publish under a namespace prevents multiple publishers to the /joint_states topic.

### DIFF
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -53,8 +53,14 @@ namespace joint_state_controller
       return false;
     }
 
+    std::string pub_namespace;
+    if (controller_nh.getParam("namespace", pub_namespace))
+    {
+        ROS_DEBUG("Publishing joint states under namespace '%s'", pub_namespace.c_str());
+    }
+
     // realtime publisher
-    realtime_pub_.reset(new realtime_tools::RealtimePublisher<sensor_msgs::JointState>(root_nh, "joint_states", 4));
+    realtime_pub_.reset(new realtime_tools::RealtimePublisher<sensor_msgs::JointState>(root_nh, pub_namespace + "/joint_states", 4));
 
     // get joints and allocate message
     for (unsigned i=0; i<num_hw_joints_; i++){


joint_state_controller/src/joint_state_controller.cpp
```diff
@@ -53,8 +53,14 @@ namespace joint_state_controller
...
 
+    std::string pub_namespace;
+    if (controller_nh.getParam("namespace", pub_namespace))
+    {
+        ROS_DEBUG("Publishing joint states under namespace '%s'", pub_namespace.c_str());
+    }
+
     // realtime publisher
-    realtime_pub_.reset(new realtime_tools::RealtimePublisher<sensor_msgs::JointState>(root_nh, "joint_states", 4));
+    realtime_pub_.reset(new realtime_tools::RealtimePublisher<sensor_msgs::JointState>(root_nh, pub_namespace + "/joint_states", 4));
```
mathias-luedtke, line 63: This would publish to `/joint_states` per default, but it should be `joint_states` (best practice: never hard code absolute namespaces!).

mathias-luedtke (changes requested):
Apart from the rename issue, I don't see why you need this patch..
You can achieve the same with a simple remap.

For Care-O-bot we run all drivers in their own launch groups, so all topics are separated.

joint_state_controller/src/joint_state_controller.cpp
mathias-luedtke: To fix this issue, you should introduce a `topic` parameter (default: `joint_states`) and pass this to the `RealtimePublisher` as-is.

rkeatin3: I had tried adding a remap argument to the launch file that launched the controller spawner, but that didn't seem to work (and I assumed that was the result of the remapping not being propagated through the controller spawner). What's the right syntax?

Agreed on your second point. I coded it that way based on my wrong assumption above that namespacing/topic remapping didn't apply.

rkeatin3: I believe the following syntax _should_ do the remapping that I want, but this hasn't had the desired effect:

```xml
<remap to="/gazebo/joint_states" from="joint_states"/>
<rosparam file="$(find myrobot_gazebo)/controller/joint_state_controller.yaml" command="load"/>
<node name="myrobot_joint_state_controller_spawner" pkg="controller_manager" type="controller_manager" args="spawn joint_state_controller" respawn="false" output="screen"/>
```
I've also tried reversing the to and from arguments in case I was misinterpreting the docs.

mathias-luedtke: The `spawner` is just a "remote control", you have to apply the remap to the node running the `controller_manager`, so `gazebo` in your case.
Instead of remapping, you could as well set the [`robotNamespace` for the Gazebo plugin](https://github.com/ros-simulation/gazebo_ros_demos/blob/kinetic-devel/rrbot_description/urdf/rrbot.gazebo#L7).

rkeatin3: Ah, that certainly makes sense now that you say it. 

Well I obviously should've gone to ROS answers first, but thanks for your help! I'll close this PR.

bmagyar:
Should we put this on the wiki as a tutorial?

rkeatin3:
That sounds like a good idea. The wiki for the joint_state_controller on the [ROS website](http://wiki.ros.org/joint_state_controller?distro=indigo) is totally bare.

I think that the wiki should be updated with the following:
-  Some sort of overview/basic description. There's not much that needs to go here, but it's probably good to emphasize that "controller" is a bit of a misnomer in this case because it just publishes joint values. 
- Information on topic remapping/namespacing. 
- I think a tutorial on integration with MoveIt and Gazebo would be really helpful. I was definitely confused about how each piece fits together in that use case when I first started using this ecosystem. 

Those last two pieces may better fit on the [higher-level ros_control wiki](http://wiki.ros.org/ros_control) though.

PR #314, opened 2018-03-12
[DiffDrive] Per wheel multiplier

artivis:
This PR allows to parametrize one calibration multiplier per wheel (left / right).  
The default behavior remains unchanged (looking for `wheel_radius_multiplier` if exists).  
The `diff_drive_multipliers.test` is duplicated to `diff_drive_left_right_multipliers.test` in order to test this.

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/odometry.h
+++ b/diff_drive_controller/include/diff_drive_controller/odometry.h
@@ -141,7 +141,7 @@ namespace diff_drive_controller
 
     /**
      * \brief Sets the wheel parameters: radius and separation
-     * \param wheel_separation   Seperation between left and right wheels [m]
+     * \param wheel_separation   Separation between left and right wheels [m]
      * \param left_wheel_radius  Left wheel radius [m]
      * \param right_wheel_radius Right wheel radius [m]
      */
--- a/diff_drive_controller/include/diff_drive_controller/odometry.h
+++ b/diff_drive_controller/include/diff_drive_controller/odometry.h
@@ -141,9 +141,9 @@ namespace diff_drive_controller
 
     /**
      * \brief Sets the wheel parameters: radius and separation
-     * \param wheel_separation Seperation between left and right wheels [m]
-     * \param wheel_radius     Left wheel radius [m]
-     * \param wheel_radius     Right wheel radius [m]
+     * \param wheel_separation   Seperation between left and right wheels [m]
+     * \param left_wheel_radius  Left wheel radius [m]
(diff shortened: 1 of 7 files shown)


artivis:
@bmagyar @efernandez Hola.

bmagyar:
LGTM on first read :+1:

bmagyar (approved):

hilario:
LGTM

bmagyar:
@efernandez I'd really like to hear your opinion on this one

efernandez (approved):
LGTM, my only concern is that a similar thing is in https://github.com/clearpathrobotics/ros_controllers/commit/9683ef19ec46e713ba42764355a0da2f2399300b

Could you please take that commit instead of yours? @artivis 

It's almost the same. Probably the only different is that the argument names are correct in the doxygen, in mine :)

diff_drive_controller/include/diff_drive_controller/odometry.h
```diff
@@ -142,9 +142,10 @@ namespace diff_drive_controller
     /**
      * \brief Sets the wheel parameters: radius and separation
      * \param wheel_separation Seperation between left and right wheels [m]
-     * \param wheel_radius     Wheel radius [m]
+     * \param wheel_radius     Left wheel radius [m]
```
efernandez: This is correct (updated) in my commit here: https://github.com/clearpathrobotics/ros_controllers/commit/9683ef19ec46e713ba42764355a0da2f2399300b

efernandez: It should be easy to cherry pick it.

artivis:
Sorry @efernandez but I either mess-up with CPR's fork history or PAL's ...

diff_drive_controller/include/diff_drive_controller/odometry.h
efernandez: @artivis If you do this change you'll have exactly my commit in CPR. I could cherry pick it myself if you want and create another PR :)

artivis: The change we are talking about is the doc diff you linked above ?

bmagyar: AFAIK yes

artivis:
@bmagyar @efernandez See aa51980 .  
@efernandez Should I also fix the `Seperation` typo ?? :P

bmagyar:
Please fix the typo too so we can end this PR :)

bmagyar:
Thanks for sticking with this PR guys :)

Issue #292, opened 2017-08-14
Tests execution is becoming somewhat erratic


bmagyar:
Our testing suite is becoming unstable...

Some tests fail on travis:
https://travis-ci.org/ros-controls/ros_controllers/builds

While on the ROS jenkins even basic things started failing:
http://build.ros.org/job/Kdev__ros_controllers__ubuntu_xenial_amd64/30/

Could someone have a look at this please? @ipa-mdl maybe? or @efernandez ?

efernandez:
Did something got merged recently? I'm not aware of any recent change.

bmagyar:
In the last pr I merged I added a test for the allow multiple publulishers
param.

mathias-luedtke:
Digging into the logs and the multiple publisher PR, it seems like the odometry contains NaNs or an invalid quaternion.

The test is calculating the velocity itself, it would be more robust to just read it from the twist message.
However, the NaN issue should get fixed.

mathias-luedtke:
Possible reason:
The NaN issue is [by design](https://github.com/ros-controls/ros_controllers/blob/kinetic-devel/diff_drive_controller/test/diffbot.h#L101).
The tests fail whenever the test went so fast that one of the first odomtry messages is read, which contains NaN.

I think one publisher should get moved into the test and the velocity should get read from the twist.

PR #316, opened 2018-03-13
[WIP - DiffDrive] Publish joint trajectory controller state 

artivis:
Allows to publish the 'wheel data' as reported by the `hardware_interface::JointHandle` functions `getVelocity() / getPosition()`.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -163,6 +163,7 @@ namespace diff_drive_controller{
     , enable_odom_tf_(true)
     , wheel_joints_size_(0)
     , publish_cmd_(false)
+    , publish_joint_trajectory_controller_state_(false)
   {
   }
 
@@ -263,6 +264,9 @@ namespace diff_drive_controller{
     // Publish limited velocity:
     controller_nh.param("publish_cmd", publish_cmd_, publish_cmd_);
 
+    // Publish wheel data:
+    controller_nh.param("publish_joint_trajectory_controller_state", publish_joint_trajectory_controller_state_, publish_joint_trajectory_controller_state_);
+
     // If either parameter is not available, we need to look up the value in the URDF
     bool lookup_wheel_separation = !controller_nh.getParam("wheel_separation", wheel_separation_);
     bool lookup_wheel_radius = !controller_nh.getParam("wheel_radius", wheel_radius_);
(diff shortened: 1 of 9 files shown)


artivis:
@bmagyar @efernandez Hola.
Notice that those commit are cherry-picked therefore @efernandez appears as the author, let me know if this is any trouble.

artivis:
Seems that CI fails due to a unrelated test:  
```terminal
[joint_trajectory_controller.rosunit-joint_trajectory_controller_vel_test/goalToleranceViolation][FAILURE]
/root/catkin_ws/src/ros_controllers/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp:1081
Value of: waitForState(action_client, SimpleClientGoalState::SUCCEEDED, long_timeout)
  Actual: false
Expected: true
```
Is it related to #292 ??

bmagyar:
Yes, possibly related. I've restarted the travis jobs for now.

bmagyar:
How positive are you about adding a message to the package? 
Shouldn't we try to put this into `control_msgs` instead?

artivis:
Shouldn't we try to put this into control_msgs instead?

artivis:
> Shouldn't we try to put this into control_msgs instead?

Do you mean moving the message as is to `control_msgs` ?

bmagyar:
Before simply moving it, we could think about what's the most general way to define a wheel data message

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -292,6 +296,32 @@ namespace diff_drive_controller{
       cmd_vel_pub_.reset(new realtime_tools::RealtimePublisher<geometry_msgs::TwistStamped>(controller_nh, "cmd_vel_out", 100));
     }
 
+    // Wheel data:
+    if (publish_wheel_data_)
+    {
+      wheel_data_pub_.reset(new realtime_tools::RealtimePublisher<WheelDataStamped>(controller_nh, "wheel_data", 100));
+
+      wheel_data_pub_->msg_.header.frame_id = "0";
```
bmagyar: What is the point of setting this?

bmagyar (commented):
We should agree on the message, the rest is ok, I have only one general note.

I find your lack of tests disturbing.
alttt

diff_drive_controller/src/diff_drive_controller.cpp
artivis: none, that's a miss from melding... About to remove

artivis: Lack of tests ?
You are hurting [diff_drive_publish_wheel_data_test.cpp](https://github.com/ros-controls/ros_controllers/pull/316/files#diff-1a8513c9c34436991145f42a03d65da8) feelings :D 
You want moar ?

artivis:
About the message, I'll replace the custom message of this PR by a [JointState.msg](http://docs.ros.org/api/sensor_msgs/html/msg/JointState.html) :+1:.

artivis:
Force pushed a new version of the feature, no more custom message but instead make use of [sensor_msgs/JointState.msg](http://docs.ros.org/api/sensor_msgs/html/msg/JointState.html).  
Mind the two tests which assert that by default the wheel data is not published and that it is when `publish_wheel_data = true`.

hilario (approved):
LGTM

artivis:
Marking **WIP** while updating to use a [JointTrajectoryControllerState.msg](http://docs.ros.org/jade/api/control_msgs/html/msg/JointTrajectoryControllerState.html) after @bmagyar recommendation.

artivis:
I ain't 100% sure about the 'desired' field of the [JointTrajectoryControllerState.msg](http://docs.ros.org/jade/api/control_msgs/html/msg/JointTrajectoryControllerState.html), if any of you could review it ??

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -411,6 +454,71 @@ namespace diff_drive_controller{
...
+        joint_trajectory_controller_state_pub_->msg_.actual.effort[i]        = left_wheel_joints_[i].getEffort();
+
+        joint_trajectory_controller_state_pub_->msg_.actual.positions[i + wheel_joints_size_]     = right_wheel_joints_[i].getPosition();
+        joint_trajectory_controller_state_pub_->msg_.actual.velocities[i + wheel_joints_size_]    = right_wheel_joints_[i].getVelocity();
+        joint_trajectory_controller_state_pub_->msg_.actual.accelerations[i + wheel_joints_size_] = right_wheel_acc;
+        joint_trajectory_controller_state_pub_->msg_.actual.effort[i+ wheel_joints_size_]         = right_wheel_joints_[i].getEffort();
+
+        // Desired
+        joint_trajectory_controller_state_pub_->msg_.desired.positions[i]    += vel_left_desired * cmd_dt;
+        joint_trajectory_controller_state_pub_->msg_.desired.velocities[i]    = vel_left_desired;
```
bmagyar, line 483: What is your desired velocity? Is it the one set directly by `cmd_vel` or after applying the limiters? 

To me an internal controller state monitor would use the current value to fill `actual` and the one you are sending out in this loop on the joint interfaces as `desired`.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -411,6 +454,71 @@ namespace diff_drive_controller{
       left_wheel_joints_[i].setCommand(vel_left);
       right_wheel_joints_[i].setCommand(vel_right);
     }
+
+    // Publish wheel data:
+    if (publish_joint_trajectory_controller_state_ && joint_trajectory_controller_state_pub_->trylock())
```
bmagyar, line 459: This section became substantial enough to deserve its own function :) 

Also `// Publish wheel data` doesn't really tell us anything new after reading the code, I think it could be removed.

diff_drive_controller/src/diff_drive_controller.cpp
artivis: Here `desired` as been filled with the `cmd_vel` prior to limitations, while `actual` is the one reported by (from) the `JointHandle`.

PR #318, opened 2018-03-14
[DiffDrive] Test fixing

artivis:
Fixing diff_drive_odom_frame_test.

In some cases the test `testOdomTopic` might be calling `getLastOdom()` before any odom message is received. It thus evaluate a default constructed msg.

I also turned the `while` loop that wait for the controller to be up into a function. It also ensure that `ros::ok()` otherwise launching manually a test without the controller get stuck in an infinite loop...

### DIFF
--- a/diff_drive_controller/test/diff_drive_test.cpp
+++ b/diff_drive_controller/test/diff_drive_test.cpp
@@ -34,10 +34,8 @@
 TEST_F(DiffDriveControllerTest, testForward)
 {
   // wait for ROS
-  while(!isControllerAlive())
-  {
-    ros::Duration(0.1).sleep();
-  }
+  waitForController();
+
   // zero everything before test
   geometry_msgs::Twist cmd_vel;
   cmd_vel.linear.x = 0.0;
@@ -54,11 +52,15 @@ TEST_F(DiffDriveControllerTest, testForward)
 
   nav_msgs::Odometry new_odom = getLastOdom();
 
+  const ros::Duration actual_elapsed_time = new_odom.header.stamp - old_odom.header.stamp;
+
+  const double expected_distance = cmd_vel.linear.x * actual_elapsed_time.toSec();
+
   // check if the robot traveled 1 meter in XY plane, changes in z should be ~~0
   const double dx = new_odom.pose.pose.position.x - old_odom.pose.pose.position.x;
   const double dy = new_odom.pose.pose.position.y - old_odom.pose.pose.position.y;
   const double dz = new_odom.pose.pose.position.z - old_odom.pose.pose.position.z;
-  EXPECT_NEAR(sqrt(dx*dx + dy*dy), 1.0, POSITION_TOLERANCE);
+  EXPECT_NEAR(sqrt(dx*dx + dy*dy), expected_distance, POSITION_TOLERANCE);
   EXPECT_LT(fabs(dz), EPS);
 
   // convert to rpy and test that way
@@ -81,10 +83,8 @@ TEST_F(DiffDriveControllerTest, testForward)
 TEST_F(DiffDriveControllerTest, testTurn)
 {
   // wait for ROS
-  while(!isControllerAlive())
-  {
-    ros::Duration(0.1).sleep();
-  }
+  waitForController();
+
   // zero everything before test
   geometry_msgs::Twist cmd_vel;
   cmd_vel.linear.x = 0.0;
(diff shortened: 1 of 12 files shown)


artivis:
@bmagyar @efernandez Hola.
Let me know if I should update all other tests with the newly added function `waitForController`.

artivis:
The failing CI job may again be due to the same issue.

```sh
ros_controllers/diff_drive_controller/test/diff_drive_multiple_cmd_vel_publishers_test.cpp:58
Expected: (fabs(roll_new - roll_old)) < (EPS), actual: nan vs 0.01
```

bmagyar:
I've restarted the jobs again. Please try adding the same fix to all the tests that seem affected. We shall see.

bmagyar:
This should fix #292

artivis:
This does not fully solves #292 but seems to help - over more than a dozen run this particular test did not fail again.  
There are other issues, one taking place in tests where a `ros::Duration::sleep` is used to wait for the controller to execute an expected trajectory - e.g. [diff_drive_test - testForward](https://github.com/ros-controls/ros_controllers/blob/kinetic-devel/diff_drive_controller/test/diff_drive_test.cpp#L61). Those tests seems to fail by a very little overshoot. I'll try to do some monitoring tomorrow, my best guess by now is that the thread may wake up a little too late (machine under heavy load, cosmic ray...), thus the controller does 1/2/3... more iterations of its loop than expected. A possible fix would be that the expected trajectory/distance traveled ain't a hard-coded value. E.g. from the link above:  
```cpp
const ros::Duration actual_elapsed = new_odom.header.stamp - old_odom.header.stamp;
const double expected_traveled = cmd_vel.linear.x * actual_elapsed;
...
EXPECT_NEAR(expected_traveled, sqrt(dx*dx + dy*dy), POSITION_TOLERANCE);
```
@bmagyar What's your take about that ??

artivis:
Currently failing the kind of issue I was trying to describe in my previous comment:

```sh
[diff_drive_controller.rosunit-skid_steer_test/testTurn][FAILURE]---------------
/root/catkin_ws/src/ros_controllers/diff_drive_controller/test/diff_drive_test.cpp:124
The difference between fabs(new_odom.twist.twist.angular.z) and 3.14159265358979323846/10.0 is 0.1029772675980104, 
which exceeds EPS, where fabs(new_odom.twist.twist.angular.z) evaluates to 0.41713653295698971,
3.14159265358979323846/10.0 evaluates to 0.31415926535897931,
and EPS evaluates to 0.01.
```

bmagyar:
That's a very good idea and would explain why we only experience this I
stability on Travis. Thanks for looking into this.

diff_drive_controller/test/diff_drive_test.cpp
```diff
@@ -54,11 +52,15 @@ TEST_F(DiffDriveControllerTest, testForward)
...
+  const ros::Duration actual_elapsed_time = new_odom.header.stamp - old_odom.header.stamp;
+
+  const double expected_distance = cmd_vel.linear.x * actual_elapsed_time.toSec();
+
   // check if the robot traveled 1 meter in XY plane, changes in z should be ~~0
   const double dx = new_odom.pose.pose.position.x - old_odom.pose.pose.position.x;
   const double dy = new_odom.pose.pose.position.y - old_odom.pose.pose.position.y;
   const double dz = new_odom.pose.pose.position.z - old_odom.pose.pose.position.z;
-  EXPECT_NEAR(sqrt(dx*dx + dy*dy), 1.0, POSITION_TOLERANCE);
+  EXPECT_NEAR(sqrt(dx*dx + dy*dy), expected_distance, POSITION_TOLERANCE);
```
artivis, line 63: @bmagyar here is the proposed change.

bmagyar: awesome!

artivis: Not really... Now [l.74](https://github.com/ros-controls/ros_controllers/pull/318/files/747ae347a7b06bfa2b8ffce6852c7e392a4102a1..fdcd29338656cb89e4d80c66c75326b492a8aff4#diff-3e80dfcdac2700fef577d7a8033ff8aaR74) fails (sometime)...  
This quite looks like a clock issue. I will try later having the `diffbot` publishing it's own clock and `use_sim_time true`.

diff_drive_controller/test/diffbot.cpp
```diff
@@ -40,19 +41,54 @@ int main(int argc, char **argv)
...
+  // This should be set in launch files
+  // as well
+  nh.setParam("/use_sim_time", true);
+
   Diffbot<> robot;
   ROS_WARN_STREAM("period: " << robot.getPeriod().toSec());
   controller_manager::ControllerManager cm(&robot, nh);
 
-  ros::Rate rate(1.0 / robot.getPeriod().toSec());
+  ros::Publisher clock_publisher = nh.advertise<rosgraph_msgs::Clock>("/clock", 1);
```
artivis, line 52: @bmagyar @efernandez The simulation now publishes its own clock which should fix time/integration erratic failures.

artivis:
`joint_trajectory_controller` is the one failing as of now... :s

bmagyar:
I kept restarting the travis jobs in the last 2 days. We have now around 10 runs when there was no failure from the patched up tests, nice job!

artivis:
:+1:

bmagyar (approved):
The reliability of `diff_drive_controller` tests has dramatically increased on Travis, thanks for these changes!

artivis:
You're welcome ;)

PR #320, opened 2018-03-15
[DiffDrive] dynamic reconfigure

artivis:
Add a dynamic_reconfigure server to diff-drive controller allowing to change some parameters on the fly, e.g. left/right wheel multiplier, enable_tf etc...

### DIFF
--- a/diff_drive_controller/CMakeLists.txt
+++ b/diff_drive_controller/CMakeLists.txt
@@ -103,4 +103,8 @@ if (CATKIN_ENABLE_TESTING)
   add_rostest(test/diff_drive_radius_param_fail.test)
   add_rostest(test/diff_drive_separation_param.test)
   add_rostest(test/diff_drive_bad_urdf.test)
+  add_rostest_gtest(diff_drive_dyn_reconf_test
+    test/diff_drive_dyn_reconf.test
+    test/diff_drive_dyn_reconf_test.cpp)
+  target_link_libraries(diff_drive_dyn_reconf_test ${catkin_LIBRARIES})
 endif()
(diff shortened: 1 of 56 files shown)


artivis:
@bmagyar Hola.
@efernandez you advised me to try a dyn-reconf scheme in #293, here it is.
I have no idea how to test that the multipliers have been changed given the current state of `diffbot`. Any suggestion ??

bmagyar:
Test: you could try to change the wheel multipliers in an extreme way and use some pre-compute your expected end state with correct ones and multipliers that are __very off__.

artivis:
The problem is that the per-wheel cmd are computed from the corrected wheel radii (multiplier * wheel_rad), thus whether you integrate with mutipliers set to 0.5 or 1.5 your integration should end-up roughly the same.

bmagyar:
That statement begs this reply: if we cannot measure the difference of setting different params, what's the point of the whole thing?

artivis:
We cannot measure any diff with the current test because the `diffbot` is too simplistic - [see here](https://github.com/ros-controls/ros_controllers/blob/kinetic-devel/diff_drive_controller/test/diffbot.h#L104).  
If we had an actual simulator (say Gazebo) then a difference would appear.  
I guess the `diffbot` could be updated to take into account the kinematic defined in the robot_model and compute the joint positions from it.

artivis:
Unwip since #314 was merged.

artivis:
`joint_trajectory_controller` tests are failing.

artivis:
Hi @efernandez, with #314 merged I'm looking forward to your review :D

bmagyar:
Please resolve the current conflicts. 

Also ping @efernandez for review

bmagyar (approved):

PR #319, opened 2018-03-15
[4ws] fix tests

artivis:
The pr is in the same spirit has #318, aiming at helping with #292 .

PR #324, opened 2018-03-27
skidsteerbot publishes its own clock

artivis:
The same simulation clock scheme as in #318 & #319 .


artivis:
@bmagyar joint_trajectory_controller failed.

bmagyar (approved):

PR #327, opened 2018-04-02
Use a copy of the pointer in update()

7675t:
See #301 and https://github.com/ros-controls/ros_controllers/pull/325 .
The original code has potential risk to access the object which is already deleted.

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -55,6 +55,7 @@ if(CATKIN_ENABLE_TESTING)
     REQUIRED COMPONENTS
       actionlib
       controller_manager
+      rosgraph_msgs
       xacro
   )
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -468,6 +468,7 @@ update(const ros::Time& time, const ros::Duration& period)
   {
     current_active_goal->preallocated_result_->error_code = control_msgs::FollowJointTrajectoryResult::SUCCESSFUL;
     current_active_goal->setSucceeded(current_active_goal->preallocated_result_);
+    current_active_goal.reset(); // do not publish feedback  
     rt_active_goal_.reset();
     successful_joint_traj_.reset();
   }
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -477,17 +477,17 @@ update(const ros::Time& time, const ros::Duration& period)
                                   desired_state_, state_error_);
 
   // Set action feedback
-  if (rt_active_goal_)
+  if (current_active_goal)
   {
-    rt_active_goal_->preallocated_feedback_->header.stamp          = time_data_.readFromRT()->time;
-    rt_active_goal_->preallocated_feedback_->desired.positions     = desired_state_.position;
-    rt_active_goal_->preallocated_feedback_->desired.velocities    = desired_state_.velocity;
-    rt_active_goal_->preallocated_feedback_->desired.accelerations = desired_state_.acceleration;
-    rt_active_goal_->preallocated_feedback_->actual.positions      = current_state_.position;
-    rt_active_goal_->preallocated_feedback_->actual.velocities     = current_state_.velocity;
-    rt_active_goal_->preallocated_feedback_->error.positions       = state_error_.position;
-    rt_active_goal_->preallocated_feedback_->error.velocities      = state_error_.velocity;
-    rt_active_goal_->setFeedback( rt_active_goal_->preallocated_feedback_ );
+    current_active_goal->preallocated_feedback_->header.stamp          = time_data_.readFromRT()->time;
+    current_active_goal->preallocated_feedback_->desired.positions     = desired_state_.position;
+    current_active_goal->preallocated_feedback_->desired.velocities    = desired_state_.velocity;
+    current_active_goal->preallocated_feedback_->desired.accelerations = desired_state_.acceleration;
+    current_active_goal->preallocated_feedback_->actual.positions      = current_state_.position;
+    current_active_goal->preallocated_feedback_->actual.velocities     = current_state_.velocity;
+    current_active_goal->preallocated_feedback_->error.positions       = state_error_.position;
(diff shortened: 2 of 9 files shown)


mathias-luedtke:
I am not sure why the tests have failed and why they seem to be resolved now.
My commit just enforces the old feedback behavior.
I will rerun the tests a couple of times.

mathias-luedtke:
`check_jta.py` does not lead to crashs anymore.
However, I get lines like these
```
[ERROR] [1522652388.206902, 87.551000]: Received comm state ACTIVE when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory
[ERROR] [1522652404.340402, 103.274000]: Received comm state ACTIVE when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory
[ERROR] [1522652404.341666, 103.276000]: Received comm state PREEMPTING when in simple state DONE with SimpleActionClient in NS /arm_controller/follow_joint_trajectory
[ERROR] [1522652404.343794, 103.279000]: SimpleActionClient received DONE twice
```
These aren't related to this fix. I get them as well (less frequent) if I comment the feedback part out or with #325.

7675t:
[Test failures](https://travis-ci.org/ros-controls/ros_controllers/builds/361065868) were from the time-limit for joint_trajectory_controller tests, so I increased the time-limit. However, [another failure](https://travis-ci.org/ros-controls/ros_controllers/jobs/361209369) occurs. We can see exactly same failure in [the previous build](https://travis-ci.org/ros-controls/ros_controllers/jobs/356249267) (emptyTopicCancelsActionTrajWithDelayStopZero), so I guess it is not from the change in this PR. I have no idea how I can investigate it.

bmagyar:
I'm wondering whether a similar solution for `joint_trajectory_controller` could work as it did on #318 .

7675t:
I'm not sure, but do we need ``ASSERT_TRUE(initState());`` on the top of every test? 
The failed test doesn't have that.

bmagyar:
I don't think we'd need the `initState()` for those tests, please remove them from the PR. They only add more waits which may fix failing on those tests, but the general issue with tests is that they are time-dependent and I think Travis sometimes pauses the runner which leads to inconsistencies in ROS-time as measured by the Wall clock. The `diff_drive_controller` had similar issues but was fixed by publishing time locally + `use_sim_time:=True`.

bmagyar:
```
build/joint_trajectory_controller/test_results/joint_trajectory_controller/rostest-test_joint_trajectory_controller_wrapping.xml: 1 tests, 1 errors, 0 failures, 0 skipped
```
is consistently erroring out, that's kind of good. 

The one below is a different story however:
```
build/joint_trajectory_controller/test_results/joint_trajectory_controller/rosunit-joint_trajectory_controller_stopramp_test.xml: 25 tests, 0 errors, 1 failures, 0 skipped
```

When you run them locally, they all pass, right?

7675t:
Thanks @bmagyar ,
I've copied codes from #318 to publish sim clock in tests. Please check if it is ok.
I'm not quite sure it's good to publish sim clock, because it makes system synchronized and seems different from actual one. But I have no idea, my intention was to avoid the crush problem.

And yes, I test locally every time before push. Please give me some time for investigation.

7675t:
Sorry, I didn't noticed joint_trajectory_wrapping.test get error even locally. Then I noticed I need same change to rrbot_wrapping.cpp :disappointed: I believe now it should be ok, at least it passes tests locally.

mathias-luedtke:
It looks like adding sim time does not help at all.
The tests seem to be adapted to the broken behavior.

I would suggest to enrich the tests that fail with more debug output.
I will try to find some time to analyze this.

bmagyar:
I restarted the builds and I think the previous failure seems to have repeated:

```
      <failure message="/root/catkin_ws/src/ros_controllers/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp:974&#x0A;The difference between state1-&gt;desired.accelerations[i] and state2-&gt;desired.accelerations[i] is 0.034293197364173758, which exceeds EPS, where&#x0A;state1-&gt;desired.accelerations[i] evaluates to -0.034293197364173758,&#x0A;state2-&gt;desired.accelerations[i] evaluates to 0, and&#x0A;EPS evaluates to 0.01." type=""><![CDATA[/root/catkin_ws/src/ros_controllers/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp:974
The difference between state1->desired.accelerations[i] and state2->desired.accelerations[i] is 0.034293197364173758, which exceeds EPS, where
state1->desired.accelerations[i] evaluates to -0.034293197364173758,
state2->desired.accelerations[i] evaluates to 0, and
EPS evaluates to 0.01.]]></failure>
```

Put this one here for reference. Let's see if it happens again.

7675t:
Hmm, the problem is the instability of the test results. Some tests fail when I did `catkin run_tests joint_trajectory_controller` locally. The result is not stable, I mean, sometimes it pass all tests but sometimes (about one fifth chance)  it doesn't. I'm afraid we have to fix these test problems before proceeding, though I have no idea where to investigate now.

One of the failure:
```
[joint_trajectory_controller.rosunit-joint_trajectory_controller_stopramp_test/emptyActionCancelsTopicTraj][FAI
LURE]
/home/tajima/Codes/ws/src/ros_controllers/joint_trajectory_controller/test/joint_trajectory_controller_
test.cpp:974
The difference between state1->desired.accelerations[i] and state2->desired.accelerations[i] is 0.0346395932971
78264, which exceeds EPS, where
state1->desired.accelerations[i] evaluates to -0.034639593297178264,
state2->desired.accelerations[i] evaluates to 0, and
EPS evaluates to 0.01.
--------------------------------------------------------------------------------
```

BTW, I noticed the latest kinetic-devel branch also fails sometimes in joint_trajectory_controller tests locally, it may be an independent issue.

bmagyar:
Have you tried running these changes rebased on top of #336 ?

7675t:
I didn't notice his PR, thank you @bmagyar 
I'm investigating #336 and will try to rebase on it.

dg-shadow:
Hey. Thanks for your work on this :+1:  This pr does fix the original problem we were having... Have you made any progress on getting the tests passing? I'm happy to help if I can :)

mathias-luedtke:
The tests fail for the ` kinetic-devel` as well.
I would suggest to merge #342 even with failing tests, as it fixes a threading issue.

7675t:
Thanks so much, I agree to proceed merging this by rebasing.

But please keep mind https://github.com/ros-controls/ros_controllers/pull/336 shows the test can fail.  I'm afraid it may be a trouble afterwards.

7675t:
Close since #342  merged.

PR #334, opened 2018-04-20
Make the compiler happy in the test.

clalancette:
This should fix the error that is happening on Melodic: http://build.ros.org/view/Mdev/job/Mdev__ros_controllers__ubuntu_bionic_amd64/3/console

The test still fails for some reason, but at least it
compiles now.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


bmagyar:
Ok, your solution is a bit shorter, I had this: https://github.com/bmagyar/ros_controllers/commit/a4d2264af7247ec77d32be0ff5824ce32269de20

I'm waiting with this until I finish the *SharedPtr refactoring I'm doing over at `ros_control`, then after the corresponding changes here I will merge this one. Kinetic doesn't have the C++11 flags enable and don't want to change that this much into the distro. Melodic however is just fine.

PR #335, opened 2018-04-23
Changes to allow inheritance from JointTrajectoryController.

agutenkunst:
I know this maybe a big step switching from private to protected, yet I would appreciate the ability to use the JointTrajectoryController as a base class. This would be convenient in order to slightly modify for example the setHoldPosition(). 

Looking forward to discuss this!

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -146,7 +146,7 @@ public:
   void update(const ros::Time& time, const ros::Duration& period);
   /*\}*/
 
-private:
+protected:
 
   struct TimeData
   {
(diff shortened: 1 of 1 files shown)


bmagyar:
In general I'm pretty happy about it. My only worry at the moment is some overhead due to the virtual table. Wonder if it's measurable at all :) 

I really like this post on this matter:
https://softwareengineering.stackexchange.com/a/191647

agutenkunst:
Yes there will be overhead but I would say if this effects running applications the respective controller was already pretty close to the edge. However in the case I have in mind https://github.com/ros-controls/ros_controllers/pull/338 would solve my problem without virtualizing :-).

Would you feel more accommodated with https://github.com/ros-controls/ros_controllers/pull/338?

agutenkunst:
I removed the virtual of
`void publishState(const ros::Time& time);`
and 
`void setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh=RealtimeGoalHandlePtr());`
since they are declared to be realtime-safe.

And we have to keep in mind that:
```
class ControllerBase
{
public:
virtual void starting(const ros::Time& time) {};
virtual void stopping(const ros::Time& time) {};
```

agutenkunst:
Any thoughts/updates on keeping just the realtime-safe functions non-virtual?

bmagyar:
Sorry if I was confusing. I think that there may be an overhead but then realized that using shared libraries entails a very similar overhead anyways (all controllers are loaded as plugins). 

I'd like to hear the opinion of @ipa-mdl  and @graiola too if possible.

I definitely prefer this to #338

agutenkunst:
Thanks for the quick answer. In hindsight I would also prefer this to #338. Closing #338 for now.

agutenkunst:
@ipa-mdl Your opinion?

bmagyar (approved):

agutenkunst:
Thank you!

PR #338, opened 2018-04-25
Introduced Holding Mode to JointTrajectoryController.

agutenkunst:
I suggest this PR in order to introduce a HOLDING Mode which keeps the JointTrajectoryController running but moves the manipulator to the holding state. Further Trajectories are not accepted in the HOLDING Mode to avoid unwanted robot movement.

The changes don't affect in-field applications but would allow derived controllers to enter this mode e.g. via external service calls.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -146,6 +146,22 @@ public:
   void update(const ros::Time& time, const ros::Duration& period);
   /*\}*/
 
+  /** \brief Operation modes of the controller.
+   *  The controller can either operate in the DEFAULT mode, accepting goals etc.
+   *  or it can be switch into a HOLDING mode where it is either moving to or staying
+   *  in a hold position. This can be used to bring the robot to a controlled stop while
+   *  avoiding mechanical stress before unloading the controller. While beeing in HOLDING mode
+   *  no further trajectories are executed.
+   */
+  enum JointTrajectoryControllerMode {DEFAULT, HOLDING} mode_;
+
+protected:
+  /**
+   * \brief Switch between modes
+   * \note This method is currently designed for the usage of derived classes.
+   */
+  void switchMode(const JointTrajectoryControllerMode& mode);
+
 private:
 
   struct TimeData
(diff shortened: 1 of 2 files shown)


bmagyar:
Could you please add a reference implementation in a form of a couple of tests? 
It'd be ideal to have something that can be ran through some visualization as well. 

About the name of `switchmode`... We have a similar concept at the joint interface level, it may be confusing to call them the same way.

agutenkunst:
So in general you would give it green light?
I will provide some tests soon. By visualization you mean trajectories? 

About the naming: My first idea was to call it a  `state` however this would interfer with the controller_interface::ControllerBase which itself has a state. Any preferences on state/mode?

I would still go with `mode` or maybe `controller_mode` for distinction.

http://wiki.ros.org/joint_trajectory_controller 2.3
Also mentions:

> Sending an empty trajectory message from the topic interface (not the action interface) will stop the execution of all queued trajectories and enter position **hold mode**. The stop_trajectory_duration parameter controls the duration of the stop motion.

miguelprada:
Isn't this something that could be dealt with with a combination of [sending empty trajectories](https://github.com/ros-controls/ros_controllers/blob/89aaf6c8bbd72a5299169e1c959a61d6693a5801/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp#L736) and having some higher level logic to avoid sending new commands?

I have a feeling that this kind of statefulness belongs in an application orchestration/coordination layer, rather than at the controller level.

agutenkunst:
Higher levels could send empty trajectories but you could not be sure that somebody else is sending another trajectory after or while you moveing into holding mode, leave aside staying there.

miguelprada:
> Higher levels could send empty trajectories but you could not be sure that somebody else is sending another trajectory after or while you moveing into holding mode, leave aside staying there.

Don't mean to sound overly critical, but neither can you be sure that somebody else isn't switching back to `DEFAULT` mode and sending additional trajectories. Sure, it is an extra layer of security. I only wanted to state that I'm not totally on-board with the idea of the control layer being responsible for restricting access to the hardware.

agutenkunst:
Using Authentication to allow switching back to `DEFAULT` mode would clearly be a overkill ;-)

I am also not really totally happy about the suggested solution. 
But if look at it from a responsibility perspective I would argue that 
the task to fully complete the hold-trajectory is the responsibility of the JointTrajectoryController and for that he needs the access to the hardware at that point.

I have trouble finding a solution that would not break the API but would still enable the desired behaviour. Any suggestions?

agutenkunst:
Closing for now since https://github.com/ros-controls/ros_controllers/pull/335 would allow to implement the desired behaviour in a derived class.

PR #340, opened 2018-04-26
fix warning about hiding overloaded method

gerkey:
This PR fixes a warning of this form:
~~~
/opt/ros/kinetic/include/controller_interface/controller.h: At global scope:
/opt/ros/kinetic/include/controller_interface/controller.h:71:16: warning: ‘bool controller_interface::Controller<T>::init(T*, ros::NodeHandle&) [with T = hardware_interface::ForceTorqueSensorInterface]’ was hidden [-Woverloaded-virtual]
   virtual bool init(T* /*hw*/, ros::NodeHandle& /*controller_nh*/) {return true;};
                ^
In file included from src/ros_controllers/force_torque_sensor_controller/src/force_torque_sensor_controller.cpp:32:0:
src/ros_controllers/force_torque_sensor_controller/include/force_torque_sensor_controller/force_torque_sensor_controller.h:50:16: warning:   by ‘virtual bool force_torque_sensor_controller::ForceTorqueSensorController::init(hardware_interface::ForceTorqueSensorInterface*, ros::NodeHandle&, ros::NodeHandle&)’ [-Woverloaded-virtual]
   virtual bool init(hardware_interface::ForceTorqueSensorInterface* hw, ros::NodeHandle &root_nh, ros::NodeHandle& controller_nh);
~~~

### DIFF
--- a/four_wheel_steering_controller/src/four_wheel_steering_controller.cpp
+++ b/four_wheel_steering_controller/src/four_wheel_steering_controller.cpp
@@ -168,10 +168,12 @@ namespace four_wheel_steering_controller{
 
     urdf_geometry_parser::UrdfGeometryParser uvk(root_nh, base_frame_id_);
     if(lookup_track)
+    {
       if(!uvk.getDistanceBetweenJoints(front_wheel_names[0], front_wheel_names[1], track_))
         return false;
       else
         controller_nh.setParam("track",track_);
+    }
 
     if(!uvk.getDistanceBetweenJoints(front_steering_names[0], front_wheel_names[0], wheel_steering_y_offset_))
       return false;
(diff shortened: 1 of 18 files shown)


gerkey:
I should add that these warnings appear only if you build with `-Woverload-virtual`, which you probably don't have enabled in your own builds. To be honest, I'm not sure of the importance of this type of warning.

mathias-luedtke:
I have now rebased/squashed both commits. Mostly to trigger a new travis build against melodic-devel.

mathias-luedtke (commented):
This allows the not-overloaded `init` functions (`return true`..) to get called.
I would prefer the old behavior of hiding these.

It might be an option to move the `using` directives into the `private:` sections.
Not sure if this is worth the effort.

four_wheel_steering_controller/src/four_wheel_steering_controller.cpp
```diff
@@ -168,27 +168,33 @@ namespace four_wheel_steering_controller{
 
     urdf_geometry_parser::UrdfGeometryParser uvk(root_nh, base_frame_id_);
     if(lookup_track)
+    {
```
mathias-luedtke, line 171: :+1:

bmagyar (approved):
Thanks for bringing this up to speed! Feel free to merge when you are ready

mathias-luedtke:
@bmagyar: For now I don't see the need to merge it, because the hiding seems to be indended.

bmagyar:
Can we fix the same by using `override` instead of virtual in the derived ones?  I'm not really familiar with this warning

mathias-luedtke:
If I understand this correctly, we would need to pull them in first (like in this PR) and then delete the ones we'd like to hide (because they are not implemented).
In this case we would have to delete the other init function (?).

bmagyar:
I'll push the release without these changes as I'm leaning more towards the `override` solution, also since we are depending quite a lot on inheritance we should probably add this warning to our build.

PR #341, opened 2018-05-03
C++11 migration

bmagyar:


joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
```diff
@@ -29,8 +29,8 @@
 
 #include <algorithm>
 #include <cmath>
+#include <memory>
 
-#include <boost/shared_ptr.hpp>
 #include <boost/thread/mutex.hpp>
```
IanTheEngineer: Any thoughts on replacing the boost mutex's during the c++11 migration as well?

gavanderhoorn:
@bmagyar: what's the biggest blocker right now for this migration to be merged?

joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
bmagyar: I'm all up for it!

bmagyar:
@gavanderhoorn Someone needs to finish it... As you can see I've started working on this quite a while ago but didn't get around to sit down with it ever since.

The biggest blocker was actionlib's set of `XYZPtr` macros which was incomplete, I fixed it, we got it merged into melodic, it caused issues early on which caused it to be reverted. 

https://github.com/ros/actionlib/pull/106

I would personally still prefer to have actionlib fixed but if that can't happen, this PR has to be adjusted for that. Do you guys have anyone who could help out with this effort?

joint_trajectory_controller/test/joint_trajectory_controller_wrapping_test.cpp
bmagyar: I've done it ;)

bmagyar:
I've redone the ugly iterator-based loops, there's almost no more `boost` left, except for accumulators and interfaces with `realtime_tools` shared ptr-s inherited from actionlib typedefs. 

It is ready for a first review, @IanTheEngineer @ipa-mdl care to take a look perhaps?

diff_drive_controller/CMakeLists.txt
```diff
@@ -89,14 +89,14 @@ if (CATKIN_ENABLE_TESTING)
     test/diff_drive_pub_cmd_vel_out.test
```
graiola, line 94: Are we enforcing C++11 in the CMakeLists?

gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h
```diff
@@ -274,8 +274,8 @@ goalCB(GoalHandle gh)
     
   // Setup goal status checking timer
   goal_handle_timer_ = controller_nh_.createTimer(action_monitor_period_,
-						  &RealtimeGoalHandle::runNonRealtime,
-						  rt_goal);
+                                                  &RealtimeGoalHandle::runNonRealtime,
+                                                  rt_goal.get());
```
graiola: Why the difference?

diff_drive_controller/CMakeLists.txt
mathias-luedtke: Please do not enforce C++11.
Instead we could [default to C++14 ](https://github.com/ros2/examples/blob/master/rclcpp/minimal_timer/CMakeLists.txt#L4)

bmagyar: thanks, defaulting to 14!

gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h
bmagyar: good catch! removed

graiola (commented):
Travis is not happy, it looks like the publisher "wheel_joint_controller_state" is not started in one of the tests.

graiola:
LGTM

graiola (approved):

PR #342, opened 2018-05-09
Use a copy of the pointer in update() (rebased)

mathias-luedtke:
rebased version of #327 without any test adaptations.


7675t (approved):

bmagyar (approved):
I agree with @ipa-mdl , let's let this one in and see if the issue persists for anyone.

bmagyar:
#343

PR #347, opened 2018-05-18
Report errors using error_string in FollowJointTrajectory action result

miguelprada:
A couple of colleagues and me just spent a couple of hours troubleshooting a failing trajectory execution that was apparently caused by unsynchronized clocks causing trajectories being received by the controller with all points in the past.

It would have saved us a lot of time if some error message was displayed.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
@@ -106,7 +106,8 @@ struct InitJointTrajectoryOptions
       rt_goal_handle(),
       default_tolerances(0),
       other_time_base(0),
-      allow_partial_joints_goal(false)
+      allow_partial_joints_goal(false),
+      error_string(0)
   {}
 
   Trajectory*                current_trajectory;
@@ -116,6 +117,7 @@ struct InitJointTrajectoryOptions
   SegmentTolerances<Scalar>* default_tolerances;
   ros::Time*                 other_time_base;
   bool                       allow_partial_joints_goal;
+  std::string*               error_string;
 };
 
 template <class Trajectory>
@@ -166,6 +168,9 @@ bool isNotEmpty(typename Trajectory::value_type trajPerJoint)
  * The typical usecase for this variable is when the \p current_trajectory option is specified, and contains data in
  * a different time base (eg. monotonically increasing) than \p msg (eg. system-clock synchronized).
  *
+ * - \b error_string Error message. If specified, an error message will be written to this string in case of failure to
+ * initialize the output trajectory from \p msg.
+ *
  * \return Trajectory container.
  *
  * \tparam Trajectory Trajectory type. Should be a \e sequence container \e sorted by segment start time.
@@ -205,17 +210,25 @@ Trajectory initJointTrajectory(const trajectory_msgs::JointTrajectory&       msg
   ROS_DEBUG_STREAM("Figuring out new trajectory starting at time "
                    << std::fixed << std::setprecision(3) << msg_start_time.toSec());
 
+  std::string error_string;
+
   // Empty trajectory
   if (msg.points.empty())
   {
-    ROS_DEBUG("Trajectory message contains empty trajectory. Nothing to convert.");
+    error_string = "Trajectory message contains empty trajectory. Nothing to convert.";
+    ROS_DEBUG_STREAM(error_string);
+    if (options.error_string)
+      *options.error_string = error_string;
     return Trajectory();
   }
 
   // Non strictly-monotonic waypoints
   if (!isTimeStrictlyIncreasing(msg))
   {
-    ROS_ERROR("Trajectory message contains waypoints that are not strictly increasing in time.");
+    error_string = "Trajectory message contains waypoints that are not strictly increasing in time.";
+    ROS_ERROR_STREAM(error_string);
+    if (options.error_string)
+      *options.error_string = error_string;
     return Trajectory();
   }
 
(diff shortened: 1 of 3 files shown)


mathias-luedtke (dismissed):
Good catch!

bmagyar:
Should we try to capture these in the action interface too?
There's plenty of space left to extend the list error codes [here](http://docs.ros.org/lunar/api/control_msgs/html/action/FollowJointTrajectory.html).

mathias-luedtke:
> Should we try to capture these in the action interface too?

:+1: , but there is no need to add an error code. From the msg definition:
> INVALID_GOAL: The reason for the invalid goal (e.g., the requested
    trajectory is in the past).

~~`updateTrajectoryCommand` could be refactored to return the error code..~~
But we need to fill out the `error_string`. We could pass a string reference to `updateTrajectoryCommand`.

miguelprada:
Actually, the comment in the source code mislead me a little bit, but all points being in the past is not the only reason `initJointTrajectory` can return an empty trajectory. Therefore, the refactoring should affect this function as well.

~My suggestion would be to change the warning I just added to a more generic message about initializing the trajectory, and to open a new issue/PR to discuss/implement how the error reporting through the action interface could be improved.~

Changed my mind. It's not such a big undertaking. My only concern is how to handle the optional `InitJointTrajectoryOptions` argument to `initJointTrajectory`. Quick inspection suggests that it is only ever used (outside of the tests) by explicitly passing an options argument, but we never know whether removing the default value can cause regression in other packages. Otherwise, I could add an overload but it adds complexity without a clear need. What would you suggest?

miguelprada:
I ended up adding an additional `std::string* error_string` argument to `updateTrajectoryCommand` and also as a field to `InitJointTrajectoryOptions`, so the signature of `initJointTrajectory` remains unchanged.

Let me know what you think.

mathias-luedtke:
> I ended up adding an additional std::string* error_string argument to updateTrajectoryCommand

That's what I proposed in a comment that I apparently forgot to actually send out :-/
In addition I wondered if [this warning](https://github.com/ros-controls/ros_controllers/blob/a9e650737d6d04d8a768c84a48763bdbc5e05800/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h#L311) was not effective.

miguelprada:
Just fixed a [faulty line](https://travis-ci.org/ros-controls/ros_controllers/jobs/381558585#L647).

> In addition I wondered if this warning was not effective.

I was supporting my colleagues on the phone and I guess they were probably more focused on the machine running the application logic. You're right to point out that printing an additional message from the controller would not have helped much in this case, but sending the message back in the action result might improve the odds of future similar errors being caught faster.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -555,7 +561,7 @@ updateTrajectoryCommand(const JointTrajectoryConstPtr& msg, RealtimeGoalHandlePt
     }
     else
     {
-      // All trajectory points are in the past, nothing new to execute. Keep on executing current trajectory
+      ROS_WARN_NAMED(name_, "Failed to initialize trajectory from message, skipping.");
```
mathias-luedtke: After actually reviewing `initJointTrajectory`, this warning seems to be superfluous.

miguelprada: Agreed.

In retrospective, its clear that the issue was not really lacking an error message, but rather that attention was not being paid to the right terminal output. 😓

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -166,8 +166,13 @@ template <class SegmentImpl, class HardwareInterface>
 inline void JointTrajectoryController<SegmentImpl, HardwareInterface>::
 trajectoryCommandCB(const JointTrajectoryConstPtr& msg)
 {
-  const bool update_ok = updateTrajectoryCommand(msg, RealtimeGoalHandlePtr());
+  std::string error_string;
+  const bool update_ok = updateTrajectoryCommand(msg, RealtimeGoalHandlePtr(), &error_string);
   if (update_ok) {preemptActiveGoal();}
+  else
+  {
+    ROS_WARN_STREAM_NAMED(name_, "Failed to update trajectory: " << error_string);
```
mathias-luedtke: redundant as well (?)

miguelprada: Indeed.

miguelprada:
I added changes to report back the potential errors in `updateTrajectoryCommand` using `error_string` as well.

Feel free to squash merge if you approve the changes.

Edit: apologies for all the amends and rebases. I’m working without a local build environment and its biting me hard...

mathias-luedtke:
> I’m working without a local build environment and its biting me hard...

https://github.com/ros-industrial/industrial_ci/pull/279 ;)

mathias-luedtke (approved):
I have resolved merge conflict and fixed a typo.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
```diff
@@ -308,10 +330,13 @@ Trajectory initJointTrajectory(const trajectory_msgs::JointTrajectory&       msg
...
-      ROS_WARN_STREAM("Dropping all " << msg.points.size() <<
-                      " trajectory point(s), as they occur before the current time.\n" <<
-                      "Last point is " << std::fixed << std::setprecision(3) << last_point_dur.toSec() <<
-                      "s in the past.");
+      error_string = "Dropping all " + std::to_string(msg.points.size());
+      error_string += " trajectory point(s), as they occur before the current time.\n";
+      error_string += "Last point is " + std::to_string(last_point_dur.toSec());
+      error_string += "s in the past.";
+      ROS_WARN_STREAM(error_string);
+      if (options.error_string)
```
bmagyar: Could we please use full brackets for these to match the format of the rest of the repo?

```
if (options.error_string)
{
  *options.error_string = error_string;
}
```

bmagyar (changes requested):

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
bmagyar: Other than this it's all green lights from me too

mathias-luedtke:
@bmagyar: Better?

bmagyar (approved):

bmagyar:
Thank you!!

PR #349, opened 2018-05-28
Updated DynamicParams ostream for diff_drive_controller

marting87:
DynamicParams ostream was implying it was printing the radius/separation instead of the multiplier. Added _multiplier_ to make this more clear.  

Changed printing of booleans to _enabled/disabled_ instead of a numeric value to be consistent with the .cpp file.


bmagyar:
Great touch-up, thanks!

bmagyar:
@artivis what do you think?

artivis:
LGT :+1:

bmagyar (approved):

PR #350, opened 2018-05-30
[diff_drive_controller] invert rotation direction

dogoepp:
We have a turtlebot-like robot base with two dynamixels actuating two wheels. Due to the geometry of the robot, we have to invert the rotation speed of one of the joint to achieve the desired result with the `diff_drive_controller`.

This can very easily be done in this controller with two extra parameters, as proposed here.


bmagyar:
I'd advise you to do this in your RobotHW instead. Also make sure the urdf is properly set up.

Any notes @artivis ?

dogoepp:
What is the role of the URDF here ?

I wonder which part should handle this situation of a joint direction being inverted. I took changed it here because at the time it seemed to make more sense but it's not clear for me where this kind of detail should be handled.

dogoepp:
OK, I made the change in our hardware interface.

PR #353, opened 2018-07-11
[WIP] Added a script to test simple arm movements using ros_controllers

hsd-dev:


bmagyar:
This script could be a great addition to the software of the arm you use this to test but unfortunately we don't have space for this here. 
Just a couple of points:
* The script assumes 6 degrees of freedom.
* Argparse should be used for command line Python tools.
* There is no safety limit on velocities, what is someone doesn't have proper safety layers in place?
* `rqt_joint_trajectory_controller` can already be used to some extent to test an arm and that works with everything.

mathias-luedtke:
I have suggested to open a PR, mostly for discussing it. This script is meant for testing the trajectory action.
It could even get evolved into a JTC CLI, with a shared code base together with `rqt_joint_trajectory_controller `.

hsd-dev:
Thanks for your comments @bmagyar. 
Actually `rqt_joint_trajectory_controller` has some nice functionalities which can be used only as a rqt plugin. It would be nice to reuse the code similar to [`controller_manager_msgs/utils.py`](https://github.com/ros-controls/ros_control/blob/melodic-devel/controller_manager_msgs/src/controller_manager_msgs/utils.py). Would it make more sense to make a library and use it in  `rqt_joint_trajectory_controller`?

bmagyar:
It could be possible but I'd rather not have `rqt_joint_trajectory_controller` depend on another package. Is it acceptable to fit a command line "version  of rqt_joint_trajectory_controller" into the same package?

PR #355, opened 2018-08-21
Add action_ns parameter to specify action name

7675t:
This PR add a parameter "action_ns" into joint_trajectory_controller to specify the action name.

The action name of joint_trajectory_controller is fixed to "follow_joint_trajectory". And as far as I tried, it cannot be remapped. Some robots (in my case Nextage Open, but some other robots also) have its own controller and different action name like "follow_joint_trajectory_action".  Usually this is not problem because MoveIt! can specify action name by "action_ns" in controller_list.

However, I want to make [fake_joint_controller](https://github.com/tork-a/fake_joint) to emulate these controller. For this purpose, I need to change the action name of the controller. 

Default of this parameter is "follow_joint_trajectory", so it has no effect for existing users.


mathias-luedtke:
I don't like this kind of parameters, because this is what remapping is for.
Remapping should work for actions as well!

7675t:
OK, please let me check again if I can use reamping for this purpose. Thanks.

7675t:
Yes, you are totally right. I misunderstood the behavior of remap. Now I can remap the action name and no problem for my purpose. Thank you so much!

PR #364, opened 2018-10-15
Switch the read and write bodies. 

wilcobonestroo:
The code was confusing with respect to the ros_control documentation. Now it is in line with that documentation.

### DIFF
--- a/diff_drive_controller/test/diffbot.h
+++ b/diff_drive_controller/test/diffbot.h
@@ -82,18 +82,7 @@ public:
 
   void read()
   {
-    std::ostringstream os;
-    for (unsigned int i = 0; i < NUM_JOINTS - 1; ++i)
-    {
-      os << cmd_[i] << ", ";
-    }
-    os << cmd_[NUM_JOINTS - 1];
-
-    ROS_INFO_STREAM("Commands for joints: " << os.str());
-  }
-
-  void write()
-  {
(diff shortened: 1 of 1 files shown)


mathias-luedtke (changes requested):
The current code is correct (the data gets updated by `write`).
The patch introduces writing in `read`, which is not intended.

I am not sure why the `read` function is printing the previous command..
Moving this into `write` might be an option, but for now I would not change it.

> Now it is in line with that documentation.

Which documentation are you referring to?

wilcobonestroo:
In the code, `pos_` and `vel_` refer to the states of the joint and `cmd_` refers to the command that is send to the joints. In the current code, the `pos_` and `vel_` are filled in the `write` and `cmd_` is read in the `read` function.

mathias-luedtke:
`write` applies the `cmd_` to the new state in `pos_`/`vel_`. It emulates one control step without any delay.
[rrbot](https://github.com/ros-controls/ros_controllers/blob/melodic-devel/joint_trajectory_controller/test/rrbot.cpp) is doing the same.
It is just a test mock-up anyway, I don't see the need to change it.

> cmd_ is read in the read function.

`cmd_` just gets printed, `read` is a NOP.

wilcobonestroo:
Ah. Now I see. I was thinking about a real robot and not about emulation. Thank you for your answer.

PR #368, opened 2018-10-28
Minor change in one of the ROS_INFO_STREAM

JaFeKl:
just a small change in the info stream to actually match the displayed variables


bmagyar:
Thanks!

bmagyar (approved):

PR #370, opened 2018-10-30
Kinetic/error string in follow joint trajectory

hslusarek:
For easier maintenance and compatibility with melodic I suggest to integrate #347 into kinetic.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
@@ -65,7 +65,7 @@ template <class T>
 inline std::vector<unsigned int> mapping(const T& t1, const T& t2)
 {
   typedef unsigned int SizeType;
-  
+
   // t1 must be a subset of t2
   if (t1.size() > t2.size()) {return std::vector<SizeType>();}
 
@@ -118,6 +118,14 @@ struct InitJointTrajectoryOptions
   ros::Time*                 other_time_base;
   bool                       allow_partial_joints_goal;
   std::string*               error_string;
+
+  void setErrorString(const std::string &msg) const
+  {
+    if(error_string)
+    {
+      *error_string = msg;
+    }
+  }
 };
 
 template <class Trajectory>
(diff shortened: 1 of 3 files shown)


mathias-luedtke (commented):
Why did you add the c++11 flags?
In kinetic we still stick to c++03 and boost..

hslusarek:
The commits which I cherry-picked from the melodic-branch use the function `std::to_string()` which only comes with C++11 (see [here](https://en.cppreference.com/w/cpp/string/basic_string/to_string)). To compile the code I had to add the C++11 flag. I also looked up the C++ version used for kinetic kame (see [here](http://www.ros.org/reps/rep-0003.html)) which should be C++11. Did I misunderstood something? If it is still necessary to support C++03, then I have to change a view things before the pull request #347 can be transferred to kinetic.

mathias-luedtke:
I see. So we have rewrite the code to not use `to_string'.
C++11 support is available on all kinetic target platforms, but these flags have to be set in derived packages as well. C++11 is not ABI-compatible with C++03, especially regardings strings, so it would be a major break.

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
```diff
@@ -308,10 +333,12 @@ Trajectory initJointTrajectory(const trajectory_msgs::JointTrajectory&       msg
     if (msg_it == msg.points.end())
     {
       ros::Duration last_point_dur = time - (msg_start_time + (--msg_it)->time_from_start);
-      ROS_WARN_STREAM("Dropping all " << msg.points.size() <<
-                      " trajectory point(s), as they occur before the current time.\n" <<
-                      "Last point is " << std::fixed << std::setprecision(3) << last_point_dur.toSec() <<
-                      "s in the past.");
+      error_string = "Dropping all " + std::to_string(msg.points.size());
```
mathias-luedtke: This should be reeimplemented with `stringstream` or some boost helper.

mathias-luedtke (changes requested):

joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h
agutenkunst: Implemented with stringstream now.

mathias-luedtke: Thanks!

mathias-luedtke (approved):

mathias-luedtke:
This introduces an ABI break, but since the switch to virtual functions was quite recent, I think it is still okay to merge.

PR #379, opened 2018-11-11
remove travis_wait

mathias-luedtke:


bmagyar (approved):
It is passed we are good. Is this related to Travis policy changes? I had to add it when we were almost consistently timing out due to test suite taking too long to run.

mathias-luedtke:
There was no change to Travis, but https://github.com/ros-industrial/industrial_ci/pull/324 should make `travis_wait` superfluous :)

bmagyar:
Thanks! Let's see how this does. Could you please cherry pick to Kinetic as well?

Issue #378, opened 2018-11-11
Investigate (and fix) doubled find_package usage in tests


mathias-luedtke:
`catkin` gets resolved twice in
* joint_trajectory_controller
* diff_drive_controller
* four_wheel_steering_controller
* ackermann_steering_controller

https://github.com/ros/catkin/issues/886

PR #380, opened 2018-11-13
Fix doubled-find_package()

MoriKen254:
Fixed https://github.com/ros-controls/ros_controllers/issues/378

for controllers as follows

- joint_trajectory_controller
- diff_drive_controller
- four_wheel_steering_controller
- ackermann_steering_controller

### DIFF
--- a/ackermann_steering_controller/CMakeLists.txt
+++ b/ackermann_steering_controller/CMakeLists.txt
@@ -3,22 +3,14 @@ project(ackermann_steering_controller)
 
 find_package(catkin REQUIRED COMPONENTS
   controller_interface
-  controller_manager
   diff_drive_controller
-  gazebo_ros
-  geometry_msgs
   hardware_interface
   nav_msgs
   pluginlib
   realtime_tools
   roscpp
-  rostest
-  rosunit
-  std_msgs
-  std_srvs
   tf
   urdf
-  xacro
 )
 
 find_package(Boost REQUIRED COMPONENTS system thread)
@@ -56,6 +48,27 @@ install(FILES ackermann_steering_controller_plugins.xml
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 
 if (CATKIN_ENABLE_TESTING)
+  find_package(catkin REQUIRED COMPONENTS
+    controller_interface
+    controller_manager
+    diff_drive_controller
+    gazebo_ros
+    geometry_msgs
+    hardware_interface
+    nav_msgs
+    pluginlib
+    realtime_tools
+    roscpp
+    rostest
+    rosunit
+    rviz
+    std_msgs
+    std_srvs
+    tf
+    urdf
+    xacro
+  )
+
   add_executable(${PROJECT_NAME}_ackermann_steering_bot test/common/src/ackermann_steering_bot.cpp)
   add_dependencies(tests ${PROJECT_NAME}_ackermann_steering_bot)
 
--- a/ackermann_steering_controller/CMakeLists.txt
+++ b/ackermann_steering_controller/CMakeLists.txt
@@ -3,14 +3,22 @@ project(ackermann_steering_controller)
 
 find_package(catkin REQUIRED COMPONENTS
   controller_interface
+  controller_manager
   diff_drive_controller
+  gazebo_ros
+  geometry_msgs
   hardware_interface
   nav_msgs
   pluginlib
   realtime_tools
   roscpp
+  rostest
+  rosunit
+  std_msgs
+  std_srvs
   tf
   urdf
+  xacro
 )
 
 find_package(Boost REQUIRED COMPONENTS system thread)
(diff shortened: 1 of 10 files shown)


mathias-luedtke (changes requested):
Thanks for the effort, but that's not what I meant.
The test dependencies should not get mixed in.

There are two ways:
1. Resolve test dependencies individually (this gets tricky with more dependencies)
2. Resolve ALL test dependencies in a separate `find_package` call (overwrites the first)

I am not sure which is the best way to go.

MoriKen254:
Thank you for your comment and excuse me for my misunderstanding.

Could https://github.com/ros-controls/ros_controllers/pull/380/commits/2d4bab0cf2b5a64ad4077a56ff8d1e850e4f4bd1 be a solution? I overwrote all dependent pkgs in test section.

Just in my opinion, in terms of maintainability it sounds better to do this even though it may look redundant. Checking dependency in test again and removing pkgs unrelated only to test afterwards sounds like a complicated and tough task...

At least, wouldn't this be better compared to the previous one?

ros_controllers/package.xml
```diff
@@ -23,6 +23,7 @@
   <exec_depend>effort_controllers</exec_depend>
   <exec_depend>force_torque_sensor_controller</exec_depend>
   <exec_depend>forward_command_controller</exec_depend>
+  <exec_depend>four_wheel_steering_controller</exec_depend>
```
mathias-luedtke, line 26: This decision is up to @bmagyar

joint_trajectory_controller/CMakeLists.txt
```diff
@@ -48,17 +51,30 @@ add_library(${PROJECT_NAME} src/joint_trajectory_controller.cpp
...
+      roscpp
+      rostest
+      rosunit
+      rqt_plot
+      trajectory_msgs
+      urdf
       xacro
   )
 
-  find_package(rostest REQUIRED)
```
mathias-luedtke, line 61: I would prefer to resolve `rostest` separately. (in all packages)

ackermann_steering_controller/package.xml
```diff
@@ -32,8 +32,9 @@
   <test_depend>controller_manager</test_depend>
   <test_depend>gazebo_ros</test_depend>
   <test_depend>geometry_msgs</test_depend>
-  <test_depend>rosunit</test_depend>
   <test_depend>rostest</test_depend>
+  <test_depend>rosunit</test_depend>
+  <test_depend>rviz</test_depend>
```
mathias-luedtke, line 37: I think it is better to not depend on rviz. It should not be crucial for the test.

ackermann_steering_controller/CMakeLists.txt
```diff
@@ -49,13 +49,23 @@ install(FILES ackermann_steering_controller_plugins.xml
 
 if (CATKIN_ENABLE_TESTING)
   find_package(catkin REQUIRED COMPONENTS
+    controller_interface
```
mathias-luedtke: only compile-time dependencies should be listed here

mathias-luedtke (changes requested):
Yes, this looks better.
But I am still unsure about which way to go here..
However, it makes sense to check/review the dependencies, even if we decide to go the other way later.

ros_controllers/package.xml
bmagyar: Yes this is intended. The controller is in the repo, let's go all the way.

ackermann_steering_controller/package.xml
bmagyar: And they should already be behind a `show_rviz` or `gui` flag.

joint_trajectory_controller/CMakeLists.txt
MoriKen254: Fixed at https://github.com/ros-controls/ros_controllers/pull/380/commits/e745c484517543b00b25111f5aa9dfe749c236ed

ackermann_steering_controller/package.xml
MoriKen254: removed rviz at https://github.com/ros-controls/ros_controllers/pull/380/commits/e745c484517543b00b25111f5aa9dfe749c236ed

ackermann_steering_controller/CMakeLists.txt
MoriKen254: Tried to fix at https://github.com/ros-controls/ros_controllers/pull/380/commits/e745c484517543b00b25111f5aa9dfe749c236ed looking at test codes inside

MoriKen254:
Thanks for your generous comments. I got your point. 

hope this PR will be a kinda current milestone to help you review dependencies and find something to decide which way to go.

bmagyar:
@ipa-mdl Should we rebase & fix this one?

bmagyar:
I've rebased and resolved the conflict on this one, not sure if all changes are required though. 

@matthew-reynolds care to take over this one?

matthew-reynolds:
Sure thing. I'll base it off what we learned/decided in https://github.com/ros-controls/ros_control/pull/404

bmagyar:
This should all be clean in Noetic

PR #385, opened 2018-11-21
Remove a deprecated warning.

AndyZe:


bmagyar:
I know the note says to remove it by indigo but kinetic I think is too stable at the moment for this.

This change also removes the parameter we'd like to keep :)

PR #386, opened 2018-11-21
Remove deprecated parameter hold_trajectory_duration

AndyZe:


bmagyar (approved):
Thanks for this.
Even though I'm approving now the CI may still fail if we have tests directly aimed at this. If any, those will need to be removed.

bmagyar (changes requested):
This change also removes the parameter we'd like to keep :)

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -229,6 +229,7 @@ bool JointTrajectoryController<SegmentImpl, HardwareInterface>::init(HardwareInt
 
   // Stop trajectory duration
   stop_trajectory_duration_ = 0.0;
+  controller_nh_.getParam("stop_trajectory_duration", stop_trajectory_duration_);
```
mathias-luedtke, line 232: For semantic reasons I'd prefer
```
stop_trajectory_duration_ =  controller_nh_.param("stop_trajectory_duration", 0.0);
```

mathias-luedtke:
Tests pass again, `hold_trajectory_duration` is not used elsewhere (in our repos).
https://github.com/ros-industrial/ur_modern_driver still has the old parameter name, but commented out.

bmagyar (approved):
Other than the point @ipa-mdl raised I'm happy with the PR. Please feel free to merge once done

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
mathias-luedtke: There is a copy&paste bug in the new commit. I rolled back to the original version again.
In addition the other parameters are retrieved the same way, so let's keep it like this for now.

PR #392, opened 2018-12-12
A new compliance controller.

AndyZe:
This adds a velocity-based compliance controller. It adjusts the robot's velocity as force/torque data is received. This works both for joint trajectories and feedforward velocity commands. If no force/torque data is available, it just works like normal.

I realize this is a pretty big PR but it adds some very cool functionality.

Here are some videos of four test cases I looked at --

1) Compliance when jogging with feedforward velocity commands:

https://drive.google.com/open?id=15Q2UxuUGHTMoC76gnOzB3C8KYzzEa9BG

2) Compliance when stationary (neither jogging nor following a trajectory)

https://drive.google.com/open?id=1CslRNmdqJmzVK_RC0dvUhipGhZl1KRBU

3) Compliance when following a trajectory. I'm pushing the robot to the left, notice that it ends up to the left of its goal pose.

https://drive.google.com/open?id=1IHM14qyhDg6BLL3YgMF2iNTHwO9hyY4Y

4) Interrupting a trajectory with a feedforward velocity command

https://drive.google.com/open?id=11mnNL_nJxzJRLr-6-o_u6K5TvHaG7UwP


AndyZe:
Sorry, I screwed this up with multiple squashes. Will redo.

PR #393, opened 2018-12-13
(WIP) New type of compliance controller

AndyZe:
This adds a velocity-based compliance controller. It adjusts the robot's velocity as force/torque data is received. This works both for joint trajectories and feedforward velocity commands. If no force/torque data is available, it just works like normal.

I realize this is a pretty big PR but it adds some very cool functionality.

Here are some videos of four test cases I looked at --

1) Compliance when jogging with feedforward velocity commands:

https://drive.google.com/open?id=15Q2UxuUGHTMoC76gnOzB3C8KYzzEa9BG

2) Compliance when stationary (neither jogging nor following a trajectory)

https://drive.google.com/open?id=1CslRNmdqJmzVK_RC0dvUhipGhZl1KRBU

3) Compliance when following a trajectory. I'm pushing the robot to the left, notice that it ends up to the left of its goal pose.

https://drive.google.com/open?id=1IHM14qyhDg6BLL3YgMF2iNTHwO9hyY4Y

4) Interrupting a trajectory with a feedforward velocity command

https://drive.google.com/open?id=11mnNL_nJxzJRLr-6-o_u6K5TvHaG7UwP


AndyZe:
~~An issue I'm finding is that a JointTrajectoryController is not aware of joint limits~~  Fixed this by including a boolean to indicate joint limits in the compliance msg.

bmagyar:
Hi,

Controllers are well aware of joint limits that are specified in the urdf.
How come your limits are defined after the controller? Is there something
I'm missing?

AndyZe:
I'm launching a UR3 like one normally would and it looks like a ros_controllers/forward_command_controller definitely doesn't do joint limit checking.

miguelprada:
> I'm launching a UR3 like one normally would and it looks like a ros_controllers/forward_command_controller definitely doesn't do joint limit checking.

In this case, it's the UR driver which should be enforcing the limits. I've been thinking of working on a PR for that for a while, but never got around to do it.

compliance_controller/compliance_control_msgs/msg/CompliantVelocities.msg
```diff
@@ -0,0 +1,5 @@
+# This array holds velocities for each joint
+std_msgs/Float64MultiArray compliant_velocities
```
gavanderhoorn, line 2: I'm just a user here, so this is me being curious: what is the reason you chose to use a `Float64MultiArray` here? Would something like this not work:

```yaml
float64[] velocity
```
perhaps even with a `string[] name` to be able to match velocities with joints actually controlled by the controller?

`MultiArray` allows for n-dim arrays though: is that (part of) the rationale?

gavanderhoorn:
> > I'm launching a UR3 like one normally would and it looks like a ros_controllers/forward_command_controller definitely doesn't do joint limit checking.
> 
> In this case, it's the UR driver which should be enforcing the limits. I've been thinking of working on a PR for that for a while, but never got around to do it.

Tracking this: https://github.com/ros-industrial/ur_modern_driver/issues/249.

compliance_controller/compliance_control_msgs/msg/CompliantVelocities.msg
AndyZe: The only reason was to stay compatible with a JointGroupVelocity controller... but yeah, I think it should contain joint names, too.

davetcoleman:
Note we do have a [joint_limit_interface](https://github.com/ros-controls/ros_control/tree/melodic-devel/joint_limits_interface) in ros_control, but it doesn't belong in the controller but rather the hardware interface.

compliance_controller/compliance_control_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(compliance_control_msgs)
```
davetcoleman, line 2: This also belongs in https://github.com/ros-controls/control_msgs

compliance_controller/compliance_controller/include/compliance_controller/compliance_controller.h
```diff
@@ -0,0 +1,127 @@
...
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef COMPLIANCE_CONTROLLER_H
+#define COMPLIANCE_CONTROLLER_H
```
davetcoleman: include package name before file name

compliance_controller/compliance_controller/include/compliance_controller/compliance_controller.h
```diff
@@ -0,0 +1,127 @@
...
+  bool init(HardwareInterface* hw, ros::NodeHandle& root_nh, ros::NodeHandle& controller_nh);
+
+protected:
+  typedef traj_or_jog_controller::TrajOrJogController<SegmentImpl, HardwareInterface>  TrajOrJogController;
+  typedef joint_trajectory_controller::JointTrajectoryController<SegmentImpl, HardwareInterface>  JointTrajectoryController;
+
+private:
+  /** \brief Override updates of the base class. */
+  virtual void update(const ros::Time& time, const ros::Duration& period);
+  /*\}*/
```
davetcoleman, line 77: ^ what is that?

compliance_controller/compliance_controller/src/compliance_controller.cpp
```diff
@@ -0,0 +1,315 @@
...
+      {
+        // Check tolerances
+        if (time_data.uptime.toSec() < segment_it->endTime())
+        {
+          // Currently executing a segment: check path tolerances
+          const joint_trajectory_controller::SegmentTolerancesPerJoint< Scalar >& joint_tolerances = segment_it->getTolerances();
+
+          if (!checkStateTolerancePerJoint(JointTrajectoryController::state_joint_error_, joint_tolerances.state_tolerance))
+          {
+            if (JointTrajectoryController::verbose_)
```
davetcoleman: at this point your 5 indentations in, this function should be broken up into a helper function for, say, the inside of the for loop

compliance_controller/compliance_controller/src/compliance_controller.cpp
```diff
@@ -0,0 +1,315 @@
...
+              control_msgs::FollowJointTrajectoryResult::PATH_TOLERANCE_VIOLATED;
+            rt_segment_goal->setAborted(rt_segment_goal->preallocated_result_);
+            JointTrajectoryController::rt_active_goal_.reset();
+            JointTrajectoryController::successful_joint_traj_.reset();
+          }
+        }
+        else if (segment_it == --curr_traj[i].end())
+        {
+          if (JointTrajectoryController::verbose_)
+            ROS_DEBUG_STREAM_THROTTLE_NAMED(1,JointTrajectoryController::name_,"Finished executing last segment, checking goal tolerances");
```
davetcoleman: it would be nice to run clang-format on this PR, for example missing spaces between arguments

compliance_controller/wrench_to_twist_pub/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(wrench_to_twist_pub)
```
davetcoleman: why is this a separate package? its just a publisher?

compliance_controller/wrench_to_twist_pub/include/wrench_to_twist_pub/compliant_control.h
```diff
@@ -0,0 +1,183 @@
...
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef COMPLIANT_CONTROL_H
+#define COMPLIANT_CONTROL_H
```
davetcoleman: add package name.

compliance_controller/wrench_to_twist_pub/src/wrench_to_twist_pub/test_compliance_publisher.cpp
```diff
@@ -0,0 +1,73 @@
...
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+// Send a small velocity command, for testing.
+// Can be used in place of a real force/torque sensor.
```
davetcoleman: this isn't a unit test, so is there a better word for this file such as "demo"?

compliance_controller/wrench_to_twist_pub/src/wrench_to_twist_pub/wrench_to_twist_pub.cpp
```diff
@@ -0,0 +1,226 @@
...
+  ros::init(argc, argv, NODE_NAME);
+
+  // Do compliance calculations in this class
+  wrench_to_twist_pub::PublishCompliantJointVelocities publish_compliance_velocities;
+
+  // Spin and publish compliance velocities, unless disabled by a service call
+  publish_compliance_velocities.spin();
+
+  return 0;
+}
```
davetcoleman: move main to bottom of file

davetcoleman (commented):
Please add a README with the contents of the PR description, perhaps with more explanations and how-tos.

bmagyar:
I'd suggest not including messages here unless it is an absolute must. Are
you willing to move them to control_msgs?

AndyZe:
@bmagyar Yep, I can do that. No rush for me right now, I want to get the MoveIt! jog_arm PR approved and add some more features here.

Here's a [nicer video](https://drive.google.com/open?id=18T9ypI12NCi56p4P81zGUmu3HG5y5kHd) with less-stiff admittance settings.

sjhansen3:
This looks like an Awesome PR. Is there an update on when it will be ready to go? Great work!

AndyZe:
It's usable right now. There are several people using it on UR3's/UR5's. I think it's mostly cosmetic changes that remain to be made.

compliance_controller/wrench_to_joint_vel_pub/config/compliance_settings.yaml
```diff
@@ -0,0 +1,17 @@
+spin_rate:  50 # Hertz
+jacobian_frame_name: base_link  # MoveIt! planning frame. Often the last link in your tf tree
+force_torque_frame_name: base # The frame of incoming force/torque data
+force_torque_topic: wrench # The topic of incoming force/torque data
```
sjhansen3, line 4: Just a user comment here - Should these settings be able to be modified by a user? I'd love to be able to tune this myself similar to controllers.yaml - perhaps it should even be in controllers.yaml or part of it?

AndyZe: Yep, they're intended to be modified by the user! Adding it to controllers.yaml is a good idea.

bmagyar:
How is this PR going?

AndyZe:
@bmagyar If there's a good chance of it getting merged, provided the changes are made, then I will get on it.  :+1:

AndyZe:
Basically I want to avoid getting sucked into a 6-months-or-more review process because I don't have that much free time anymore. Happy to work on this if I get positive feedback from the maintainers.

AndyZe:
#434 is an update. Closing this PR.

PR #394, opened 2018-12-19
correctly parse joint trajectory options

Karsten1987:
when compiling on my osx machine with clang (x86_64-apple-darwin16.7.0) this actually didn't compile. This fix was necessary.


mathias-luedtke:
I have the same problem on my Ubuntu 18.04 machine (with gcc) as well..
However, Travis and build-farm are still fine.

mathias-luedtke (approved):

bmagyar:
Thanks for the fix!

bmagyar (approved):

PR #402, opened 2019-01-31
use operators instead of aliases

kejxu:
it is pretty straight-forward to just use the operators instead of alias

to access alias for logical `and`/`or`, extra setup is needed (https://docs.microsoft.com/en-us/cpp/cpp/logical-and-operator-amp-amp?view=vs-2017)

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -177,7 +177,7 @@ namespace diff_drive_controller{
 
     // Get joint names from the parameter server
     std::vector<std::string> left_wheel_names, right_wheel_names;
-    if (!getWheelNames(controller_nh, "left_wheel", left_wheel_names) or
+    if (!getWheelNames(controller_nh, "left_wheel", left_wheel_names) ||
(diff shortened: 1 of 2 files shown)


mathias-luedtke (approved):

bmagyar:
Indeed. At the same time those operators are cryptic and keep the language looking medieval while the alternatives are much more readable. 

Perhaps in 10 years we'll get a compiler from Microsoft that implements most of C++
https://cs.nyu.edu/courses/fall11/CSCI-GA.2110-003/documents/c++2003std.pdf Page 40, Table 2.

For now I'm merging this

kejxu:
> Indeed. At the same time those operators are cryptic and keep the language looking medieval while the alternatives are much more readable.
> 
> Perhaps in 10 years we'll get a compiler from Microsoft that implements most of C++
> https://cs.nyu.edu/courses/fall11/CSCI-GA.2110-003/documents/c++2003std.pdf Page 40, Table 2.
> 
> For now I'm merging this

Thanks for the insight and attaching the link! Hopefully this would happen soon =)

PR #407, opened 2019-03-07
Reset commands to zero for stopping controller

sjhansen3:
I have implemented a hardware interface for a Kinova Jaco and am currently using a controller of type velocity_controllers/JointTrajectoryController.

I have witnessed the following problem:
If I command the arm and kill the controller in the middle of the arm motion the last sent velocity command continues to be sent to the arm until the controller is reloaded.

It makes more sense to me that the controller manager should ensure there is a zero command sent when the controller is stopped.

This PR attempts to implement stopping behavior for the joint_trajectory controller which publishes zero commands on stop.

I am a noobie to ros_control so I very well could have some user error going on here. Feel free to take a look at the attached hardware interface, launch file and config file for my controller implementation.

[kinova_ros_control.zip](https://github.com/ros-controls/ros_controllers/files/2939531/kinova_ros_control.zip)

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h
@@ -190,7 +190,15 @@ public:
     }
   }
 
-  void stopping(const ros::Time& /*time*/) {}
+  void stopping(const ros::Time& /*time*/) {
+    if (!joint_handles_ptr_) {return;}
+
+    // zero commands
+    for (unsigned int i = 0; i < pids_.size(); ++i)
+    {
+      (*joint_handles_ptr_)[i].setCommand(0.0);
+    }
+  }
 
   void updateCommand(const ros::Time&     /*time*/,
                      const ros::Duration& period,
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -157,8 +157,16 @@ starting(const ros::Time& time)
 
 template <class SegmentImpl, class HardwareInterface>
 inline void JointTrajectoryController<SegmentImpl, HardwareInterface>::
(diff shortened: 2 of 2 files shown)


sjhansen3:
Whelp, looks like I definitely broke some stuff. I have to assume this feature im thinking I want is more of a bug than a feature... but I still have the same fundamental problem that when I stop and unload a controller I want there to be no commands sent to my robot anymore - curious if there is any advice on usage out there :)

mathias-luedtke:
> Whelp, looks like I definitely broke some stuff.

Just have a closer look ;)
Only one of 3 jobs failed..
Most likely because of the unstable `joint_trajaectory_controller` tests.
I have restarted the failed test.

sjhansen3:
@ipa-mdl did you have a chance to briefly look over my suggested change? I’m a bit confused - should the control manager publish zero commands on stopping or was this avoided for some reason?

sjhansen3:
@ipa-mdl friendly ping. :)

mathias-luedtke:
> should the control manager publish zero commands on stopping or was this avoided for some reason?

The controller manager is able to switch controllers in real-time (-> one control cycle).
So setting the commad to 0 could disrupt the motion.

> It makes more sense to me that the controller manager should ensure there is a zero command sent when the controller is stopped.

All motion might better be stopped if no controller is running anymore (NOT one controller get stopped).
Currently there is no interface to enable the controller manager to stop any motion.
However, it will will inform the hardware interface about the switching.

I would recommend to implement [RobotHW::doSwitch](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1RobotHW.html#ab45a506fd1bab8e371bd109c1bc0cdb3) and reset the commands to 0 for all joints that get stopped, but not started.

bmagyar:
I hope @sjhansen3 you are happy with @ipa-mdl 's suggestion. I agree that we should minimize the amount of hardware-specific logic in the ros_control framework and leave the implementation of `RobotHW` bring these in. I'm closing this issue for now, please feel free to open new ones in the future.

PR #408, opened 2019-03-07
[WIP - DiffDrive] Publish joint trajectory controller state

jordan-palacios:
This is #316 adapted to the `melodic-devel` branch.

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
@@ -144,7 +144,7 @@ namespace diff_drive_controller{
     Odometry odometry_;
 
     /// Joint Trajectory Controller State
-    boost::shared_ptr<realtime_tools::RealtimePublisher<control_msgs::JointTrajectoryControllerState> > joint_trajectory_controller_state_pub_;
+    boost::shared_ptr<realtime_tools::RealtimePublisher<control_msgs::JointTrajectoryControllerState> > controller_state_pub_;
 
     /// Wheel separation, wrt the midpoint of the wheel width:
     double wheel_separation_;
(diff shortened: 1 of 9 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -459,6 +502,71 @@ namespace diff_drive_controller{
       left_wheel_joints_[i].setCommand(vel_left);
       right_wheel_joints_[i].setCommand(vel_right);
     }
+
+    // Publish wheel data:
```
bmagyar: If the big section below is contained within a function, we may not need this comment at all

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -459,6 +502,71 @@ namespace diff_drive_controller{
...
+    if (publish_joint_trajectory_controller_state_ && joint_trajectory_controller_state_pub_->trylock())
+    {
+      joint_trajectory_controller_state_pub_->msg_.header.stamp = time;
```
bmagyar: Please factor the contents of this if block into a function

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
```diff
@@ -131,6 +143,9 @@ namespace diff_drive_controller{
     boost::shared_ptr<realtime_tools::RealtimePublisher<tf::tfMessage> > tf_odom_pub_;
     Odometry odometry_;
 
+    /// Joint Trajectory Controller State
```
bmagyar, line 146: I think it'd be neater to not refer to this publisher as `joint_trajectory_controller_state_pub_` as that only refers to the type. This is in fact just controller state in my opinion. Would it be better to refactor it to reflect that?

bmagyar (changes requested):

bmagyar (approved):

bmagyar:
Thanks for picking this up!

PR #410, opened 2019-03-08
Added param to remap joint_states topic

hvpandya:
Optional parameter to remap the joint_states topic.

### DIFF
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -52,9 +52,16 @@ namespace joint_state_controller
       ROS_ERROR("Parameter 'publish_rate' not set");
       return false;
     }
+    
+    // get topic name 
+    std::string topic_name;
+    if (!controller_nh.getParam("topic", topic_name)){
+      topic_name = "joint_states";
+      ROS_WARN("Parameter 'topic' not set, using default topic name 'joint_states'");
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
This is a variation of #310.
I am sorry, bu I still see no point in adding this functionality.
Why do you need this?

hvpandya:
There are situations in which one might want multiple joint_state topics (desired vs current) concurrently and adding the ability to rename the joint_states topic to something more meaningful in a given context would be great.

For instance renaming it to something like current_state or current/joint_state or desired_state and so on.

mathias-luedtke:
> There are situations in which one might want multiple joint_state topics (desired vs current) concurrently and adding the ability to rename the joint_states topic to something more meaningful in a given context would be great.

This is not the purpose of `joint_state_controller`.

hvpandya:
I did not mean multiple instances of joint_state_controllers, I meant multiple topics with the joint_state msg type. 

The fact that it cannot be renamed to even 'joint_state' instead of 'joint_states' is a bit weird. Feel free to close the PR if you still disagree.

mathias-luedtke:
> The fact that it cannot be renamed to even 'joint_state' instead of 'joint_states' is a bit weird.

You can remap the name quite easily.
Or you could run the driver in its own namespace.

PR #419, opened 2019-05-17
Add service for checking activity of controller

agutenkunst:
I came up with this solution for the problem that I need to know if the robot is currently 
running a trajectory.

I know there is also the query state option but "running" just means that the controller is loaded by the controller_manager and has active resources.

Is there a better way? Any suggestions?

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -306,6 +306,11 @@ bool JointTrajectoryController<SegmentImpl, HardwareInterface>::init(HardwareInt
                                                          &JointTrajectoryController::queryStateService,
                                                          this);
 
+  // ROS API: Provided services
+  query_activity_service_ = controller_nh_.advertiseService("is_active",
+                                                         &JointTrajectoryController::queryActivityService,
+                                                         this);
+
   // Preeallocate resources
   current_state_       = typename Segment::State(n_joints);
   desired_state_       = typename Segment::State(n_joints);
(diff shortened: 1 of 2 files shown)


joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -344,6 +349,34 @@ bool JointTrajectoryController<SegmentImpl, HardwareInterface>::init(HardwareInt
...
+
+  typename Segment::State state;
+
+  bool is_active{false};
+
+  for (unsigned int i = 0; i < joints_.size(); ++i)
+  {
+    typename TrajectoryPerJoint::const_iterator segment_it = sample(curr_traj[i], time_data_.readFromRT()->uptime.toSec(), state);
+
+    if (segment_it != --curr_traj[i].end() || time_data_.readFromRT()->uptime.toSec() <= segment_it->endTime()){
```
martiniil, line 372: I think you want `segment_it < --curr_traj[i].end()` ?

bmagyar:
@agutenkunst sorry for the late reaction on this one.

How about defining a new service in `control_msgs` where querying the state would return a ` 	JointTrajectoryControllerState` message within? I wouldn't be opposed to extending that message if there's a good proposal or alternatively you could add the additional required fields to the proposed new service message. If something like this is added to the controller then I think we should aim to make it as informative as possible.

agutenkunst:
@bmagyar What would be your suggestion for states? Something like running, idle, inactive, error?

bmagyar:
I think it'd be good if we could adopt the ROS2 lifecycle node states (or at least a subset of them)
https://design.ros2.org/articles/node_lifecycle.html

PR #423, opened 2019-06-25
Use C++11 std::this_thread::sleep_for to replace usleep

seanyen:
Use C++11 `std::this_thread::sleep_for` to replace `usleep`, which makes test code more portable for other platforms.

Verified in https://aka.ms/ros project.

### DIFF
--- a/diff_drive_controller/test/diffbot.cpp
+++ b/diff_drive_controller/test/diffbot.cpp
@@ -29,6 +29,7 @@
 
 #include "diffbot.h"
 #include <chrono>
+#include <thread>
 #include <controller_manager/controller_manager.h>
 #include <ros/ros.h>
 #include <rosgraph_msgs/Clock.h>
@@ -78,7 +79,7 @@ int main(int argc, char **argv)
     else
     {
       ROS_DEBUG_STREAM_THROTTLE(1.0, "Control cycle is, elapsed: " << elapsed_secs);
-      usleep((dt.toSec() - elapsed_secs) * 1e6);
+      std::this_thread::sleep_for(std::chrono::duration<double>(dt.toSec() - elapsed_secs));
     }
 
     rosgraph_msgs::Clock clock;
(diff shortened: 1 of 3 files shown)


diff_drive_controller/test/diffbot.cpp
```diff
@@ -78,7 +79,7 @@ int main(int argc, char **argv)
     else
     {
       ROS_DEBUG_STREAM_THROTTLE(1.0, "Control cycle is, elapsed: " << elapsed_secs);
-      usleep((dt.toSec() - elapsed_secs) * 1e6);
+      std::this_thread::sleep_for(std::chrono::microseconds(static_cast<uint64_t>((dt.toSec() - elapsed_secs) * 1e6)));
```
mathias-luedtke: Why not just
```
      std::this_thread::sleep_for(std::chrono::duration<double>(dt.toSec() - elapsed_secs);
```

mathias-luedtke (changes requested):
Please consider removing the factor and set a meaningful commit message (e.g. "use std::this_thread::sleep_for instead of usleep").

IMHO the code should get refactored completely to use `std::chrono` instead of 3 different unit systems (ROS, chrono, raw double).

diff_drive_controller/test/diffbot.cpp
bmagyar: Agreed :+1:

bmagyar:
@ipa-mdl I agree we should aim to use consistent time measurement although I'd add that whenever we want to consider simulated time we have to use ROS time, `chrono` is only good for local measurement.

seanyen:
@ipa-mdl @bmagyar Thanks for the quick feedback! I updated the commit messages and simplified the code by using `std::chrono::duration` as suggested.

mathias-luedtke (approved):
Great, thanks!

bmagyar (approved):
Thanks for this @seanyen , I'll pester the CI until ready :)

PR #433, opened 2019-10-22
Dynamics controllers for Computed Torque Control / Gravity Compensation

francofusco:
Hello! During the last year I worked on some Computed Torque Controllers and I finally thought that this might be a nice addition to ROS control. In this PR, I added a new package, `dynamics_controllers` whose aim is to take into account the inverse dynamics model.

Two controllers are added: one works on serial chains, while the other on whole trees. KDL is used to evaluate the efforts via the Recursive Newton-Euler algorithm. What I think is really interesting is that these controllers work only at a "high-level". In practice, they allow to instanciate a "sub-controller" (operating on an EffortJointInterface) to obtain an acceleration command - technically speaking, the produced command should be an effort, but is interpreted in a different way inside the  package.

One important remark: to compute the dynamics of kinematic trees it is necessary to use the latest version of KDL.

Let me know what you think about it and what should be improved!

### DIFF
--- a/dynamics_controllers/README.md
+++ b/dynamics_controllers/README.md
@@ -126,7 +126,6 @@ should not depend on a specific controller (at least in principle!).
 
 ### Limitations
 
-There are some limitations for these controllers:
 - They all assume a static base. The main implication is that you cannot
   "serially join" controllers for different parts of the robot.
 - They require proper identification of the Dynamic Parameters.
@@ -143,6 +142,8 @@ There are some limitations for these controllers:
     - Allowing to dynamically modify the `Chain`/`Tree` instances
 - [ ] Support for moving bases. This likely requires to re-implement the
       controllers using another library, such as pinocchio, or to expand KDL.
+      A possible workaround is to add a "floating joint" (in the form of a
+      3T3R chain) as the parent of the base.
 - [ ] Additional parameters such as joint inertia and viscous friction.
 - [ ] Allow multiple sub-controllers to manage different joints.
 
@@ -151,7 +152,7 @@ There are some limitations for these controllers:
 
 - [ ] How should the license be pasted? Some headers contain only WillowGarage,
   some many institutions. Should I put myself as well?
-- [ ] Check CMakeLists.txt (install targets etc).
+- [x] Check CMakeLists.txt (install targets etc).
 - [x] Tell about the KDL version that is needed!
-- [ ] Add documentation as in other ROS-controllers packages.
 - [x] Complete `KdlTreeController` so that proper sub-tree extraction is possible.
+- [ ] Enforce effort limits
--- a/dynamics_controllers/README.md
+++ b/dynamics_controllers/README.md
@@ -187,4 +187,4 @@ by yourself or to find proper workarounds.
 - [x] Check CMakeLists.txt (install targets etc).
 - [x] Tell about the KDL version that is needed!
 - [x] Complete `KdlTreeController` so that proper sub-tree extraction is possible.
-- [ ] Enforce effort limits
+- [x] Enforce effort limits
--- a/dynamics_controllers/README.md
+++ b/dynamics_controllers/README.md
@@ -18,9 +18,6 @@ Currently, there exist two controllers: `KdlChainController` and
 They are both based on KDL's implementation of the Recursive Newton-Euler
 algorithm for the inverse dynamics.
 
-**Note**: as of today, you need to install KDL from source (master branch)
-in order to have the `KdlTreeController`.
-
 
 
 ### Gravity Compensation vs Full Inverse Dynamics
--- a/dynamics_controllers/README.md
+++ b/dynamics_controllers/README.md
@@ -18,6 +18,9 @@ Currently, there exist two controllers: `KdlChainController` and
 They are both based on KDL's implementation of the Recursive Newton-Euler
 algorithm for the inverse dynamics.
 
+**Note**: as of today, you need to install KDL from source (master branch)
+in order to have the `KdlTreeController`.
+
 
 
 ### Gravity Compensation vs Full Inverse Dynamics
@@ -147,8 +150,9 @@ There are some limitations for these controllers:
 
 ### Personal todo list:
 
-- How should the license be pasted? Some headers contain only WillowGarage,
-  some many institutions. Should I put myself?
-- Check CMakeLists.txt (install targets etc)
-- Tell about the KDL version that is needed!
-- Add documentation as in other packages
+- [ ] How should the license be pasted? Some headers contain only WillowGarage,
+  some many institutions. Should I put myself as well?
+- [ ] Check CMakeLists.txt (install targets etc).
+- [x] Tell about the KDL version that is needed!
+- [ ] Add documentation as in other ROS-controllers packages.
+- [ ] Complete `KdlTreeController` so that proper sub-tree extraction is possible.
--- a/dynamics_controllers/README.md
+++ b/dynamics_controllers/README.md
@@ -6,9 +6,9 @@ existing effort controllers.
 The controllers contained in this package allow to take into account the
 dynamics of a kinematic chain/tree and to perform Computed Torque Control.
 Controllers are meant to operate in conjunction with existing controllers that
-operate with `hardware_interface::EffortJointInterface`, which act as a
-"*sub-controller*".
-The command from the sub-controller - which would normally be an effort - is
+operate with `hardware_interface::EffortJointInterface`, which act as
+"*sub-controllers*".
+The command from the sub-controllers - which would normally be an effort - is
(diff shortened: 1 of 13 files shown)


bmagyar:
First of all, we need to either get someone to update the ROS Melodic upstream KDL (tricky) or we ship a custom KDL internally with this package (not ideal).
How did you deploy this to robots locally?

francofusco:
In my case I have a local copy of KDL - I wrote the solver for trees so it was convenient for me to have it locally.

As far as I can tell, the only "extra" files not already in Melodic's KDL are those of the tree solver (it's just one header + the associated cpp). So as a quick and easy solution I could simply copy these files locally until the upstream is updated.

gavanderhoorn:
> So as a quick and easy solution I could simply copy these files locally until the upstream is updated.

If this approach is chosen, I would recommend to either only include them in the compilation units that need them (ie: only include in the `.cpp`) or appropriately namespace them, to avoid polluting the include path of dependants.

francofusco:
I manually added the sources for the Tree solver: they are all inside the `src` directory, so that they cannot be included in another package by mistake. In principle, once the solver becomes available, it will be necessary to simply delete the files and edit `CMakeLists.txt` to have everything working.

dynamics_controllers/README.md
```diff
@@ -0,0 +1,158 @@
...
+instead interpreted as an acceleration and injected in the Inverse Dynamic Model
+(IDM) of the mechanism, allowing to better compensate for the nonlinearity of
+robots' Dynamic Models.
+
+Currently, there exist two controllers: `KdlChainController` and
+`KdlTreeController`.
+They are both based on KDL's implementation of the Recursive Newton-Euler
+algorithm for the inverse dynamics.
+
+**Note**: as of today, you need to install KDL from source (master branch)
```
bmagyar: This can be removed now

dynamics_controllers/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+
+### Limitations
+
+- They all assume a static base. The main implication is that you cannot
+  "serially join" controllers for different parts of the robot.
+- They require proper identification of the Dynamic Parameters.
+- The parameters must be provided in URDF format. This is a limitation since
+  the identification will often return a smaller set of regrouped parameters.
+- The sub-controller must control all joints of the internal `Chain`/`Tree`.
```
bmagyar: Couldn't this be detected via parsing the KDL tree?

dynamics_controllers/README.md
```diff
@@ -0,0 +1,158 @@
...
+  from the robot description.
+- `gravity_compensation_only`: as discussed above. Default: `false`.
+
+
+
+### Limitations
+
+- They all assume a static base. The main implication is that you cannot
+  "serially join" controllers for different parts of the robot.
+- They require proper identification of the Dynamic Parameters.
```
bmagyar, line 161: Could you please add some pointers on how one would go about this?

dynamics_controllers/README.md
```diff
@@ -0,0 +1,158 @@
...
+    - Allowing to dynamically modify the `Chain`/`Tree` instances
+- [ ] Support for moving bases. This likely requires to re-implement the
+      controllers using another library, such as pinocchio, or to expand KDL.
+      A possible workaround is to add a "floating joint" (in the form of a
+      3T3R chain) as the parent of the base.
+- [ ] Additional parameters such as joint inertia and viscous friction.
+- [ ] Allow multiple sub-controllers to manage different joints.
+
+
+### Personal todo list:
```
bmagyar: Let's work through this before merging ;)

dynamics_controllers/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+- [ ] How should the license be pasted? Some headers contain only WillowGarage,
```
bmagyar: For stuff you wrote from scratch, add your uni or yourself as license holder. No need to add previous contributors.

dynamics_controllers/include/dynamics_controllers/kdl_tree_controller.h
```diff
@@ -0,0 +1,99 @@
...
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/*
+ * Author: Franco Fusco - franco.fusco@ls2n.fr
+ */
+
+#ifndef DYNAMICS_CONTROLLERS_KDL_TREE_CONTROLLER_H
```
bmagyar: You can use `#pragma once` with c++11

dynamics_controllers/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>dynamics_controllers</name>
+  <version>0.0.0</version>
+  <description>Adds a "dynamics control layer" to standard effort controllers.</description>
+
+  <author email="franco.fusco@ls2n.fr">Franco FUSCO</author>
+  <maintainer email="franco.fusco@ls2n.fr">Franco FUSCO</maintainer>
```
bmagyar, line 8: Could you please add my me as maintainer so I also get the build farm emails?

bmagyar (changes requested):

bmagyar:
I left a few quick points from my phone, excuse my brevity.

francofusco:
Thanks for all the feedback, I will try to address these points in few more commits!

dynamics_controllers/README.md
francofusco: Addressed in last commit

dynamics_controllers/README.md
francofusco: Addressed in last commit

dynamics_controllers/include/dynamics_controllers/kdl_tree_controller.h
francofusco: Addressed in last commit

dynamics_controllers/package.xml
francofusco: Addressed in last commit

dynamics_controllers/README.md
francofusco: If you mean that it is necessary to make sure that all joints are handled by the sub-controller, then yes, it can be checked from the parsed tree and is addressed in the code [here](https://github.com/francofusco/ros_controllers/blob/1fb22f8473153cdd918f5f34fe5ece857637a512/dynamics_controllers/src/dynamics_controller_base.cpp#L185). My comment in the readme is meant to warn users that it is not possible to run dynamics controllers if the sub-controller evaluates the efforts just for a subset of the joints.

dynamics_controllers/README.md
francofusco: That is supposed to be a warning about the fact that model-based computed torque controls require a sufficiently good identification of the parameters. Do you want me to add references about possible identification strategies?

dynamics_controllers/README.md
francofusco: The "personal todo list" is now completely checked! I will actually remove it since it does not make much sense to keep it there :sweat_smile:

francofusco:
In the last two commits I added support for multiple sub-controllers (so that it is not strictly necessary to use a `JointGroup`-like controller) and also made sure that efforts are saturated according to URDF's limits.

I was considering to handle safety limits, however as far as I understand they should be enforced at the very end of the control flow directly by the `RobotHW`, isn't it?

carlosjoserg:
My two-cents from a long-time-user point of view.... and facing new-comers having issues with naming conventions in ros-controls.....

Since the proposed controllers send torque to the `hardware_interface::EffortJointInterface`, their namespace should be `effort_controllers` instead of `dynamic_controllers`, for consistency with the rest of controllers here.

For example, `JointTrajectoryController` can be used as `effort_controllers/JointTrajectoryControllers` which already gives the idea the the output is effort, and a PID is in the middle, if you are already familiar with the naming.

To me `effort_controllers/GravityCompensation` would make more sense.

---

Great work btw! It remided me of an very old work here ... https://github.com/CentroEPiaggio/kuka-lwr/tree/master/lwr_controllers

francofusco:
> their namespace should be `effort_controllers` instead of `dynamic_controllers`

I see your point, and for me it would not be a big deal moving from `dynamics_controllers` to `effort_controllers`. However, I believe that I should then update also the name of the controllers themselves, *.eg.*, from `dynamics_controllers::KdlChainController` to something like `effort_controllers::KdlDynamicsChainController`, since `KdlChainController` would not give much information otherwise. What do you think? `KdlDynamicsChainController` seems a little ugly perhaps, if you have a better suggestion it will be more than welcome! :smile:

> Great work btw!

Thanks! :tada:

PR #434, opened 2019-10-30
Compliance Controller Improvements

AdamPettinger:
This is an update to this pull request to kinetic-devel: [(WIP) New type of compliance controller #393](https://github.com/ros-controls/ros_controllers/pull/393).

Updated to Melodic, added singularity checking, and improved directionally constrained compliance.

Tested on UR5 with ATI Gamma Force/Torque sensor. Videos showing improved directional control:

1. [Compliant only in EEF Forward and EEF roll directions](https://drive.google.com/file/d/1i77SVG8YFoIGYAzKRylypvYj8J8b838f/view?usp=sharing)
2. [Grasping a valve with compliance](https://drive.google.com/file/d/1FAsqDhwRbRmkkmAsjUnJbrf_N2DZzJQ1/view?usp=sharing) and no other user input

PR #437, opened 2019-11-20
(WIP) Compliance Controller to Melodic, with updates

AdamPettinger:
This is an update to the compliant controller first introduced in: [(WIP) New type of compliance controller #393](https://github.com/ros-controls/ros_controllers/pull/393) and [Compliance Controller Improvements #434](https://github.com/ros-controls/ros_controllers/pull/434). The second of which was a messy PR and I decided to close it and open this new, clean one instead.

This new PR includes most of the code fixes requested in #434, as well as new features:
- A tweak to the frame use, so the manipulator moves and rotates around the tool frame (instead of the manipulator last link frame)
- Services to explicitly enable and disable compliance (as opposed to only toggling it)
- Parameters to account for the end-effector mass in the measured wrench, and remove it before doing compliance calculations. 

Videos of the compliance controllers can be seen in the first 2 PR's. A combined video of all of the features, including the ones added for this PR, [can be seen here](https://www.youtube.com/watch?v=osYV9sok3jY).

One of the requested changes in #434 was adding tests, and this is still in progress, hopefully coming soon.


jbeck28:
This may not be the correct spot to post this question, but I'm curious if this can be used to apply a constant force, rather than to avoid a force? I imagine adjusting what is considered zero would have this effect?

AdamPettinger:
> This may not be the correct spot to post this question, but I'm curious if this can be used to apply a constant force, rather than to avoid a force? I imagine adjusting what is considered zero would have this effect?

The equation for finding a velocity in the output frame is v = F / k - F_dot / b where F and F_dot are the force/torque and its derivative, and k and b are the user-specified stiffness and damping. 

I believe this could be adjusted to apply a force rather easily. Hackily, if you increase the deadzone, you will be able to 'drive' the manipulator into applying the desired force (which would be the deadzone value).

Less hacky, the equation above could be modified. Changing the F term to be (desired force - sensed force). Thus it would behave the same with desired force = 0, but also would move the robot into applying the desired force

jbeck28:
@AdamPettinger Thank you very much for your response, it seems like a straightforward modification after sifting through some of the code. I however, ran into some other problems in getting the controller to work with the new `Universal_Robots_ROS_Driver` (the velocity_interface branch), which I'll bring up here. 

The first problem I have is that when planning some simple moves through MoveIt, it sometimes takes a few presses of the Execute button to get the robot to actually move. Further, the robot does not seem to execute any trajectory not planned through the typical "Plan" and "Execute" buttons in MoveIt. For example, I had a program written previously which simply played back a recorded motion, by copying the `joint_states` topic into a trajectory... this no longer functions, and might be related to the following quirk.

I have very little experience with ROS control, but I did manage to set up ROS_Control on an ABB robot, and found that the topics published by the controller,

 `/joint_traj_vel_controller/state/desired/positions.i`

and the corresponding actual positions were published to constantly. However, with this controller, the topic (whose name is different) is only published to briefly when a command is issued, then quickly stops publishing anything. This results in the robot almost always being quite far (perhaps ~.1 radians in joint space) away from the desired position at the end of motions... or at least might be contributing. I'm not entirely sure. 

If I have some fundamental misunderstanding of the operation of this controller, or velocity only controllers in general, please let me know... as I'm not exactly sure where to start to get better performance out of this, nor how to get more of moveit's functionality working (which is paramount, for me anyway). 

Lastly, when moving the robot around by hand with compliance enabled, I found that the movement is a little bit jiggly. The end effector seems to bob up and down very noticeably (on the order of a millimeter). I'm not sure where to look to improve this behavior, but I was drawn to the `wrench_to_joint_vel_pub` node, which at first glance, seems not to be realtime safe, and therefore may not be quite as performant as the UR driver itself...? I almost figured this might be a matter of filtering, as the command velocities do seem to oscillate a bit. Any insight into whether this behavior was noticed by you or anyone else, as well as potential places to look to improve it, would be hugely appreciated. 

Thank you very much for your work on this, and everyone else who contributed. It is pretty sublime, and does almost exactly what I need.

AdamPettinger:
@jbeck28 Thanks for the extensive feedback. I will look into these problems further over the next few days. For now:

> The first problem I have is that when planning some simple moves through MoveIt, it sometimes takes a few presses of the Execute button to get the robot to actually move

I am able to replicate this consistently and will take a look. At first glance, this problem appears to be in the compliance controller or its parent traj_or_jog_controller 

>  Further, the robot does not seem to execute any trajectory not planned through the typical "Plan" and "Execute" buttons in MoveIt. For example, I had a program written previously which simply played back a recorded motion, by copying the joint_states topic into a trajectory... this no longer functions, and might be related to the following quirk

Our use of the compliant controller so far has only been through MoveIt Plan + Execute actions and joint velocity control. Creating a valid trajectory by manually populating the trajectory message should work, I will test some here and try to get it working. In the mean time, if your code records the joint velocities (as part of the joint states) and can play them back on the `compliant_controller/velocity_command` topic, the controller would treat the joint velocity commands as a jog command, and this could replicate your recorded motion

> This results in the robot almost always being quite far (perhaps ~.1 radians in joint space) away from the desired position at the end of motions

Not sure about this one, or if it is linked to motions stopping early. I will look at it

> Lastly, when moving the robot around by hand with compliance enabled, I found that the movement is a little bit jiggly. 

From our experiences testing the compliance controller, the most important thing for better performance was the source of the measured Wrench. Testing with the UR3 and UR5's showed mixed results with the UR drivers published wrench topic. The UR3 was OK (not perfect but definitely compliant), and the UR5 was fairly bad (oscillatory behavior, undesired motion in response to bad wrench data). I imagine this varies across individual UR5's and 3's as well. Placing a nice load cell at the end of our UR5 dramatically increased performance. How are you measuring the contact forces and torques?

The `wrench_to_joint_vel_pub` and it's parameters are definitely the place to look to try and fix this. These include a filter and the stiffness/damping parameters. I will include the parameters below that are currently working great for our setup. Bumping the `low_pass_filter_param` and `stiffness` up to ~10 and ~10000 respectively should give you a good idea if filtering (or a lowered response in the stiffness case) will help.

In general, I have worked more on the `wrench_to_joint_vel_pub` side and less on the ros control side (@AndyZe wrote the first version of both, and I was making updates mostly to `wrench_to_joint_vel_pub`). I will look into some of the things you mentioned and try to get them squared away, and would be more than happy to help you get running on your robot.

Parameters:
```
compliance_library:
  stiffness: [8000, 1000, 1000, 5, 40, 60] # First three element are Newtons/(m/s), last 3 are N*m/(rad/s). Should be positive.
  damping: [50000, 10000, 10000, 300, 600, 600] # First three element are Newtons/m, last 3 are N*m/rad. Should be positive.
  low_pass_filter_param: 3 # Larger > more smoothing on compliance calcs, but more lag
  deadband: [1.5, 2.5, 2.5, 0.3, 2, 2] # No reaction until force exceeds X [N, N, N, Nm, Nm, Nm]
  end_condition_wrench: [80, 80, 80, 60, 60, 60] # Safety feature. Halt if force exceeds [N, N, N, Nm, Nm, Nm]
  highest_allowable_force: 88 # Safety feature. Halt if force magnitude exceeds this across all dimensions [N]
  highest_allowable_torque: 50 # Safety feature. Halt if torque magnitude exceeds this across all dimensions [Nm]
  default_compliant_dimensions: [1, 1, 1, 0, 0, 0] 
```

jbeck28:
@AdamPettinger Thank you for the help! That was informative. 

>  Creating a valid trajectory by manually populating the trajectory message should work, I will test some here and try to get it working.

I also found that I'm not able to execute cartesian paths planned through the builtin moveit cartesian path planning functionality. It seems like typical `planning_interface::MoveGroupInterface` commands don't work. I've not tested just yet whether or not I can execute trajectories which were not planned using the cartesian path functionality, my gut would say that it should work since that's essentially what the buttons we were talking about do (I believe). 

My initial crack at replaying recorded joint_states was only to populate the trajectory with positions, which worked fine since the `Universal_Robots_ROS_Driver` has a position controller built in. However, I modified the code during troubleshooting of this driver to copy velocities as well, to no avail. 

> the most important thing for better performance was the source of the measured Wrench.

I have a UR5e, and was using the built-in F/T sensor (Robotiq I believe). Once I get this working, I intend to purchase an ATI load cell, so that should fix the issue. I'm happy to hear that this is easily fixed. I'll try the filtering settings you provide, and see how that does. 


Thanks again for your help and support! If there's anything I can do to help, let me know.

AndyZe:
@AdamPettinger please add yourself as an author, too, in package.xml. That's definitely deserved.

AdamPettinger:
Ok a few updates. I just committed a simple change, but seemed to solve a few of the problems. Also, while digging I uncovered a few things that might help you out @jbeck28 

There are 3 main ways I was getting the trajectory execution to fail:

1. **Path tolerances violation**. Was occurring in the middle of trajectories when the path tolerances of the joints were violated. This is also probably responsible for the robot position by being off by such a large amount. After the recent change, I was unable to get this failure mode to happen again. I also plotted a bunch of trajectories desired and actual joint positions. The largest deviation at the goal I found was ~0.002 radians. 
2. **Initial state error**. Happens regularly if I use 2 different move interfaces (e.g. one in the Rviz GUI and another one in my own code) because the robot position changes and the initial position for one of them (e.g. rviz) does not update until it tries and fails to execute. This is still happening, but only for plans through the RVIZ interface. I also testing placing a heavy-ish object in the robot gripper and could get this error with compliance on. The robot would be moving slowly (due to the weight of the object), so after planning the robot position would be different from the plan. Can be solved by biasing the compliance after the object is gripped. Also, you should probably bias your FT stream when you start the compliance controller or you might get the same effect
3. **Execution timeout**. The robot does not move, then a message pops up saying the execution timed out. I still do not know what is going on here, but I think it might be communication problem, or in the move interface (not in the compliance controller). This has been rare for me (1 out of ~30-50 trajectories maybe). When it happens, I do not see the goal published to the `compliance_controller/goal` topic that the controller listens to, which is why I think it is not the controller.

I had some other thoughts while troubleshooting this. You should try disabling compliance and running a few trajectories. You should get identical performance to a normal `JointTrajectoryController` as the compliance controller inherits from this and runs its `update` function while not in compliant mode. Biasing before you do a trajectory is also a good idea, as a F/T measurement will cause the robot to drift during execution - which you want, but only if it is actually in contact with something. Also, because of the inheritance from `JointTrajectoryController`, the compliant controller should be able to do the Cartesian path plans just like that one. I am still looking into that.

I am still looking through some of the other things you've mentioned, just wanted to give a quick update and hopefully get some better performance now

jbeck28:
@AdamPettinger Thanks for the updates, and apologies for my late response. I hadn't had time to test out your modifications until today. Unfortunately, I haven't noticed too much of an improvement in my use case. Although I still haven't tried to publish my recorded joint velocities to the `compliance_controller/command_velocity` topic, which I may give a try soon. In lieu of this, I computed accelerations and velocities from timestamps to fully define the trajectory I was successfully forwarding to the topic `/compliance_controller/follow_joint_trajectory/goal`. This still (as you eluded to) did not execute. No errors were thrown, the robot simply didn't move. Anyway, I have noticed that motions planned through the `move_group_interface` in my code do work reliably.

Also, the behavior I was mentioning before, about the robot not being in quite the correct spot at the end of motions, it seems that the controller relinquishes control once a given trajectory has reached the time of its last time stamp, regardless of how close the goal position is. The following image demonstrates this. 
compliance_controller_actual_desired


Again, thank you very much for your work on this, and the consistent support!

jbeck28:
@AdamPettinger 
I just wanted to update this as I've had success improving the robot reaching goals by simply adjusting the goal_time and PID values. It now works quite well. However, I still cannot get it to execute cartesian paths. No errors are thrown, it just won't move. I am seeing the trajectory being published on `compliance_controller/follow_joint_trajectory/goal`

gavanderhoorn:
@jbeck28: as you mention using `ur_robot_driver`, it would be greatly appreciated if you could share your tuned PID values. Even if they are application-specific (ie: depend on payload and/or inertias of EEFs), having some ballpark figures and an idea of how they influence the behaviour of the robot would be great.

Could you perhaps post them in an issue on `Universal_Robots_ROS_Driver`?

jbeck28:
@gavanderhoorn 
Posted an issue [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/124), not exactly sure if this was the format you were looking for, but it gets the information out there.

rfn123:
> The equation for finding a velocity in the output frame is v = F / k - F_dot / b

I am really sorry for bumping this issue, but I am working on a similar compliance task now and stumbled upon this PR. I am new to this topic and trying to understand the maths behind the velocity calculation. Is there any recommended reference (literature or similar) where the above equation is somehow adressed @AdamPettinger ?

AdamPettinger:
> Is there any recommended reference (literature or similar) where the above equation is somehow adressed @AdamPettinger ?

Hi @rfn123, this work was partially adapted from this [Master's thesis](https://repositories.lib.utexas.edu/bitstream/handle/2152/39083/VONSTERNBERG-THESIS-2016.pdf?sequence=1). Chapter 2.2 specifically lists a lot of great references for different compliant control schemes. 

The specific equation you mentioned is addressed in the implementation chapter

AdamPettinger:
Closing this for now as I haven't really had time to work on it. In the future this is something I might come back to

PR #442, opened 2020-01-10
Split JointTrajectoryController::update() into smaller pieces

martiniil:
This introduces some subfunctions of update() in order to:
* make it more readable,
* facilitate overriding in derived classes.

The main objective is that we want to add a check on the updated states before they are set in the hardware_interface. Currently, we call the parent update() in the derived class and perform the check afterwards. But actually we need it before the commands are set.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -250,7 +250,7 @@ protected:
    * \brief Internal update function.
    * Update time data, current states and errors and check tolerances.
    */
-  virtual void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period);
+  virtual void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period, TimeData& time_data);
 
   /**
    * \brief Set current active goal to succeeded (if present).
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -248,21 +248,18 @@ protected:
 
   /**
    * \brief Internal update function.
-   *
    * Update time data, current states and errors and check tolerances.
    */
   void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period);
 
   /**
    * \brief Set current active goal to succeeded (if present).
-   *
    * \note This method is realtime-safe.
    */
   void setGoalAsSucceeded();
 
   /**
    * \brief Set action feedback (if active goal present).
-   *
    * \note This method is realtime-safe.
    */
   void setActionFeedback();
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -246,7 +246,40 @@ protected:
    */
   void setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh=RealtimeGoalHandlePtr());
 
+  /**
+   * \brief Internal update function.
+   *
+   * Update time data, current states and errors and check tolerances.
+   */
+  void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period);
+
+  /**
+   * \brief Set current active goal to succeeded (if present).
+   *
+   * \note This method is realtime-safe.
+   */
+  void setGoalAsSucceeded();
+
+  /**
+   * \brief Set action feedback (if active goal present).
+   *
+   * \note This method is realtime-safe.
+   */
+  void setActionFeedback();
+
+  /**
+   * \brief Check if all segments finished successfully.
+   */
+  bool allSegmentsCompleted();
+
 };
+
+template <class SegmentImpl, class HardwareInterface>
+inline bool JointTrajectoryController<SegmentImpl, HardwareInterface>::
+allSegmentsCompleted()
+{
+  return successful_joint_traj_.count() == joints_.size();
+}
 
 } // namespace
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -270,13 +270,13 @@ protected:
   /**
    * \brief Check if all segments finished successfully.
    */
-  bool allSegmentsCompleted();
+  bool allSegmentsCompleted() const;
 
 };
 
 template <class SegmentImpl, class HardwareInterface>
 inline bool JointTrajectoryController<SegmentImpl, HardwareInterface>::
-allSegmentsCompleted()
+allSegmentsCompleted() const
 {
   return successful_joint_traj_.count() == joints_.size();
 }
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -250,19 +250,19 @@ protected:
    * \brief Internal update function.
    * Update time data, current states and errors and check tolerances.
    */
-  void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period);
+  virtual void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period);
 
   /**
    * \brief Set current active goal to succeeded (if present).
    * \note This method is realtime-safe.
    */
-  void setGoalAsSucceeded();
+  virtual void setGoalAsSucceeded();
 
   /**
    * \brief Set action feedback (if active goal present).
    * \note This method is realtime-safe.
    */
-  void setActionFeedback();
+  virtual void setActionFeedback();
 
   /**
    * \brief Check if all segments finished successfully.
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -253,6 +253,11 @@ protected:
   virtual void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period, TimeData& time_data);
 
   /**
+   * \brief Update commands on the hardware interface.
+   */
+  virtual void updateHardwareInterface(const TimeData& time_data);
+
+  /**
    * \brief Set current active goal to succeeded (if present).
    * \note This method is realtime-safe.
    */
@@ -270,6 +275,14 @@ protected:
   bool allSegmentsCompleted() const;
 
 };
+
+template <class SegmentImpl, class HardwareInterface>
+inline void JointTrajectoryController<SegmentImpl, HardwareInterface>::
+updateHardwareInterface(const TimeData& time_data)
+{
+  hw_iface_adapter_.updateCommand(time_data.uptime, time_data.period,
+                                  desired_state_, state_error_);
+}
 
 template <class SegmentImpl, class HardwareInterface>
 inline bool JointTrajectoryController<SegmentImpl, HardwareInterface>::
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -348,145 +348,20 @@ template <class SegmentImpl, class HardwareInterface>
 void JointTrajectoryController<SegmentImpl, HardwareInterface>::
 update(const ros::Time& time, const ros::Duration& period)
 {
-  // Get currently followed trajectory
-  TrajectoryPtr curr_traj_ptr;
-  curr_trajectory_box_.get(curr_traj_ptr);
-  Trajectory& curr_traj = *curr_traj_ptr;
+  updateStateAndTimeData(time, period);
 
-  // Update time data
-  TimeData time_data;
-  time_data.time   = time;                                     // Cache current time
-  time_data.period = period;                                   // Cache current control period
-  time_data.uptime = time_data_.readFromRT()->uptime + period; // Update controller uptime
-  time_data_.writeFromNonRT(time_data); // TODO: Grrr, we need a lock-free data structure here!
-
-  // NOTE: It is very important to execute the two above code blocks in the specified sequence: first get current
-  // trajectory, then update time data. Hopefully the following paragraph sheds a bit of light on the rationale.
-  // The non-rt thread responsible for processing new commands enqueues trajectories that can start at the _next_
-  // control cycle (eg. zero start time) or later (eg. when we explicitly request a start time in the future).
-  // If we reverse the order of the two blocks above, and update the time data first; it's possible that by the time we
-  // fetch the currently followed trajectory, it has been updated by the non-rt thread with something that starts in the
-  // next control cycle, leaving the current cycle without a valid trajectory.
-
-  // Update current state and state error
-  for (unsigned int i = 0; i < joints_.size(); ++i)
+  if (allSegmentsCompleted())
   {
-    current_state_.position[i] = joints_[i].getPosition();
-    current_state_.velocity[i] = joints_[i].getVelocity();
-    // There's no acceleration data available in a joint handle
-
-    typename TrajectoryPerJoint::const_iterator segment_it = sample(curr_traj[i], time_data.uptime.toSec(), desired_joint_state_);
-    if (curr_traj[i].end() == segment_it)
-    {
-      // Non-realtime safe, but should never happen under normal operation
-      ROS_ERROR_NAMED(name_,
-                      "Unexpected error: No trajectory defined at current time. Please contact the package maintainer.");
-      return;
-    }
-    desired_state_.position[i] = desired_joint_state_.position[0];
-    desired_state_.velocity[i] = desired_joint_state_.velocity[0];
-    desired_state_.acceleration[i] = desired_joint_state_.acceleration[0]; ;
-
-    state_joint_error_.position[0] = angles::shortest_angular_distance(current_state_.position[i],desired_joint_state_.position[0]);
-    state_joint_error_.velocity[0] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
-    state_joint_error_.acceleration[0] = 0.0;
-
-    state_error_.position[i] = angles::shortest_angular_distance(current_state_.position[i],desired_joint_state_.position[0]);
-    state_error_.velocity[i] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
-    state_error_.acceleration[i] = 0.0;
-
-    //Check tolerances
-    const RealtimeGoalHandlePtr rt_segment_goal = segment_it->getGoalHandle();
-    if (rt_segment_goal && rt_segment_goal == rt_active_goal_)
-    {
-      // Check tolerances
-      if (time_data.uptime.toSec() < segment_it->endTime())
-      {
-        // Currently executing a segment: check path tolerances
-        const SegmentTolerancesPerJoint<Scalar>& joint_tolerances = segment_it->getTolerances();
-        if (!checkStateTolerancePerJoint(state_joint_error_, joint_tolerances.state_tolerance))
-        {
-          if (verbose_)
-          {
-            ROS_ERROR_STREAM_NAMED(name_,"Path tolerances failed for joint: " << joint_names_[i]);
-            checkStateTolerancePerJoint(state_joint_error_, joint_tolerances.state_tolerance, true);
-          }
-          rt_segment_goal->preallocated_result_->error_code =
-          control_msgs::FollowJointTrajectoryResult::PATH_TOLERANCE_VIOLATED;
-          rt_segment_goal->setAborted(rt_segment_goal->preallocated_result_);
-          rt_active_goal_.reset();
-          successful_joint_traj_.reset();
-        }
-      }
-      else if (segment_it == --curr_traj[i].end())
-      {
-        if (verbose_)
-          ROS_DEBUG_STREAM_THROTTLE_NAMED(1,name_,"Finished executing last segment, checking goal tolerances");
-
-        // Controller uptime
-        const ros::Time uptime = time_data_.readFromRT()->uptime;
-
-        // Checks that we have ended inside the goal tolerances
-        const SegmentTolerancesPerJoint<Scalar>& tolerances = segment_it->getTolerances();
-        const bool inside_goal_tolerances = checkStateTolerancePerJoint(state_joint_error_, tolerances.goal_state_tolerance);
-
-        if (inside_goal_tolerances)
-        {
-          successful_joint_traj_[i] = 1;
-        }
-        else if (uptime.toSec() < segment_it->endTime() + tolerances.goal_time_tolerance)
-        {
-          // Still have some time left to meet the goal state tolerances
-        }
-        else
-        {
-          if (verbose_)
-          {
-            ROS_ERROR_STREAM_NAMED(name_,"Goal tolerances failed for joint: "<< joint_names_[i]);
-            // Check the tolerances one more time to output the errors that occurs
-            checkStateTolerancePerJoint(state_joint_error_, tolerances.goal_state_tolerance, true);
-          }
-
-          rt_segment_goal->preallocated_result_->error_code = control_msgs::FollowJointTrajectoryResult::GOAL_TOLERANCE_VIOLATED;
-          rt_segment_goal->setAborted(rt_segment_goal->preallocated_result_);
-          rt_active_goal_.reset();
-          successful_joint_traj_.reset();
-        }
-      }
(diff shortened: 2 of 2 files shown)


joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,41 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
...
+   * \brief Set action feedback (if active goal present).
+   *
+   * \note This method is realtime-safe.
+   */
+  void setActionFeedback();
+
+  /**
+   * \brief Check if all segments finished successfully.
+   */
+  bool allSegmentsCompleted();
```
bmagyar: this could definitely be a `const` method

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -348,145 +348,20 @@ template <class SegmentImpl, class HardwareInterface>
...
-    current_active_goal->setSucceeded(current_active_goal->preallocated_result_);
-    current_active_goal.reset(); // do not publish feedback
-    rt_active_goal_.reset();
-    successful_joint_traj_.reset();
+    setGoalAsSucceeded();
   }
 
   // Hardware interface adapter: Generate and send commands
-  hw_iface_adapter_.updateCommand(time_data.uptime, time_data.period,
+  hw_iface_adapter_.updateCommand(time_data_.readFromRT()->uptime, time_data_.readFromRT()->period,
```
bmagyar: let's keep the cached approach, this locks the mutex twice but also, period is taken from the argument of `update()` in the original code

if you keep the caching timedata part at the beginning of this function, you can pass those values around without the need for many mutex calls

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +686,152 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+
+template <class SegmentImpl, class HardwareInterface>
+void JointTrajectoryController<SegmentImpl, HardwareInterface>::
+setGoalAsSucceeded()
+{
+  RealtimeGoalHandlePtr current_active_goal(rt_active_goal_);
+  if (current_active_goal)
+  {
+    current_active_goal->preallocated_result_->error_code = control_msgs::FollowJointTrajectoryResult::SUCCESSFUL;
+    current_active_goal->setSucceeded(current_active_goal->preallocated_result_);
```
bmagyar, line 811: The original code had a `current_active_goal.reset()` here. Could you elaborate please?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,41 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
    */
   void setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh=RealtimeGoalHandlePtr());
 
+  /**
+   * \brief Internal update function.
+   *
+   * Update time data, current states and errors and check tolerances.
+   */
+  void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period);
```
bmagyar: Overriding in derived classes will need some `virtual`s here

bmagyar (changes requested):
I'm in support of this idea, let's ensure that these smaller bits have

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,41 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
    */
   void setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh=RealtimeGoalHandlePtr());
 
+  /**
+   * \brief Internal update function.
+   *
```
bmagyar: Could we please remove these newlines in the brief doc? and also between `\brief` and `\note` blocks

bmagyar (changes requested):

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
martiniil: Done.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
martiniil: Before, `current_active_goal` was reused later in `update()`. Now its scope ends with `setGoalAsSucceeded()` so this has no effect. The main thing is to keep `rt_active_goal_.reset();`

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
martiniil: Done

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
martiniil: Thanks! Done.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
martiniil: Ok, reading period from `time_data_` is unnecessary.

I can not keep the timedata caching at the beginning since I would have to break up the structure of the new update-function:

>   // NOTE: It is very important to execute the two above code blocks in the specified sequence: first get current
  // trajectory, then update time data.

My latest approach is passing time_data by reference. What do you think?

martiniil:
> I'm in support of this idea, let's ensure that these smaller bits have

Thanks for your review. Is something missing here?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+    state_joint_error_.velocity[0] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
+    state_joint_error_.acceleration[0] = 0.0;
+
+    state_error_.position[i] = angles::shortest_angular_distance(current_state_.position[i],desired_joint_state_.position[0]);
+    state_error_.velocity[i] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
+    state_error_.acceleration[i] = 0.0;
+
+    //Check tolerances
+    const RealtimeGoalHandlePtr rt_segment_goal = segment_it->getGoalHandle();
+    if (rt_segment_goal && rt_segment_goal == rt_active_goal_)
```
HiddenBug, line 743: Please use `continue` to avoid nesting caused by this if statement. Deep nesting simply makes it difficult to read and understand code.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+    current_state_.velocity[i] = joints_[i].getVelocity();
+    // There's no acceleration data available in a joint handle
+
+    typename TrajectoryPerJoint::const_iterator segment_it = sample(curr_traj[i], time_data.uptime.toSec(), desired_joint_state_);
+    if (curr_traj[i].end() == segment_it)
+    {
+      // Non-realtime safe, but should never happen under normal operation
+      ROS_ERROR_NAMED(name_,
+                      "Unexpected error: No trajectory defined at current time. Please contact the package maintainer.");
+      return;
```
HiddenBug, line 727: Originally, this `return` caused the originally `update()` function to return. Now, however, only the sub-function returns. So, in other words, the original behavior was altered. Is this intended?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+    successful_joint_traj_.reset();
+  }
+}
+
+template <class SegmentImpl, class HardwareInterface>
+void JointTrajectoryController<SegmentImpl, HardwareInterface>::
+setActionFeedback()
+{
+  RealtimeGoalHandlePtr current_active_goal(rt_active_goal_);
+  if (current_active_goal)
```
HiddenBug, line 822: You could also write:
```
if (!current_active_goal)
{
return;
}
```
This avoids unnecessary nesting and allows easier code editing in the future.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+    rt_active_goal_.reset();
+    successful_joint_traj_.reset();
+  }
+}
+
+template <class SegmentImpl, class HardwareInterface>
+void JointTrajectoryController<SegmentImpl, HardwareInterface>::
+setActionFeedback()
+{
+  RealtimeGoalHandlePtr current_active_goal(rt_active_goal_);
```
HiddenBug: Please use pure pointer because you do not want to take ownership of `rt_active_goal_`.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+      }
+    }
+  }
+}
+
+template <class SegmentImpl, class HardwareInterface>
+void JointTrajectoryController<SegmentImpl, HardwareInterface>::
+setGoalAsSucceeded()
+{
+  RealtimeGoalHandlePtr current_active_goal(rt_active_goal_);
```
HiddenBug: Please use pure pointer because you do not want to take ownership of rt_active_goal_.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+    desired_state_.velocity[i] = desired_joint_state_.velocity[0];
+    desired_state_.acceleration[i] = desired_joint_state_.acceleration[0]; ;
+
+    state_joint_error_.position[0] = angles::shortest_angular_distance(current_state_.position[i],desired_joint_state_.position[0]);
+    state_joint_error_.velocity[0] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
+    state_joint_error_.acceleration[0] = 0.0;
+
+    state_error_.position[i] = angles::shortest_angular_distance(current_state_.position[i],desired_joint_state_.position[0]);
+    state_error_.velocity[i] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
+    state_error_.acceleration[i] = 0.0;
```
HiddenBug, line 739: This calculation was already done in the lines above. You could just copy from `state_joint_error_`.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+          {
+            ROS_ERROR_STREAM_NAMED(name_,"Goal tolerances failed for joint: "<< joint_names_[i]);
+            // Check the tolerances one more time to output the errors that occurs
+            checkStateTolerancePerJoint(state_joint_error_, tolerances.goal_state_tolerance, true);
+          }
+
+          rt_segment_goal->preallocated_result_->error_code = control_msgs::FollowJointTrajectoryResult::GOAL_TOLERANCE_VIOLATED;
+          rt_segment_goal->setAborted(rt_segment_goal->preallocated_result_);
+          rt_active_goal_.reset();
+          successful_joint_traj_.reset();
```
HiddenBug, line 796: Code duplication: These lines are basically are a copy from line 757-761 except for the `error_code`. Maybe create new function for these lines of code?

You even already created a function on this abstraction level by implementing function `setGoalAsSucceeded()`.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+            checkStateTolerancePerJoint(state_joint_error_, joint_tolerances.state_tolerance, true);
+          }
+          rt_segment_goal->preallocated_result_->error_code =
+          control_msgs::FollowJointTrajectoryResult::PATH_TOLERANCE_VIOLATED;
+          rt_segment_goal->setAborted(rt_segment_goal->preallocated_result_);
+          rt_active_goal_.reset();
+          successful_joint_traj_.reset();
+        }
+      }
+      else if (segment_it == --curr_traj[i].end())
```
HiddenBug, line 764: Difficult to read. Maybe create inline helper function?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+      {
+        if (verbose_)
+          ROS_DEBUG_STREAM_THROTTLE_NAMED(1,name_,"Finished executing last segment, checking goal tolerances");
+
+        // Controller uptime
+        const ros::Time uptime = time_data.uptime;
```
HiddenBug, line 770: Why suddenly a copy? For what is it needed, really?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+          rt_active_goal_.reset();
+          successful_joint_traj_.reset();
+        }
+      }
+      else if (segment_it == --curr_traj[i].end())
+      {
+        if (verbose_)
+          ROS_DEBUG_STREAM_THROTTLE_NAMED(1,name_,"Finished executing last segment, checking goal tolerances");
+
+        // Controller uptime
```
HiddenBug, line 769: Redundant -> remove.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+
+    state_joint_error_.position[0] = angles::shortest_angular_distance(current_state_.position[i],desired_joint_state_.position[0]);
+    state_joint_error_.velocity[0] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
+    state_joint_error_.acceleration[0] = 0.0;
+
+    state_error_.position[i] = angles::shortest_angular_distance(current_state_.position[i],desired_joint_state_.position[0]);
+    state_error_.velocity[i] = desired_joint_state_.velocity[0] - current_state_.velocity[i];
+    state_error_.acceleration[i] = 0.0;
+
+    //Check tolerances
```
HiddenBug, line 741: Redundant -> remove.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+
+  // NOTE: It is very important to execute the two above code blocks in the specified sequence: first get current
+  // trajectory, then update time data. Hopefully the following paragraph sheds a bit of light on the rationale.
+  // The non-rt thread responsible for processing new commands enqueues trajectories that can start at the _next_
+  // control cycle (eg. zero start time) or later (eg. when we explicitly request a start time in the future).
+  // If we reverse the order of the two blocks above, and update the time data first; it's possible that by the time we
+  // fetch the currently followed trajectory, it has been updated by the non-rt thread with something that starts in the
+  // next control cycle, leaving the current cycle without a valid trajectory.
+
+  // Update current state and state error
```
HiddenBug, line 714: Why this comment? It's also incomplete because there are also checks performed.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+  // Update time data
+  time_data.time   = time;                                     // Cache current time
+  time_data.period = period;                                   // Cache current control period
+  time_data.uptime = time_data_.readFromRT()->uptime + period; // Update controller uptime
+  time_data_.writeFromNonRT(time_data); // TODO: Grrr, we need a lock-free data structure here!
+
+  // NOTE: It is very important to execute the two above code blocks in the specified sequence: first get current
+  // trajectory, then update time data. Hopefully the following paragraph sheds a bit of light on the rationale.
+  // The non-rt thread responsible for processing new commands enqueues trajectories that can start at the _next_
+  // control cycle (eg. zero start time) or later (eg. when we explicitly request a start time in the future).
```
HiddenBug, line 709: Can be removed because these lines only marginally adds useful information.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
...
+        // Controller uptime
+        const ros::Time uptime = time_data.uptime;
+
+        // Checks that we have ended inside the goal tolerances
+        const SegmentTolerancesPerJoint<Scalar>& tolerances = segment_it->getTolerances();
+        const bool inside_goal_tolerances = checkStateTolerancePerJoint(state_joint_error_, tolerances.goal_state_tolerance);
+
+        if (inside_goal_tolerances)
+        {
+          successful_joint_traj_[i] = 1;
```
HiddenBug: Magic number.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -811,6 +688,151 @@ setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh)
   curr_trajectory_box_.set(hold_trajectory_ptr_);
 }
 
+template <class SegmentImpl, class HardwareInterface>
+void JointTrajectoryController<SegmentImpl, HardwareInterface>::
+updateStateAndTimeData(const ros::Time& time, const ros::Duration& period, TimeData& time_data)
```
HiddenBug, line 693: Why does this function do two things (even indicated by the "and" in the name)? For the sake of clarity I suggest to create functions with one responsibility only. (For more information about the single responsibility see also "Clean Code" by Robert C. Martin Chapter 3  "Functions")

I suggest separating the getting of the trajectory and the time update from this function and just make this function about the state updates and checks.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,51 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
...
+   * \brief Set current active goal to succeeded (if present).
+   * \note This method is realtime-safe.
+   */
+  virtual void setGoalAsSucceeded();
+
+  /**
+   * \brief Set action feedback (if active goal present).
+   * \note This method is realtime-safe.
+   */
+  virtual void setActionFeedback();
```
HiddenBug, line 270: I'm not a fan of all this `virtual` stuff because, in my opinion, it's violates the encapsulation. What is it you are trying to archive? Only expose the stuff, for what you actually have a usage.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,51 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
...
   void setHoldPosition(const ros::Time& time, RealtimeGoalHandlePtr gh=RealtimeGoalHandlePtr());
 
+  /**
+   * \brief Internal update function.
+   * Update time data, current states and errors and check tolerances.
+   */
+  virtual void updateStateAndTimeData(const ros::Time& time, const ros::Duration& period, TimeData& time_data);
+
+  /**
+   * \brief Update commands on the hardware interface.
```
HiddenBug, line 256: Redundant -> Remove or improve. What is it the user of the function should know?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,51 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
...
+   */
+  virtual void updateHardwareInterface(const TimeData& time_data);
+
+  /**
+   * \brief Set current active goal to succeeded (if present).
```
HiddenBug, line 261: Redundant -> Remove or improve. What is it the user of the function should know?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,51 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
...
+   * \note This method is realtime-safe.
+   */
+  virtual void setGoalAsSucceeded();
+
+  /**
+   * \brief Set action feedback (if active goal present).
```
HiddenBug, line 267: Redundant -> Remove or improve. What is it the user of the function should know?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
```diff
@@ -246,8 +246,51 @@ class JointTrajectoryController : public controller_interface::Controller<Hardwa
...
+   * \note This method is realtime-safe.
+   */
+  virtual void setActionFeedback();
+
+  /**
+   * \brief Check if all segments finished successfully.
```
HiddenBug, line 273: Redundant -> Remove or improve. What is it the user of the function should know?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -348,144 +348,21 @@ template <class SegmentImpl, class HardwareInterface>
 void JointTrajectoryController<SegmentImpl, HardwareInterface>::
 update(const ros::Time& time, const ros::Duration& period)
 {
-  // Get currently followed trajectory
-  TrajectoryPtr curr_traj_ptr;
-  curr_trajectory_box_.get(curr_traj_ptr);
-  Trajectory& curr_traj = *curr_traj_ptr;
-
-  // Update time data
+  // Prepare time data caching
```
HiddenBug, line 351: What information does this comment give me? -> Remove

HiddenBug (changes requested):

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
hslusarek: !!! Changed behavior:
The original lines looked like this:
```
const ros::Time uptime = time_data_.readFromRT()->uptime;
```
But now you use the local variable `time_data` which is:
```
time_data.uptime = time_data_.readFromRT()->uptime + period
```
Is this change intended?

hslusarek:
@martiniil I opened an alternative PR regarding the refactoring of the update function. Please see PR #450. 

Please note:
PR #450 only deals with the improvement of the update function. It does not concern itself with:
> The main objective is that we want to add a check on the updated states before they are set in the hardware_interface

For this I will open an extra PR.

martiniil:
> @martiniil I opened an alternative PR regarding the refactoring of the update function. Please see PR #450.
> 
> Please note:
> PR #450 only deals with the improvement of the update function. It does not concern itself with:
> 
> > The main objective is that we want to add a check on the updated states before they are set in the hardware_interface
> 
> For this I will open an extra PR.

:+1:

PR #443, opened 2020-01-12
Remove rqt_joint_trajectory_controller from metapackage

### LINKED ISSUE #440: Remove rqt_joint_trajectory_controller dependency
For consideration:

The `ros_control` metapackage does not depend on `rqt_controller_manager`, so installing `ros_control` does not pull in any GUI dependencies.

In contrast, the `ros_controllers` metapackage _does_ depend on `rqt_joint_trajectory_controller`, and so it _does_ pull in GUI dependencies via `rqt`.

Thoughts on removing this dependency from the metapackage?

matthew-reynolds:
Closes #440

### DIFF
--- a/ros_controllers/package.xml
+++ b/ros_controllers/package.xml
@@ -27,7 +27,6 @@
   <exec_depend>joint_state_controller</exec_depend>
   <exec_depend>joint_trajectory_controller</exec_depend>
   <exec_depend>position_controllers</exec_depend>
-  <exec_depend>rqt_joint_trajectory_controller</exec_depend>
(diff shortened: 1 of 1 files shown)


matthew-reynolds:
I also noticed that `four_wheel_steering_controller` is not part of the metapackage. Is this intentional?

bmagyar:
There's a PR polishing that package up which contains that change.

matthew-reynolds:
Ah, got it, I missed that part of #380. Thanks.

bmagyar (approved):

PR #446, opened 2020-01-23
Use #pragma once

matthew-reynolds:
Part of ros-controls/ros_control#403. Corresponds to ros-controls/ros_control#406.

Replace all preprocessor header guards with `#pragma once`. Also adds missing header guard to a few files, including some tests. Every `.h` file now includes `#pragma once`.

(Note: my editor stripped all trailing whitespace in the files I touched. This is particularly apparent in `gripper_action_controller`. Although this change is unrelated to this PR, it doesn't really warrant its own PR, so I left it here since it ought to be dealt with at some point)

### DIFF
--- a/ackermann_steering_controller/include/ackermann_steering_controller/ackermann_steering_controller.h
+++ b/ackermann_steering_controller/include/ackermann_steering_controller/ackermann_steering_controller.h
@@ -36,6 +36,9 @@
  * Author: Masaru Morita, Bence Magyar, Enrique Fernández
  */
 
+#pragma once
+
+
 #include <ackermann_steering_controller/odometry.h>
 #include <controller_interface/controller.h>
 #include <controller_interface/multi_interface_controller.h>
--- a/ackermann_steering_controller/include/ackermann_steering_controller/odometry.h
+++ b/ackermann_steering_controller/include/ackermann_steering_controller/odometry.h
@@ -40,8 +40,8 @@
  * Author: Masaru Morita
  */
 
-#ifndef ODOMETRY_ACKERMANN_STEERING_H_
-#define ODOMETRY_ACKERMANN_STEERING_H_
+#pragma once
+
 
 #include <ros/time.h>
 #include <boost/accumulators/accumulators.hpp>
@@ -206,5 +206,3 @@ namespace ackermann_steering_controller
     IntegrationFunction integrate_fun_;
   };
 }
-
-#endif /* ODOMETRY_ACKERMANN_STEERING_H_ */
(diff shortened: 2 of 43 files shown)


mathias-luedtke:
`#pragma once` is not defined in the standard.
It is supported by most modern compilers, but might not work properly in all cases.
I am note sure, if all compilers for embedded systems support it.

I would rather keep the include guards (and add the missing ones).

One option would be to include both, `#pragma once` and include guards, but I don't think that it is worth the effort.

matthew-reynolds:
>One option would be to include both, #pragma once and include guards, but I don't think that it is worth the effort.

I agree, I don't see much benefit to using both. AFAIK most compilers (at least GCC and clang, I believe) optimize them in a similar way to `#pragma once` anyways. If we want the #ifndef guards, I think we should just remove `#pragma once`.

>I would rather keep the include guards

This is fine by me, but it sounds like this is a difference of opinion between you and @bmagyar. I will wait until you two (and perhaps other maintainers) reach a consensus.

> and add the missing ones

:+1: If we decide against this PR, I will certainly open up another to add those missing guards.

bmagyar:
@ipa-mdl let's find some common ground here. 
I've got this ref: https://en.wikipedia.org/wiki/Pragma_once#Portability
Which particular embedded compiler are we talking about? Most standard ones use GCC by now as far as I know.

mathias-luedtke:
> Which particular embedded compiler are we talking about?

None in particular. This was more of an open question.

> I've got this ref: https://en.wikipedia.org/wiki/Pragma_once#Portability

This only states some level of support. But since there is no standards, the semantics vary.
TBH I don't see the point in replacing the  working include guards for now.

matthew-reynolds:
Let's loop back on this:

`ros_control` has been using `#pragma once` since Jan 3 (ros-controls/ros_control#406). From scrolling through Discourse and ROS Answers, I haven't seen any backlash on this, but of course lots of big companies maintain their own forks, and we know that the vast majority of users wouldn't encounter issues anyways since compiler support is so broad.

But there is of course the important question of what benefit this really brings us, and whether that benefit outweighs any downsides to losing standard-compliance.

Where do we go from here? My only real preference is consistency between our repos, so either this should be merged or ros-controls/ros_control#406 should be reverted.

bmagyar:
Let's go ahead with merging, we can revert both if there's complaints.
Could you please fix the conflict?

matthew-reynolds:
Rebased on latest `melodic-devel`

bmagyar (approved):

PR #447, opened 2020-01-23
Use nullptr

matthew-reynolds:
Part of ros-controls/ros_control#403. Corresponds to ros-controls/ros_control#405.

Use `nullptr` instead of 0 or `NULL` for null pointers.

Note: `ackermann_steering_controller.cpp` and `diff_drive_controller.cpp` contain the following snippet:

```
if (!link)
{
 ROS_ERROR("Link == NULL.");
  return false;
}
```

Here, I did not change the `NULL` to `nullptr` since it's in a string. Would you like this updated as well? "Link == nullptr" or perhaps "Link pointer is null"?

### DIFF
--- a/ackermann_steering_controller/src/ackermann_steering_controller.cpp
+++ b/ackermann_steering_controller/src/ackermann_steering_controller.cpp
@@ -59,7 +59,7 @@ static bool isCylinder(const urdf::LinkConstSharedPtr& link)
 {
   if (!link)
   {
-    ROS_ERROR("Link == NULL.");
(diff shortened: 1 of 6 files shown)


bmagyar:
Let's make the above snippet print a more informative message. The rest of the PR looks good to me!

matthew-reynolds:
Updated to "Link pointer is null." This isn't that much more informative, but additional information is given the surrounding context, and this is at least more human-friendly to read.

bmagyar (approved):

bmagyar (approved):
Thank you!

PR #448, opened 2020-01-27
removed cyclic debug output from ackermann_steering_bot

mathias-luedtke:
this got logged at 100Hz and took >17k lines of test output


mathias-luedtke:
should get cherry-picked to kinetic as well

bmagyar:
Thanks!

PR #450, opened 2020-01-30
WIP: Refactor update function

hslusarek:
Objectives:
* Increase readability
  * create sub-functions
  * better naming of variables
  * better naming of functions
* Remove/Replace unnecessary variables
* Improve separation of verbose output from production code
* Remove unnecessary access to lock-guarded variables

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h
@@ -39,7 +39,7 @@
 
 // Boost
 #include <boost/shared_ptr.hpp>
-#include <boost/dynamic_bitset.hpp>
+#include <boost/optional.hpp>
 
 // ROS
 #include <ros/node_handle.h>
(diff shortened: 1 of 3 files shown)


bmagyar:
Is this still a WIP?
Did you take the review points of @HiddenBug on #442 into consideration?

hslusarek:
> Is this still a WIP?
> Did you take the review points of @HiddenBug on #442 into consideration?

Yes, it's still work in progress because the tests are not running. I think there exists flaky tests on the melodic-devel branch because `sleep` is used in tests which, in my experiences, causes flaky tests. Furthermore, I'm not sure if enough tests existed in the first place to ensure that I don't change the functionality of the `update()` function. We will try to remedy this.
I'm also going to add documentations for the new functions.

bmagyar:
Thanks for the heads up! Yes indeed there are a few flaky tests that we never got around to properly stabilize, I've clicked those retry buttons many times.

martiniil:
> Thanks for the heads up! Yes indeed there are a few flaky tests that we never got around to properly stabilize, I've clicked those retry buttons many times.

Trying to make a progress here: https://github.com/ros-controls/ros_controllers/pull/457

bmagyar:
Is this ready? Since tests should be stabilized now, could you please rebase your branch on top of the latest melodic-devel?

hslusarek:
> Is this ready? Since tests should be stabilized now, could you please rebase your branch on top of the latest melodic-devel?

As you can see a some changes made here are also included in PR  #454 which is the more important PR for us. Therefore, I decided to close this PR.

PR #451, opened 2020-02-04
[melodic] add missing pluginlib deps.

seanyen:
A recent refactoring work done in [`ros_control`](https://github.com/ros-controls/ros_control/pull/404) has removed the extra dependencies to `pluginlib` and which makes the downstream projects, which depends on `pluginlib` but have not yet to be explicit on that, have build breaks when running `catkin_make_isolated`.

This pull request is to be explicitly adding the dependencies on `pluginlib` for the project using plugin model.

### DIFF
--- a/diff_drive_controller/CMakeLists.txt
+++ b/diff_drive_controller/CMakeLists.txt
@@ -14,6 +14,7 @@ find_package(catkin REQUIRED COMPONENTS
     realtime_tools
     tf
     urdf
+    pluginlib
 )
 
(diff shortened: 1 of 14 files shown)


seanyen:
@bmagyar This is ready for review and merge. Thanks!

seanyen:
It looks like the CI does build but there are some test failures. However, those failure are also manifesting on the upstream, so I figure they are safe to ignore?

bmagyar:
@seanyen thank you for spotting this, I am re-running the CI to sort out the flaky tests.

bmagyar (approved):

PR #452, opened 2020-02-05
Added robot namespace to gui of rqt_joint_trajectory_controller

etsiogas:
Mainly useful in multi-robot situations where multiple instances of rqt_joint_trajectory_controller run in different namespaces. Prevents confusion by showing the correct robot namespace to choose from the "controller manager ns" dropdown menu in the controller's window.


bmagyar:
I've tested this with the following:

```
<launch>
  <group ns="hola">
    <node name="rqt_jtc1" type="rqt_joint_trajectory_controller" pkg="rqt_joint_trajectory_controller"/>
  </group>
  <node name="rqt_jtc2" type="rqt_joint_trajectory_controller" pkg="rqt_joint_trajectory_controller"/>
</launch>
```
When using a dummy single robot setup, copy the robot description over to the namespace as well like so:

`$ rosparam get /hola/robot_description`

Proof by screenshot:
Screenshot from 2020-02-21 08 12 47

bmagyar (approved):

Issue #453, opened 2020-02-05
Separate tests into dedicated packages


matthew-reynolds:
I'm proposing splitting out our tests into dedicated packages. These packages will live in the repo, but will not be part of the metapackage, and will not be released. This is similar to [`ros_comm`](https://github.com/ros/ros_comm).

The rationale for this change is to pull out the heavy test dependencies from the metapackage. In particular, the metapackage currently transitively depends on `gazebo_ros`, since it is used by tests in `ackerman_steering_controller`, `diff_drive_controller`, and `four_wheel_steering_controller`.

Would appreciate feedback before I open a PR. Also curious on your thoughts on naming of the test packages (In `ros_control`, our test packages are suffixed with `_tests` but were originally designed with a different intention. In `ros_comm`, they are prefixed `test_`).

bmagyar:
I am happy with either approach, would probably be better to follow `ros_comm`s example. What's your take on this @ipa-mdl?

mathias-luedtke:
IMHO tests belong with the tested code.

The existing test packages have been created to not expose the test plugins.

mathias-luedtke:
> This is similar to ros_comm.

The `ros_comm` test  packages create a lot of test message and service definitions, I guess that its why they were moved out (8 years ago).

> but will not be part of the metapackage, and will not be released

I am okay with removing them from the metapackage (who uses them anyway?).
But we could still release them, in case others want to use the test plugins in their tests.

>  In particular, the metapackage currently transitively depends on gazebo_ros

The released metapackage does not depend on `gazebo_ros`.

matthew-reynolds:
> IMHO tests belong with the tested code

I would agree, but I think the cost of pulling in all these heavy dependencies outweighs the preference of putting the tests alongside the tested code.

> But we could still release them

Good point.

> The released metapackage does not depend on `gazebo_ros`

This is not true - On Melodic, `rosinstall_generator --deps ros_controllers` lists `gazebo_dev`, `gazebo_msgs`, and `gazebo_ros` as recursive dependencies, since they are required by `ackermann_steering_controller`. These can also be seen using `rospack` or `rqt_dep`.

PR #455, opened 2020-02-12
Split tests into dedicated test packages

matthew-reynolds:
Addresses #453

This PR attempts to be as minimal as possible. As much as possible, I only moved code, avoiding changing/adding anything else. The main exception is updating and adding all the required dependencies to the new test packages. Note that, in some cases, these dependencies are somewhat excessive (See follow-up 1) - My goal is to simply match deps to what's currently in the tests, cleanup will come later.

Related follow up PRs I'm working on:
1. Remove unused files from tests. These include several rviz files and "view model" launch files, which don't belong in our test packages and pull in excessive dependencies (`rviz`, `rqt_plot`, and in some cases old Hydro-era packages like `gazebo` and `gazebo_worlds`)
2. Cleanup CMakeLists.txt, package.xml, and *give a thorough pass at dependencies* (Similar to ros-controls/ros_control#404). I gave a thorough pass at deps for the new test packages, but didn't touch existing packages.


test_diff_drive_controller/package.xml
```diff
@@ -0,0 +1,38 @@
...
+  <build_depend>realtime_tools</build_depend>
+  <build_depend>roscpp</build_depend>
+  <build_depend>rosgraph_msgs</build_depend>
+  <build_depend>std_msgs</build_depend>
+  <build_depend>std_srvs</build_depend>
+  <build_depend>tf</build_depend>
+
+  <test_depend>diff_drive_controller</test_depend>
+  <!-- <test_depend>gazebo</test_depend> -->
+  <!-- <test_depend>gazebo_worlds</test_depend> -->
```
matthew-reynolds, line 33: Note that these two deps are commented since they have not existed since Hydro, but are used in `view_diffbot.launch` and `view_skidsteerbot.launch`.

Once those two files are deleted as part of my follow-up cleanup, these deps will be deleted as well. Same goes for all of the excessive `<test_depends>` throughout this PR (`rqt_plot`, `rviz`, etc.) - All will be deleted once the tests using them are deleted in my follow-up.

matthew-reynolds:
This PR also resolves #378

bmagyar:
We have recently merged #463 and #457 which should improve the stability of all unit tests. Sorry for the inconvenience of conflicts on this.

bmagyar:
@ipa-mdl do you have any comments regarding this?

mathias-luedtke:
I would not split the tests..

> Remove unused files from tests. 

If I recall it correctly, these "unused" files are meant to be used for local (non-CI) tests.

Issue #344, opened 2018-05-11
[joint_trajectory_controller] tests fails randomly about 1/3 chance


7675t:
I have noticed the tests for joint_trajectory_controller sometimes fail. The chance of failure is about 1/3, so I believe you can reproduce it to run `catkin run_test` several times locally. The branch is kinetic-devel.

```
$ catkin run_tests joint_trajectory_controller | tee log.txt | grep FAILURES

Warnings << joint_trajectory_controller:make /home/tajima/Codes/ws/logs/
joint_trajectory_controller/build.make.010.log
...............................................................................
 * FAILURES: 0
 * FAILURES: 0
 * FAILURES: 0
 * FAILURES: 0
 * FAILURES: 4
```

And one of the failures is:
```
[joint_trajectory_controller.rosunit-joint_trajectory_controller_vel_test/topicSingleTraj][FAILURE]
/home/tajima/Codes/ws/src/ros_controllers/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp:427
The difference between traj.points.back().positions[i] and state->desired.positions[i] is 0.78539816339744828, which exceeds EPS, where
traj.points.back().positions[i] evaluates to -0.78539816339744828,
state->desired.positions[i] evaluates to 0, and
EPS evaluates to 0.01.
```

@k-okada #336 may fix some of this, but I'm afraid it is not enough.

7675t:
#336 is merged, but I still get failure randomly.

```
[joint_trajectory_controller.rosunit-joint_trajectory_controller_vel_test/emptyTopicCancelsActionTrajWithDelayStopZero][FAILURE]
/home/tajima/Codes/aisinaw_ws/src/ros_controllers/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp:896
Value of: waitForState(action_client, SimpleClientGoalState::SUCCEEDED, long_timeout)
  Actual: false
Expected: true
```

PR #457, opened 2020-02-12
Fix instabilities in JTC unittests

martiniil:
The unittests of the joint_trajectory_controller are highly unstable. See https://github.com/ros-controls/ros_controllers/issues/344 and recent Travis logs.

This PR aims to stabilize the existing test cases in `unittest_joint_trajectory_controller.cpp`.

- Introduced waiting-functions in order to make the tests more deterministic. E.g.
  - wait for start of trajectory execution
  - wait for stop
  - wait for action result
  - wait for (smoothing/delay-)parameter updated.
- Fix how we assert that controller is running (especially after reloading)
- Increased time-limits
- Revised all test cases (especially stop-ramp tests).

### DIFF
--- a/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
+++ b/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
@@ -51,18 +51,16 @@
 #include <controller_manager_msgs/UnloadController.h>
 #include <controller_manager_msgs/SwitchController.h>
 
-// Floating-point value comparison threshold
-static constexpr double JOINT_STATES_COMPARISON_EPS = 0.01;
-static constexpr double EPS = 1e-9;
-
-static constexpr double TIMEOUT_CONNECTIONS_S = 10.0;
-static constexpr double TIMEOUT_ACTION_RESULT_S = 5.0;
-static constexpr double TIMEOUT_TRAJ_EXECUTION_S = 5.0;
+#include "test_common.h"
 
 using actionlib::SimpleClientGoalState;
 using testing::AssertionResult;
 using testing::AssertionFailure;
 using testing::AssertionSuccess;
+
+using namespace joint_trajectory_controller_tests;
+
+static constexpr double TIMEOUT_TRAJ_EXECUTION_S = 5.0;
 
 class JointTrajectoryControllerTest : public ::testing::Test
 {
@@ -165,7 +163,7 @@ public:
    */
   void SetUp() override
   {
-    ASSERT_TRUE(waitForActionServer());
+    ASSERT_TRUE(waitForActionServer(action_client));
     ASSERT_TRUE(waitForInitializedState());
 
     if (!checkPointReached(traj_home.points.back()))
@@ -267,26 +265,6 @@ protected:
     return AssertionSuccess();
   }
 
-  AssertionResult waitForActionServer(const ros::Duration& timeout = ros::Duration(TIMEOUT_CONNECTIONS_S))
-  {
-    if (!action_client->waitForServer(timeout))
-    {
-      return AssertionFailure() << "Timed out after " << timeout.toSec()
-                                << "s waiting for connection to action server.";
-    }
-    return AssertionSuccess();
-  }
-
-  AssertionResult waitForActionResult(const ActionClientPtr& action_client,
-                                      const ros::Duration& timeout = ros::Duration(TIMEOUT_ACTION_RESULT_S))
-  {
-    if (!action_client->waitForResult(timeout))
-    {
-      return AssertionFailure() << "Timed out after " << timeout.toSec() << "s waiting for action result.";
-    }
-    return AssertionSuccess();
-  }
-
   std::vector<double> getMinActualVelocity()
   {
     std::lock_guard<std::mutex> lock(mutex);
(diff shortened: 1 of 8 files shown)


joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -49,17 +52,23 @@
 #include <controller_manager_msgs/SwitchController.h>
 
 // Floating-point value comparison threshold
-const double EPS = 0.01;
+static constexpr double JOINT_STATES_COMPARISON_EPS = 0.01;
+static constexpr double EPS = 1e-9;
+
+static constexpr double TIMEOUT_CONNECTIONS_S  {10.0};
```
bmagyar: Could you make these consistent please? Use the equal operator and curly braces style.

`i = {2};`

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -102,53 +110,83 @@ class JointTrajectoryControllerTest : public ::testing::Test
...
+
     // Query state service client
-    query_state_service = nh.serviceClient<control_msgs::QueryTrajectoryState>("query_state");
+    query_state_service = controller_nh.serviceClient<control_msgs::QueryTrajectoryState>("query_state");
 
     // Controller management services
-    load_controller_service = nh.serviceClient<controller_manager_msgs::LoadController>("/controller_manager/load_controller");
-    unload_controller_service = nh.serviceClient<controller_manager_msgs::UnloadController>("/controller_manager/unload_controller");
-    switch_controller_service = nh.serviceClient<controller_manager_msgs::SwitchController>("/controller_manager/switch_controller");
+    {
```
bmagyar, line 141: Why the scope?

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -102,53 +110,83 @@ class JointTrajectoryControllerTest : public ::testing::Test
...
+  {
+    ASSERT_TRUE(waitForActionServer());
+    ASSERT_TRUE(waitForInitializedState());
+
+    if (!checkPointReached(traj_home.points.back()))
+    {
+      traj_home_goal.trajectory.header.stamp = ros::Time(0);
+      action_client->sendGoal(traj_home_goal);
+
+      auto timeout{getTrajectoryDuration(traj_home_goal.trajectory) + ros::Duration(TIMEOUT_ACTION_RESULT_S)};
```
bmagyar: = please

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -189,79 +231,109 @@ class JointTrajectoryControllerTest : public ::testing::Test
...
-    std::transform(controller_max_actual_velocity.begin(), controller_max_actual_velocity.end(),
-                   state->actual.velocities.begin(), controller_max_actual_velocity.begin(),
-                   [](double a, double b){return std::max(a, b);});
   }
 
   StateConstPtr getState()
   {
     std::lock_guard<std::mutex> lock(mutex);
-    return controller_state;
+    return boost::make_shared<const control_msgs::JointTrajectoryControllerState>(*controller_state);
```
bmagyar: If returning a shared ptr doesn't work well but you want a copy instead, let's return by value.

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -289,14 +361,82 @@ class JointTrajectoryControllerTest : public ::testing::Test
...
+      ros::Duration(0.1).sleep();
+      if (count == repeat)
+      {
+        return AssertionSuccess();
+      }
+    }
+    return AssertionFailure() << "ROS shutdown.";
+  }
+
+  static bool trajectoryPointsAlmostEqual(const trajectory_msgs::JointTrajectoryPoint& p1,
```
bmagyar: If you didn't make some vars static above, this could be a non static function. Can we do without statics please unless they actually bring a benefit?

bmagyar (changes requested):
This is only half of my review

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: To limit the scope of the following `using` statement.

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: Done.

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: How about making them global functions? Maybe some of these static functions could also be used in other tests. So it would even make sense to put them into a separate header?

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: Done.

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: Ok, in order to be consistent I use initialization as follows: `int i = 2;`

bmagyar: thanks

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
bmagyar: :+1:

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
bmagyar: love the idea!

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -414,117 +558,104 @@ TEST_F(JointTrajectoryControllerTest, invalidMessages)
...
-  ASSERT_TRUE(initState());
-
   // Send trajectory
   traj.header.stamp = ros::Time(0); // Start immediately
   traj_pub.publish(traj);
-  ros::Duration wait_duration = traj.points.back().time_from_start + ros::Duration(0.5);
-  wait_duration.sleep(); // Wait until done
+  ASSERT_TRUE(waitForTrajectoryExecution());
+
+  auto timeout{getTrajectoryDuration(traj) + ros::Duration(TIMEOUT_TRAJ_EXECUTION_S)};
```
bmagyar: {}{}

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
```diff
@@ -414,117 +558,104 @@ TEST_F(JointTrajectoryControllerTest, invalidMessages)
...
   reorder_goal.trajectory.points.front().positions[0] = M_PI / 4.0;
   reorder_goal.trajectory.points.front().positions[1] = 0.0;
 
-
   reorder_goal.trajectory.header.stamp = ros::Time(0); // Start immediately
   action_client->sendGoal(reorder_goal);
-  ASSERT_TRUE(waitForState(action_client, SimpleClientGoalState::SUCCEEDED, long_timeout));
-  ros::Duration(0.5).sleep(); // Allows values to settle to within EPS, especially accelerations
+
+  auto timeout{getTrajectoryDuration(reorder_goal.trajectory) + ros::Duration(TIMEOUT_ACTION_RESULT_S)};
```
bmagyar: I think you may be overusing {} a little bit :) 
```suggestion
  const auto timeout = getTrajectoryDuration(reorder_goal.trajectory) + ros::Duration(TIMEOUT_ACTION_RESULT_S);
```
is just so much more readable and there's not loss, `{}` is generally recommended for member variables, while this is local *and* in a test so I wouldn't push to beautify

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
bmagyar: makes sense, thanks!

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: Yep, I prepared a commit changing this already

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: Done.

joint_trajectory_controller/test/joint_trajectory_controller_test.cpp
martiniil: Thanks! I will push another commit.

bmagyar (approved):

bmagyar (approved):
Thank you so much for this, this is great!

PR #458, opened 2020-02-12
[rqt joint trajectory controller] Python3 fixes

gleichdick:
Two small fixes to make it Python 3 ready:
- Use explicit relative import (with leading dot). Implicit relative imports [are gone in Python 3.](https://docs.python.org/release/3.0.1/whatsnew/3.0.html#removed-syntax)
- `print` function with parentheses (Syntax Error otherwise)


bmagyar (approved):
Tested, sweet! Thank you for this

Issue #365, opened 2018-10-19
JointPositionController oscillation


skrede:
I've been trying to get a simulation of the Kuka KR6 Agilus running in Gazebo, and while almost there, there is one problem that remains.

I'm running Gazebo 9 with ROS Melodic. 

Each joint is controlled by a JointPositionController, and all but one joint behaves sufficiently. I won't paste configurations here, but the repository is available on Github: https://github.com/skrede/kuka_gazebo

The joint in question is joint_a4, and the problem is that the joint has a given setpoint but somehow the controller seems confused as to how to get there. If I echo the following topic (of message type control_msgs/JointControllerState)

> rostopic echo /joint_a4_position_controller/state

to terminal, I can see that the setpoint is specified, but the command and error fields seem to perturb violently, as shown in the following printout of the topic. 

    ---
    header: 
      seq: 74830
      stamp: 
        secs: 748
        nsecs: 825000000
      frame_id: ''
    set_point: -1.69285065965e-08
    process_value: -3.05827630699
    process_value_dot: 1.96346264844
    error: 3.05827629006
    time_step: 0.001
    command: 1175.71498185
    p: 500.0
    i: 5.0
    d: 180.0
    i_clamp: 0.0
    antiwindup: False
    ---
    header: 
      seq: 74831
      stamp: 
        secs: 748
        nsecs: 835000000
      frame_id: ''
    set_point: -1.69285065965e-08
    process_value: -3.03920837904
    process_value_dot: 1.76346476062
    error: -3.24397694507
    time_step: 0.001
    command: -1939.41204719
    p: 500.0
    i: 5.0
    d: 180.0
    i_clamp: 0.0
    antiwindup: False

This might not be the correct place to raise the issue, but I'm not sure where the problem lies. I suppose it could be the URDF or some other configuration that confuse the controller, but it seemed to work just fine using `position_controllers/JointTrajectoryController`. This problem occured after using JointPositionController. 
I started looking at the urdf configurations, but I'm not even sure what to look for. I also tried to adjust PID gains, but that didn't help either. 

A video of the problem is available here: https://youtu.be/PqQh9spXjgM

The package is somewhat easy to get running, given dependencies are already installed. I haven't listed them in the repository, but the package.xml should state dependencies.

When running Gazebo, the joint starts to perturb right away. This can be seen visually or by echoing the abovementioned topic to terminal. The setpoint does not change, i.e., nothing is sent via the topic  

> /joint_a4_position_controller/command

[...]

PR #466, opened 2020-02-28
Solving issues with large rotational limits

francofusco:
As found out by others as well, there is a small issue in some controllers when a revolute joints has "large" limits, see eg #365 

The problem is that some position controllers call the function `angles::shortest_angular_distance_with_limits`, which assumes that the total allowed range of rotation is lower than `2*PI`. Therefore, if one has as limits such as `[-PI-eps, PI+eps)`, `angles` will assume that the valid range has size `2*eps` instead of `2*(PI+eps)`. To allow dealing with larger rotations, the PR ros/angles#16 introduced the function `shortest_angular_distance_with_large_limits`.

I updated three controllers to use such function in case `upper_limit-lower_limit` is greater than or equal to `2*PI`. I tested it and the oscillations that could be experienced before now disappear.

### DIFF
--- a/effort_controllers/CMakeLists.txt
+++ b/effort_controllers/CMakeLists.txt
@@ -57,3 +57,18 @@ install(TARGETS ${PROJECT_NAME}
 
 install(FILES effort_controllers_plugins.xml
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+
+if(CATKIN_ENABLE_TESTING)
+  find_package(catkin REQUIRED COMPONENTS controller_manager rostest)
+  include_directories(include ${catkin_INCLUDE_DIRS})
+
+  add_executable(simple_bot_hw test/simple_bot_hw.cpp)
+  target_link_libraries(simple_bot_hw ${catkin_LIBRARIES})
+  add_dependencies(tests simple_bot_hw)
+
+  add_rostest_gtest(effort_position_controller_test
+    test/effort_position_controller.test
+    test/simple_bot_goto_position.cpp
+  )
+  target_link_libraries(effort_position_controller_test ${catkin_LIBRARIES})
+endif()
--- a/effort_controllers/CMakeLists.txt
+++ b/effort_controllers/CMakeLists.txt
@@ -59,16 +59,35 @@ install(FILES effort_controllers_plugins.xml
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 
 if(CATKIN_ENABLE_TESTING)
-  find_package(catkin REQUIRED COMPONENTS controller_manager rosgraph_msgs rostest)
-  include_directories(include ${catkin_INCLUDE_DIRS})
+  find_package(rostest REQUIRED)
+  find_package(controller_manager REQUIRED)
+  find_package(controller_manager_msgs REQUIRED)
+  find_package(rosgraph_msgs REQUIRED)
+  find_package(sensor_msgs REQUIRED)
+  find_package(std_msgs REQUIRED)
+
+  include_directories(
+    ${controller_manager_INCLUDE_DIRS}
+    ${hardware_interface_INCLUDE_DIRS}
+    ${rosgraph_msgs_INCLUDE_DIRS}
+    ${sensor_msgs_INCLUDE_DIRS}
+    ${std_msgs_INCLUDE_DIRS}
+  )
 
   add_executable(simple_bot_hw test/simple_bot_hw.cpp)
-  target_link_libraries(simple_bot_hw ${catkin_LIBRARIES})
+  target_link_libraries(simple_bot_hw
+    ${catkin_LIBRARIES}
+    ${controller_manager_LIBRARIES}
+    ${hardware_interface_LIBRARIES}
+  )
   add_dependencies(tests simple_bot_hw)
 
   add_rostest_gtest(effort_position_controller_test
     test/effort_position_controller.test
     test/simple_bot_goto_position.cpp
   )
-  target_link_libraries(effort_position_controller_test ${catkin_LIBRARIES})
+  target_link_libraries(
+    effort_position_controller_test
+    ${catkin_LIBRARIES}
+  )
 endif()
--- a/effort_controllers/CMakeLists.txt
+++ b/effort_controllers/CMakeLists.txt
@@ -59,7 +59,7 @@ install(FILES effort_controllers_plugins.xml
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 
 if(CATKIN_ENABLE_TESTING)
-  find_package(catkin REQUIRED COMPONENTS controller_manager rostest)
+  find_package(catkin REQUIRED COMPONENTS controller_manager rosgraph_msgs rostest)
   include_directories(include ${catkin_INCLUDE_DIRS})
 
   add_executable(simple_bot_hw test/simple_bot_hw.cpp)
--- a/effort_controllers/package.xml
+++ b/effort_controllers/package.xml
@@ -28,6 +28,7 @@
   <test_depend>rostest</test_depend>
   <test_depend>controller_manager</test_depend>
   <test_depend>xacro</test_depend>
+  <test_depend>robot_state_publisher</test_depend>
 
   <export>
     <controller_interface plugin="${prefix}/effort_controllers_plugins.xml"/>
--- a/effort_controllers/package.xml
+++ b/effort_controllers/package.xml
@@ -26,10 +26,13 @@
   <depend>pluginlib</depend>
 
   <test_depend>rostest</test_depend>
-  <test_depend>rosgraph_msgs</test_depend>
   <test_depend>controller_manager</test_depend>
-  <test_depend>xacro</test_depend>
+  <test_depend>hardware_interface</test_depend>
   <test_depend>robot_state_publisher</test_depend>
+  <test_depend>rosgraph_msgs</test_depend>
+  <test_depend>sensor_msgs</test_depend>
+  <test_depend>std_msgs</test_depend>
+  <test_depend>xacro</test_depend>
 
   <export>
     <controller_interface plugin="${prefix}/effort_controllers_plugins.xml"/>
--- a/effort_controllers/package.xml
+++ b/effort_controllers/package.xml
@@ -17,13 +17,17 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <depend>angles</depend>
-  <depend>controller_interface</depend> 
-  <depend>control_msgs</depend> 
-  <depend>control_toolbox</depend> 
-  <depend>realtime_tools</depend> 
-  <depend>urdf</depend> 
+  <depend>controller_interface</depend>
+  <depend>control_msgs</depend>
+  <depend>control_toolbox</depend>
+  <depend>realtime_tools</depend>
+  <depend>urdf</depend>
   <depend>forward_command_controller</depend>
   <depend>pluginlib</depend>
+
+  <test_depend>rostest</test_depend>
+  <test_depend>controller_manager</test_depend>
+  <test_depend>xacro</test_depend>
 
   <export>
     <controller_interface plugin="${prefix}/effort_controllers_plugins.xml"/>
--- a/effort_controllers/package.xml
+++ b/effort_controllers/package.xml
@@ -26,6 +26,7 @@
   <depend>pluginlib</depend>
 
   <test_depend>rostest</test_depend>
+  <test_depend>rosgraph_msgs</test_depend>
   <test_depend>controller_manager</test_depend>
   <test_depend>xacro</test_depend>
   <test_depend>robot_state_publisher</test_depend>
--- a/effort_controllers/src/joint_group_position_controller.cpp
+++ b/effort_controllers/src/joint_group_position_controller.cpp
@@ -138,12 +138,24 @@ namespace effort_controllers
         // Compute position error
         if (joint_urdfs_[i]->type == urdf::Joint::REVOLUTE)
         {
-         angles::shortest_angular_distance_with_limits(
-            current_position,
-            command_position,
-            joint_urdfs_[i]->limits->lower,
-            joint_urdfs_[i]->limits->upper,
-            error);
+          if(joint_urdfs_[i]->limits->upper - joint_urdfs_[i]->limits->lower >= 2*M_PI)
+          {
+            angles::shortest_angular_distance_with_large_limits(
+              current_position,
+              command_position,
+              joint_urdfs_[i]->limits->lower,
+              joint_urdfs_[i]->limits->upper,
+              error);
+          }
+          else
+          {
+            angles::shortest_angular_distance_with_limits(
+              current_position,
+              command_position,
+              joint_urdfs_[i]->limits->lower,
+              joint_urdfs_[i]->limits->upper,
+              error);
+          }
         }
         else if (joint_urdfs_[i]->type == urdf::Joint::CONTINUOUS)
         {
@@ -165,9 +177,9 @@ namespace effort_controllers
   void JointGroupPositionController::commandCB(const std_msgs::Float64MultiArrayConstPtr& msg)
   {
     if(msg->data.size()!=n_joints_)
-    { 
+    {
       ROS_ERROR_STREAM("Dimension of command (" << msg->data.size() << ") does not match number of joints (" << n_joints_ << ")! Not executing!");
-      return; 
+      return;
     }
     commands_buffer_.writeFromNonRT(msg->data);
   }
--- a/effort_controllers/src/joint_group_position_controller.cpp
+++ b/effort_controllers/src/joint_group_position_controller.cpp
@@ -138,24 +138,12 @@ namespace effort_controllers
         // Compute position error
         if (joint_urdfs_[i]->type == urdf::Joint::REVOLUTE)
         {
-          if(joint_urdfs_[i]->limits->upper - joint_urdfs_[i]->limits->lower >= 2*M_PI)
-          {
-            angles::shortest_angular_distance_with_large_limits(
-              current_position,
-              command_position,
-              joint_urdfs_[i]->limits->lower,
-              joint_urdfs_[i]->limits->upper,
-              error);
-          }
-          else
-          {
-            angles::shortest_angular_distance_with_limits(
-              current_position,
-              command_position,
-              joint_urdfs_[i]->limits->lower,
-              joint_urdfs_[i]->limits->upper,
-              error);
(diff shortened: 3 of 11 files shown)


bmagyar (changes requested):
The changes look good to me on first read. Could you please add a test that fails without these changes? Could reuse the same code for all controllers.

francofusco:
Glad to! However, how should I proceed? The issue can be seen "visually" without much problems, however I am not so sure about what strategy to follow to analyze it in a test - I believe you want a test that can be checked by continuous integration, isn't it? I think the easier way to spot the problem in the code is to check the return value of `angles::shortest_angular_distance_with_limits` (it will likely return `false` at some point), but I am not sure that it will always do so... I will think about it and try to come up with a satisfactory solution!

effort_controllers/src/joint_group_position_controller.cpp
```diff
@@ -138,12 +138,24 @@ namespace effort_controllers
...
-            error);
+          if(joint_urdfs_[i]->limits->upper - joint_urdfs_[i]->limits->lower >= 2*M_PI)
+          {
+            angles::shortest_angular_distance_with_large_limits(
+              current_position,
+              command_position,
+              joint_urdfs_[i]->limits->lower,
+              joint_urdfs_[i]->limits->upper,
+              error);
+          }
```
matthew-reynolds: Is it necessary to perform this check? Since `ros_control` would have several issues if `lower < upper` is violated, I think it is safe to assume that `lower` will always be less than `upper`. As such, wouldn't it be equivalent to just always use the new `shortest_angular_distance_with_large_limits`?

matthew-reynolds (commented):
Looks good to me, I agree that tests would be great. Perhaps we could check the sign of the commanded vel/effort for a specific set of limits and current position to see if it is using the correct limits? I don't know what angle `shortest_angular_distance_with_limits` returns if the limits are violated, so maybe this wouldn't work.

effort_controllers/src/joint_group_position_controller.cpp
francofusco: No, I don't think it is necessary - I left it since I was not entirely sure if at any point one could have `lower > upper`. I will remove it!

bmagyar:
Ping

francofusco:
Sorry for being inactive for such a long time.

I removed the check `lower < upper`, I will now work on a test to make sure things work as intended.

francofusco:
I added a simple test for `effort_controllers::JointPositionController`: it involves a simple robot with a single revolute joint with limits `[-6.3, 6.3]`. I created a supersimple `RobotHW` to fake a real robot, and the test simply tries to move the single joint to the positions `1.0`, `-1.0` and `0.0` in succession. Without the fix, the joint oscillates around the initial position and thus never reaches the targets - teh tset should fail due to the 50s timeout set in `[effort_position_controller.test](effort_controllers/test/effort_position_controller.test)`. With the fix, the position is reached and the test should run just fine. To "see" the test, you can launch the test manually:
```
roslaunch effort_controllers simple_bot_hw.launch
rosrun effort_controllers effort_position_controller_test
```
If you start RViz, set the fixed frame as `world` and add a `RobotModel` plugin you should be able to see the robot (nothing more than a rotating beam). Recompile with and without the fix to see the difference. 

Is this good enough? The test is quite simple, but I could not think of any better solution... If you think it is ok, I can then add a similar test also for `velocity_controllers/JointPositionController`.

Note: the test is failing during the CI due to the timeout. Since on my PC it was working well, I think it might be due to a lower processing power of the hosting virtual machine (?). I will try to solve the issue by increasing both the control period (it is currently 1ms) and the timeout. Sorry if this adds unnecessary commits...

francofusco:
I was finally able to make the tests succeed. Again, I apologize for the long commit list... I can rebase and squash all test-related commits into one, let me know!

bmagyar (approved):

francofusco:
Thanks for the approval! Before merging, do you want me to rebase the commit history and/or add the same test also for `velocity_controllers/JointPositionController`?

effort_controllers/test/simple_bot_goto_position.cpp
```diff
@@ -0,0 +1,70 @@
+#include <ros/ros.h>
```
matthew-reynolds: Missing licence comment

effort_controllers/test/simple_bot_hw.cpp
```diff
@@ -0,0 +1,120 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+// ROS
+#include <chrono>
+#include <thread>
+#include <ros/ros.h>
```
matthew-reynolds: Nitpick: `chrono` and `thread` aren't ROS packages.

effort_controllers/test/simple_bot_hw.cpp
```diff
@@ -0,0 +1,120 @@
...
+  void read() {
+    // update current state
+    pos_ += period * vel_ + 0.5*std::pow(period, 2.) * eff_;
+    vel_ += period * eff_;
+  }
+
+  void write() {
+    // store current command
+    eff_ = cmd_;
+    if(std::fabs(eff_) > max_eff_) {
```
matthew-reynolds, line 83: Let's `#include <cmath>` for this `fabs` and the `pow` above.

effort_controllers/test/simple_bot_hw.cpp
```diff
@@ -0,0 +1,120 @@
...
+      ROS_WARN("Large effort command received: %f; cutting down to % f", eff_, max_eff_);
+      eff_ = eff_ > 0 ? max_eff_ : -max_eff_;
```
matthew-reynolds: Nitpick:
```suggestion
      eff_ = eff_ > 0.0 ? max_eff_ : -max_eff_;
```

effort_controllers/test/simple_bot_hw.cpp
```diff
@@ -0,0 +1,120 @@
...
+  ros::init(argc, argv, "simple_bot_hw");
+  ros::NodeHandle nh;
+  SimpleBot bot;
+  ros::Publisher clock_publisher = nh.advertise<rosgraph_msgs::Clock>("/clock", 1);
+  nh.setParam("/use_sim_time", true);
+  controller_manager::ControllerManager cm(&bot, nh);
+  ros::AsyncSpinner spinner(2);
+  spinner.start();
+
+  ros::Rate rate(1./bot.period);
```
matthew-reynolds: Nitpick: I don't think we use `1.` anywhere in this codebase, do you mind switching to `1.0`? Same above at line 65.

effort_controllers/test/simple_bot_hw.cpp
```diff
@@ -0,0 +1,120 @@
...
+    cm.update(ros::Time(sim_time), period);
+    bot.write();
+    auto end = std::chrono::system_clock::now();
+    double elapsed = std::chrono::duration_cast<std::chrono::duration<double> >((end - begin)).count();
+    if(bot.period > elapsed)
+      std::this_thread::sleep_for(std::chrono::duration<double>(bot.period - elapsed));
+    rosgraph_msgs::Clock clock;
+    clock.clock = ros::Time(sim_time);
+    clock_publisher.publish(clock);
+    sim_time += bot.period;
```
matthew-reynolds, line 125: Just curious, why are you managing the sim time yourself rather than just using the ROS clock?

matthew-reynolds: Oh, I see you're just matching what the other tests do. Must be a good reason for it that I just haven't thought of :)

effort_controllers/CMakeLists.txt
```diff
@@ -57,3 +57,18 @@ install(TARGETS ${PROJECT_NAME}
...
+
+  add_executable(simple_bot_hw test/simple_bot_hw.cpp)
+  target_link_libraries(simple_bot_hw ${catkin_LIBRARIES})
+  add_dependencies(tests simple_bot_hw)
+
+  add_rostest_gtest(effort_position_controller_test
+    test/effort_position_controller.test
+    test/simple_bot_goto_position.cpp
+  )
+  target_link_libraries(effort_position_controller_test ${catkin_LIBRARIES})
```
matthew-reynolds: Careful here, when you `find_package(catkin REQUIRED COMPONENTS ...)` it wipes out `catkin_INCLUDE_DIRS` and `catkin_LIBRARIES` so we lose everything that was imported by the first `find_package` above. We're also missing a few packages.

You could either save a copy of those vars before you `find_package`, or list each package individually like this:

```suggestion
  find_package(rostest REQUIRED)
  
  find_package(controller_manager REQUIRED)
  find_package(hardware_interface REQUIRED)
  find_package(rosgraph_msgs REQUIRED)
  find_package(sensor_msgs REQUIRED)
  find_package(std_msgs REQUIRED)

  include_directories(
    ${controller_manager_INCLUDE_DIRS}
    ${hardware_interface_INCLUDE_DIRS}
    ${rosgraph_msgs_INCLUDE_DIRS}
    ${sensor_msgs_INCLUDE_DIRS)
    ${std_msgs_INCLUDE_DIRS}
  )

  add_executable(simple_bot_hw test/simple_bot_hw.cpp)
  target_link_libraries(simple_bot_hw
    ${catkin_LIBRARIES}
    ${controller_manager_LIBRARIES}
    ${hardware_interface_LIBRARIES}
  )
  add_dependencies(tests simple_bot_hw)

  add_rostest_gtest(effort_position_controller_test
    test/effort_position_controller.test
    test/simple_bot_goto_position.cpp
  )
  target_link_libraries(effort_position_controller_test ${catkin_LIBRARIES})
```

With either approach, please add the missing `hardware_interface`, `sensor_msgs`, and `std_msgs` dependencies, and also pull `find_package(rostest REQUIRED)` out to be its own line, at the beginning of the `if(CATKIN_ENABLE_TESTING)` block.

effort_controllers/package.xml
```diff
@@ -17,14 +17,20 @@
...
+  <depend>realtime_tools</depend>
+  <depend>urdf</depend>
   <depend>forward_command_controller</depend>
   <depend>pluginlib</depend>
 
+  <test_depend>rostest</test_depend>
+  <test_depend>rosgraph_msgs</test_depend>
+  <test_depend>controller_manager</test_depend>
+  <test_depend>xacro</test_depend>
+  <test_depend>robot_state_publisher</test_depend>
```
matthew-reynolds: Please alphabetize and add additional depends as described above.

effort_controllers/test/simple_bot_goto_position.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+
+class SimpleBotFixture : public ::testing::Test {
+protected:
+  ros::Publisher pub;
+  ros::Subscriber sub;
+  ros::NodeHandle nh;
+  double position;
+
+  int val;
```
matthew-reynolds: This looks unused, let's remove it.

effort_controllers/test/simple_bot_goto_position.cpp
```diff
@@ -0,0 +1,70 @@
...
+  }
+};
+
+
+TEST_F(SimpleBotFixture, TestPosition) {
+  ros::Duration(1.0).sleep();
+  std::string err;
+  ASSERT_TRUE(waitToReachTarget(1.0, err)) << err;
+  ASSERT_TRUE(waitToReachTarget(-1.0, err)) << err;
+  ASSERT_TRUE(waitToReachTarget(0.0, err)) << err;
```
matthew-reynolds: I think these should be `EXPECT`, not `ASSERT`.

effort_controllers/test/simple_bot_hw.cpp
```diff
@@ -0,0 +1,120 @@
...
+class SimpleBot : public hardware_interface::RobotHW
+{
+public:
+  static constexpr double period = 0.01;
+
+  SimpleBot()
+  : pos_(0)
+  , vel_(0)
+  , eff_(0)
+  , max_eff_(10)
```
matthew-reynolds: Nitpick: Can we add `.0` to all of these?

matthew-reynolds (changes requested):
This looks great, thanks! I think we just need to touch up the CMakeLists & `package.xml` to get the dependencies sorted out, and then I have a handful of small requests.

Can you confirm that these tests fail with the old `angles::shortest_angular_distance_with_limits`?

bmagyar (approved):
I'm happy with these, can merge as soon as @matthew-reynolds is happy too :)

effort_controllers/test/simple_bot_goto_position.cpp
francofusco: I added the license and I also updated the one of `simple_bot_hw.cpp`. They are now a copy of the one from `joint_position_controller.cpp` with my name added. I am never 100% sure about licenses, so feel free to correct me if something is wrong with this.

effort_controllers/test/simple_bot_hw.cpp
francofusco: Absolutely right, I moved them up, adding the `#include <cmath>` there as well.

effort_controllers/test/simple_bot_hw.cpp
francofusco: See [comment on `#include <chrono>` and `#include <thread>`](https://github.com/ros-controls/ros_controllers/pull/466#discussion_r406656321).

effort_controllers/test/simple_bot_hw.cpp
francofusco: I updated all doubles to be in the format `XX.0`.

effort_controllers/test/simple_bot_hw.cpp
francofusco: Sorry, taken care of - see [other comment](https://github.com/ros-controls/ros_controllers/pull/466#discussion_r406657023)

effort_controllers/CMakeLists.txt
francofusco: Thanks for pointing this out, I was actually not aware of that! I chose the second approach, so every package is found separately wih its own `find_package`. I sorted the entries alphabetically - while leaving `rostest` on top of everything.

effort_controllers/test/simple_bot_goto_position.cpp
francofusco: Totally right.

effort_controllers/test/simple_bot_goto_position.cpp
francofusco: As far as I understand, the only difference between `ASSERT` and `EXPECT` is that the former will stop the test right away, while the latter will record the failure but let the test finish. As there is no cleanup needed in the test, if one of the three `waitToReachTarget` calls fail, we can already exit. This is why I used `ASSERT` rather than `EXPECT`, but if there is any reason to change it, I will!

effort_controllers/test/simple_bot_hw.cpp
francofusco: [Done](https://github.com/ros-controls/ros_controllers/pull/466#discussion_r406657023).

francofusco:
@matthew-reynolds I addressed your comments, let me know if anything needs to be updated :smiley: 

> Can you confirm that these tests fail with the old `angles::shortest_angular_distance_with_limits`?

Yes, if I change the `angles` call to the old one, the joint does not rotate (actually, it keeps oscillating around zero).

*Update*: I uploaded two videos, showing the difference between the [proposed changes](https://youtu.be/Z9KjIDEziaM) and the [old behavior](https://youtu.be/Wdn7o0pLc80).

francofusco:
Just found a stupid mistake in the `simple_bot_hw.cpp`: the variable `cmd_` was not initialized in the constructor, I believe it was causing troubles in the tests...

effort_controllers/test/simple_bot_goto_position.cpp
matthew-reynolds: Yep, my only thinking was that `EXPECT` might help if there's any flakiness. If 0.1% of the time a test fails, if we use `EXPECT` we can see "Oh, 2/3 passed, probably just need to re-run the test" rather than thinking the changes broke the test entirely.

effort_controllers/test/simple_bot_goto_position.cpp
matthew-reynolds: I'm never sure either :)

@bmagyar look good?

matthew-reynolds (approved):
Awesome, thanks for the videos, this is great.

I'm good as soon as @bmagyar gives the license a peek, and once we swap to `EXPECT`.

effort_controllers/CMakeLists.txt
matthew-reynolds: Yep, it's on my TODO list to fix up the other packages that have `find_package(catkin ...` twice as well. Seems like a common problem, it's a shame `find_package(catkin ...` doesn't have an option to append to the lists rather than overwrite. Thanks for fixing it!

effort_controllers/test/simple_bot_goto_position.cpp
bmagyar: TBH I wouldn't be too fussed about putting license notes in every file, especially since the package.xml states BSD3. If you'd like though, that's ok but don't attribute new files to Willow Garage or HiDof, if it's just you, put yourself.

effort_controllers/test/simple_bot_goto_position.cpp
francofusco: I get it, makes sense! I will change them accordingly

effort_controllers/test/simple_bot_hw.cpp
francofusco: To be honest, I decided to put it because in the first versions of the test I was getting no errors on my PC but checks were failing during CI. I imagined that one issue was related to the virtual machine: perhaps it had limited processing power and thus I wanted to "slow down" things accordingly at least in the RobotHW. I saw that other packages were using the simulated clock and I thus used the same procedure. Now that I found I forgot to initialize the `cmd_` variable, I think that was the main source of issues 😅 But I think it might still be useful, in case the VM slows down...

francofusco:
I updated the code according to the last comments. I think all points from your reviews have been addressed, but let me know if I forgot something instead!

bmagyar:
Thanks very much for sticking with us through this review! Merging...

PR #468, opened 2020-03-10
Add code_coverage target to CMakeLists.txt

martiniil:
Add option `ENABLE_COVERAGE_TESTING` for running coverage analysis of package joint_trajectory_controller. See https://github.com/mikeferguson/code_coverage

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -42,6 +42,11 @@ catkin_package(
 
 include_directories(include ${Boost_INCLUDE_DIR} ${catkin_INCLUDE_DIRS})
 
+if(CATKIN_ENABLE_TESTING AND ENABLE_COVERAGE_TESTING)
+  find_package(code_coverage REQUIRED)
+  APPEND_COVERAGE_COMPILER_FLAGS()
+endif()
+
 add_library(${PROJECT_NAME} src/joint_trajectory_controller.cpp
                             include/joint_trajectory_controller/hardware_interface_adapter.h
                             include/joint_trajectory_controller/init_joint_trajectory.h
@@ -120,6 +125,11 @@ if(CATKIN_ENABLE_TESTING)
                     test/joint_trajectory_controller_wrapping.test
                     test/joint_trajectory_controller_wrapping_test.cpp)
   target_link_libraries(joint_trajectory_controller_wrapping_test ${catkin_LIBRARIES})
+
+  if(ENABLE_COVERAGE_TESTING)
+    set(COVERAGE_EXCLUDES "*/${PROJECT_NAME}/test*")
+    add_code_coverage(NAME ${PROJECT_NAME}_coverage)
+  endif()
 
 endif()
 
--- a/joint_trajectory_controller/package.xml
+++ b/joint_trajectory_controller/package.xml
@@ -34,6 +34,7 @@
   <test_depend>rostest</test_depend>
   <test_depend>controller_manager</test_depend>
   <test_depend>xacro</test_depend>
+  <test_depend>code_coverage</test_depend>
 
   <export>
     <controller_interface plugin="${prefix}/ros_control_plugins.xml"/>


joint_trajectory_controller/CMakeLists.txt
```diff
@@ -42,6 +42,11 @@ catkin_package(
 
 include_directories(include ${Boost_INCLUDE_DIR} ${catkin_INCLUDE_DIRS})
 
+if(ENABLE_COVERAGE_TESTING AND ENABLE_COVERAGE_TESTING)
```
bmagyar: You want to be double sure? ;)

bmagyar (commented):
I'd be very happy if we could enable this by default on Travis. I believe @matthew-reynolds looked into this before

joint_trajectory_controller/CMakeLists.txt
martiniil: Thanks, I fixed it.

joint_trajectory_controller/CMakeLists.txt
```diff
@@ -42,6 +42,11 @@ catkin_package(
 
 include_directories(include ${Boost_INCLUDE_DIR} ${catkin_INCLUDE_DIRS})
 
+if(CATKIN_ENABLE_TESTING AND ENABLE_COVERAGE_TESTING)
+  find_package(code_coverage REQUIRED)
+  APPEND_COVERAGE_COMPILER_FLAGS()
+endif()
```
matthew-reynolds, line 48: Any reason we have two separate `if`s? I think we should move the `find_package` and `APPEND_COVERAGE_COMPILER_FLAGS` into the other `if(ENABLE_COVERAGE_TESTING)` below.

matthew-reynolds (changes requested):
I haven't looked coverage with CI before, but I remember we were both impressed by the [codecov](https://codecov.io/) tool that catkin_lint uses. Looking into that a little bit, uploading the coverage reports to codecov and integrating with GitHub looks super simple.

To actually run the code coverage with CI, we can just add the CMake args (`-DENABLE_COVERAGE_TESTING=ON -DCMAKE_BUILD_TYPE=Debug`) to our build matrix, and then add the `${PROJECT_NAME}_coverage` target to some aggregating target within the `if(ENABLE_COVERAGE_TESTING)` to simplify running coverage for all packages.
 - The easy way is just add `add_dependencies(run_tests ${PROJECT_NAME}_coverage)`.
 - A "tidier" solution might be to create a `coverage` target, and `add_dependencies(coverage ${PROJECT_NAME}_coverage)` in each package, but I don't think `industrial_ci` has a way to build custom targets, since there's no `catkin-make-args` or equivalent. So I'm not sure how we'd actually build this `coverage` target

matthew-reynolds:
Sorry, that was a bit rambly.

Basically I'm suggesting the `CMakeLists.txt` could contain this:

```
if(CATKIN_ENABLE_TESTING)

 ...

  if(ENABLE_COVERAGE_TESTING)
    find_package(code_coverage REQUIRED)
    APPEND_COVERAGE_COMPILER_FLAGS()
    set(COVERAGE_EXCLUDES "*/${PROJECT_NAME}/test*")
    add_code_coverage(NAME ${PROJECT_NAME}_coverage)
    add_dependencies(run_tests ${PROJECT_NAME}_coverage)
  endif()
endif()
```

And the `.travis.yml` should add the following to the build matrix:
```
CMAKE_ARGS="-DENABLE_COVERAGE_TESTING=ON -DCMAKE_BUILD_TYPE=Debug"
```

From there, the codecov integration should be easy.

joint_trajectory_controller/CMakeLists.txt
martiniil: I think the idea is to append the compiler flags before defining any targets. See the usage example in https://github.com/mikeferguson/code_coverage

martiniil:
You could also build the coverage target using the `AFTER_SCRIPT` variable of `industrial_ci`. See
https://github.com/PilzDE/pilz_robots/blob/melodic-devel/.travis.yml and
https://github.com/PilzDE/pilz_robots/blob/melodic-devel/.coverage.sh

@matthew-reynolds which solution do you prefer?

joint_trajectory_controller/CMakeLists.txt
matthew-reynolds: Ah, makes sense

matthew-reynolds:
Nice, the `AFTER_SCRIPT` solution is definitely cleaner, let's go with that. Big +1 for not "abusing" the `run_tests` target like I was suggesting.

I don't really like the idea of having to manually list all the packages we're testing coverage of, seems a bit fragile, but that's easy to work around in the `.coverage.sh` script (Maybe just grep each package's `CMakeLists.txt` for `add_code_coverage`).

Plus, if we use a script like this, perhaps we can also look into how to merge multiple coverage reports together! Not an immediate need, of course, we're only testing the one package right now.

bmagyar (approved):
Green lights from me, up for you @matthew-reynolds

matthew-reynolds (approved):
LGTM :) Looking forward to the follow-up PR to run the coverage during CI and upload the results - Happy to work on that myself, unless @martiniil you'd like to take it.

bmagyar:
Thanks to everyone involved, merging...

martiniil:
Thank you! @matthew-reynolds feel free to go ahead.

PR #469, opened 2020-03-12
Check for nan cmd_vel

piraka9011:
Ignore `NaN` velocity commands sent to the differential drive controller.
cc: @bmagyar for review

Signed-off-by: Anas Abou Allaban <aabouallaban@pm.me>


bmagyar (approved):
Thanks for addressing all points from the previous PR!

PR #471, opened 2020-03-15
Refactor nan test

bmagyar:
* EXPECT_NE(x, bool) -> EXPECT_TRUE/FALSE(x)
* EXPECT_EQ(x, double) -> EXPECT_DOUBLE_EQ(x, double)
* clang default reformat

Complementary to #469

### DIFF
--- a/diff_drive_controller/test/diff_drive_nan_test.cpp
+++ b/diff_drive_controller/test/diff_drive_nan_test.cpp
@@ -33,11 +33,9 @@
 #include <limits>
 
 // TEST CASES
-TEST_F(DiffDriveControllerTest, testNaN)
-{
+TEST_F(DiffDriveControllerTest, testNaN) {
   // wait for ROS
-  while(!isControllerAlive())
-  {
+  while (!isControllerAlive()) {
     ros::Duration(0.1).sleep();
   }
   // zero everything before test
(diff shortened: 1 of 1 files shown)


matthew-reynolds (changes requested):
+1 for auto-formatting in principle, but I think we need to decide on a style (that is consistent with as much of the code-base as possible) and add a `.clang-format` to the repo first. I looked through the rest of the `.h` and `.cpp` files and, as far as I can tell, every one is indented with 2 spaces, but this formatting is uses 4 spaces.

The `EXPECT_FALSE` and `EXPECT_DOUBLE_EQ` look good :+1:

piraka9011:
In principal, this LGTM. But echoing @matthew-reynolds it seems that the formatting here is different from the rest of the files. Is that intended?
I did my best to follow the convention of the other tests in the package.

bmagyar:
Thanks for catching that guys, that was my personal clang settings. I've pushed a new version where I used the clang-format file from MoveIt as it is quite streamlined and simple. There's also a PR with it: #472

diff_drive_controller/test/diff_drive_nan_test.cpp
```diff
@@ -93,25 +93,25 @@ TEST_F(DiffDriveControllerTest, testNaNCmd)
...
     ros::Duration(0.1).sleep();
   }
 
   nav_msgs::Odometry odom = getLastOdom();
-  EXPECT_EQ(odom.twist.twist.linear.x, 0.0);
-  EXPECT_EQ(odom.pose.pose.position.x, 0.0);
-  EXPECT_EQ(odom.pose.pose.position.y, 0.0);
+  EXPECT_DOUBLE_EQ(odom.twist.twist.linear.x, 0.0);
+  EXPECT_DOUBLE_EQ(odom.pose.pose.position.x, 0.0);
+  EXPECT_DOUBLE_EQ(odom.pose.pose.position.y, 0.0);
```
martiniil, line 106: I would always use EXPECT_NEAR to compare double values.

bmagyar: This is pretty much that with a tiny tiny epsilon
https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#floating-point-comparison

diff_drive_controller/test/diff_drive_nan_test.cpp
```diff
@@ -93,25 +93,25 @@ TEST_F(DiffDriveControllerTest, testNaNCmd)
...
 
   geometry_msgs::TwistStamped odom_msg = getLastCmdVelOut();
-  EXPECT_EQ(fabs(odom_msg.twist.linear.x), 0);
+  EXPECT_DOUBLE_EQ(fabs(odom_msg.twist.linear.x), 0);
```
matthew-reynolds: Why are we getting the absolute value here? Also, let's make this value `0` -> `0.0` to be extra explicit about it being a `double` comparison.

piraka9011 (approved):
Aside from Matt's suggestion this LGTM.

diff_drive_controller/test/diff_drive_nan_test.cpp
bmagyar: Wishful thinking I suppose, if something `x` is expected to be 0, `abs(x)` will be 0 as well :)

bmagyar: I've fixed this

matthew-reynolds (approved):
Looks good!

PR #472, opened 2020-03-16
Add clang-format file

bmagyar:
I wouldn't necessarily start reformatting everything just yet but by simply having this here people's IDEs will probably start picking it up. I'd set a one month grace period for this, then go ahead and format all the repo? This is for ROS1-only.

What do you guys think? @ipa-mdl @matthew-reynolds @piraka9011

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,66 @@
+---
+BasedOnStyle:  Google
+AccessModifierOffset: -2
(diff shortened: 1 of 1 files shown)


piraka9011:
I propose we include all formatting changes in one PR (including the clang-format file) so that there's no divergence in style since the format file's introduction.
It would be easier to revert that way if needed.

Issue #462, opened 2020-02-13
Input and output of read() and write() methods in hardware_interface


fjp:
In the [diffbot test](https://github.com/ros-controls/ros_controllers/blob/melodic-devel/diff_drive_controller/test/diffbot.h), I am confused about the definitions of [read()](https://github.com/ros-controls/ros_controllers/blob/884f4c2605151684bd61a78a5e7e9a0bf5b928cc/diff_drive_controller/test/diffbot.h#L83) and [wirte()](https://github.com/ros-controls/ros_controllers/blob/884f4c2605151684bd61a78a5e7e9a0bf5b928cc/diff_drive_controller/test/diffbot.h#L95) methods. In my opinion the content of the write method should be swaped with the content of the read method.
I think the write() method of a hardware_interface should write to the actuators and therefore fill the cmd_ member. And the same for the read method: The members vel_ and pos_ are filled in the read method because the hardware_interface reads the data from the sensors.

The read method of the hardware_interface in [this repo](https://github.com/eborghi10/my_ROS_mobile_robot/blob/e04acfd3e7eb4584ba0aab8a969a74d6c30eed34/my_robot_base/include/my_robot_hw_interface.h#L81) seems correct to me. The blog post: https://slaterobots.com/blog/5abd8a1ed4442a651de5cb5b/how-to-implement-ros_control-on-a-custom-robot states the following:

> 1. init(): Register joint-specific handles to each type of controller interfaces (joint state, position, effort, etc.)
> 2. read(): Read joint positions from the robot's hardware and set the joint_position_ array with that data.
> 3. write(): Actuate the robot's joints using the joint_effort_command_ variable. Based on our position controllers defined in config/controllers.yaml, this will be set by ros_control by using the desired joint position (joint_position_command_), the error between joint_position_ and joint_position_command_, and the PID parameter for that controller in config/controllers.yaml--all of which gets calculated when we call update().
> 4. update(): simply read joint state, call update() on our controller manager, and write/actuate the joints based on what's calculated.

Therefore, I think the logic of read and write in the diffbot test should be swapped.

Could someone please explain the correct or intended way of how to define the read and write methods?

matthew-reynolds:
[...]

PR #474, opened 2020-03-19
[diff_drive_controller] swap implementation of read and write methods of Diffbot class

fjp:
This PR was discussed with @matthew-reynolds in #462 and swaps the implementations of the `read()` and `write()` methods of the `Diffbot` class in the `diff_drive_controller` package.

I've run all the tests for `diff_drive_controller` successfully. 

If this is PR gets accepted I can start looking at the other mentioned tests in #462.

### DIFF
--- a/diff_drive_controller/test/diffbot.h
+++ b/diff_drive_controller/test/diffbot.h
@@ -82,18 +82,7 @@ public:
 
   void read()
   {
-    std::ostringstream os;
-    for (unsigned int i = 0; i < NUM_JOINTS - 1; ++i)
-    {
-      os << cmd_[i] << ", ";
-    }
-    os << cmd_[NUM_JOINTS - 1];
-
-    ROS_INFO_STREAM("Commands for joints: " << os.str());
-  }
-
-  void write()
-  {
+    // Read the joint state of the robot into the hardware interface
     if (running_)
     {
       for (unsigned int i = 0; i < NUM_JOINTS; ++i)
@@ -110,6 +99,19 @@ public:
       std::fill_n(pos_, NUM_JOINTS, std::numeric_limits<double>::quiet_NaN());
       std::fill_n(vel_, NUM_JOINTS, std::numeric_limits<double>::quiet_NaN());
     }
+  }
+
+  void write()
+  {
+    // Write the commands to the joints
+    std::ostringstream os;
+    for (unsigned int i = 0; i < NUM_JOINTS - 1; ++i)
+    {
+      os << cmd_[i] << ", ";
+    }
+    os << cmd_[NUM_JOINTS - 1];
+
+    ROS_INFO_STREAM("Commands for joints: " << os.str());
   }
 
   bool start_callback(std_srvs::Empty::Request& /*req*/, std_srvs::Empty::Response& /*res*/)


matthew-reynolds (approved):
Personally this looks good to me, but looking at @ipa-mdl's review of #364, I think this has a better chance of more wide-spread approval if we focus on only moving the command print, and leaving the calculations where they were.

Ping @bmagyar @ipa-mdl for your thoughts. Full discussion in #462

bmagyar:
Thanks guys for the discussions. I think it is easy to swap read/write as it largely depends on one's directional perspective, probably this is what happened when we first wrote those tests.

I was tempted to accept this right away *but* the little engineer on my shoulder says: "If you are unhappy with unclear documentation and want to fix code, please also take time to update said documentation.". In this case it'd mean updating the docstrings for `RobotHW::read()` and `RobotHW::write()`

fjp:
Thanks @matthew-reynolds and @bmagyar looking into this PR. 

I think with this PR the current docstrings of [`RobotHW::read`](https://github.com/ros-controls/ros_control/blob/b78bccb5f3768d1726db79acbe8837cc5187b515/hardware_interface/include/hardware_interface/robot_hw.h#L164) and [`RobotHW::write()`](https://github.com/ros-controls/ros_control/blob/b78bccb5f3768d1726db79acbe8837cc5187b515/hardware_interface/include/hardware_interface/robot_hw.h#L172) describe the correct behaviour and don't necessarily need to be changed. 

To summarise why I created #462 : I first read the documentation (of the hardware_interface), then looked at the tests of `DiffBot` and was confused with the implementation of `DiffBot`'s `read()` and `write()` methods because it didn't fit to the documentation. Now with this PR, I think doc and code fit together.

However, I guess it could help to put some more details into the docstrings regarding "subjective perspective" of these two methods. Should we add `@brief` and `@details` to the `RobotHW::write()` and `RobotHW::read()` methods?

---

Another thought that came to my mind and as @matthew-reynolds [mentioned](https://github.com/ros-controls/ros_controllers/issues/462#issuecomment-595829508):

> these calculations don't perfectly fit into the model of the RobotHW read/update/write cycle anyways.

Maybe those tests of `DiffBot` weren't written with the `RobotHW` and its virtual methods in mind? If they were, I think we should use override for `DiffBot`'s `read` and `write` methods with the same method signatures as in `RobotHW`.

bmagyar:
@fjp +1 for `@brief`

fjp:
Ok, I'll try to improve the documentation for [`RobotHW::read`](https://github.com/ros-controls/ros_control/blob/b78bccb5f3768d1726db79acbe8837cc5187b515/hardware_interface/include/hardware_interface/robot_hw.h#L164) and [`RobotHW::write()`](https://github.com/ros-controls/ros_control/blob/b78bccb5f3768d1726db79acbe8837cc5187b515/hardware_interface/include/hardware_interface/robot_hw.h#L172) methods in another PR this week. 

But please bear with me, as I am new to ROS control and struggling a bit with the documentation as it is rather sparse and not all in one place... No offence though, I think ROS control is great and I just need to use it more in order to understand it better.

bmagyar:
Anything you find worth mentioning please add here: https://github.com/ros-controls/roadmap/blob/master/documentation_resources.md in the form of a PR, we are trying to collect all docs on `ros_control` and unify them.

bmagyar (approved):

PR #475, opened 2020-03-20
fix minor typo in documentation of setGains method

fjp:
Changes Get to Set of setGains method documentation


bmagyar (approved):
Nice catch, thanks!

PR #477, opened 2020-04-01
Allow position control limits outside [-pi,pi]

### LINKED ISSUE #261: Don't wrap angle differences optionally in joint position controllers
I'd like to be able to command a continuous joint to for example 20 pi and see it revolve 10 times and then stop.   Currently the angle wrapping (calls to angles.h `shortest_angular_distance()`) prevents this from happening.  Would it make sense to create a new joint type that doesn't have this property, or add options to the existing controller types to support it?

Another feature to go with this would be to reset the position, so that if the controller gets many revolutions away from zero the angle can be wrapped back down to near zero but only when requested.

http://answers.ros.org/question/197805/are-jointpositioncontrollers-supposed-to-wrap-angles-on-continuous-joints/

scpeters:
Several joint position controllers use the [angles::shortest_angular_distance_with_limits](http://docs.ros.org/melodic/api/angles/html/namespaceangles.html#a4436fe67ae0c9df020f6779101bbefab) function for `REVOLUTE` joints to compute the position error, but this function only supports joint limits within the interval [-pi, pi].

The [angles::shortest_angular_distance_with_large_limits](http://docs.ros.org/melodic/api/angles/html/namespaceangles.html#a0a260194cd877fe061195116274f1d85) function supports limits outside the interval [-pi, pi], which I think is more suitable for this computation. A difference from the other API is that the lower limit is required to be smaller than the upper limit.

Closes #261.

### DIFF
--- a/effort_controllers/src/joint_group_position_controller.cpp
+++ b/effort_controllers/src/joint_group_position_controller.cpp
@@ -138,7 +138,7 @@ namespace effort_controllers
         // Compute position error
         if (joint_urdfs_[i]->type == urdf::Joint::REVOLUTE)
         {
-         angles::shortest_angular_distance_with_limits(
+          angles::shortest_angular_distance_with_large_limits(
             current_position,
             command_position,
             joint_urdfs_[i]->limits->lower,
--- a/effort_controllers/src/joint_position_controller.cpp
+++ b/effort_controllers/src/joint_position_controller.cpp
@@ -181,7 +181,7 @@ void JointPositionController::update(const ros::Time& time, const ros::Duration&
   // Compute position error
   if (joint_urdf_->type == urdf::Joint::REVOLUTE)
   {
-   angles::shortest_angular_distance_with_limits(
+    angles::shortest_angular_distance_with_large_limits(
       current_position,
       command_position,
       joint_urdf_->limits->lower,
--- a/velocity_controllers/src/joint_position_controller.cpp
+++ b/velocity_controllers/src/joint_position_controller.cpp
@@ -182,7 +182,7 @@ void JointPositionController::update(const ros::Time& time, const ros::Duration&
   // Compute position error
   if (joint_urdf_->type == urdf::Joint::REVOLUTE)
   {
-   angles::shortest_angular_distance_with_limits(
+    angles::shortest_angular_distance_with_large_limits(
       current_position,
       command_position,
       joint_urdf_->limits->lower,


matthew-reynolds (changes requested):
LGTM but I think this should target Noetic since it changes behaviour. If people are taking advantage of the wrapping that `angles::shortest_angular_distance_with_limits()` does, then requiring `lower` < `upper` might make a mess for them.

bmagyar:
#466  ?

scpeters:
> LGTM but I think this should target Noetic since it changes behaviour.

I don't think this repository has a `noetic-devel` branch yet

scpeters:
> #466 ?

Oops, you're right I didn't notice that when I made this PR. I believe this PR matches what is recommended in the following comment thread: https://github.com/ros-controls/ros_controllers/pull/466#discussion_r389576405

As I look back further, #264 is also related, but I think it predates the `angles::shortest_angular_distance_with_large_limits` function.

matthew-reynolds:
> #466 ?

Whoops, I already forgot about that guy. Thanks @bmagyar :)

scpeters:
> > LGTM but I think this should target Noetic since it changes behaviour.

> I don't think this repository has a noetic-devel branch yet

bump @bmagyar; if this should target `noetic-devel`, can you create that branch?

scpeters:
never mind, #466 is a better version of this since it has a test.

bmagyar:
@matthew-reynolds while I agree that it changes behaviour slightly, some may say this is a bug, not a feature, even if people found ways to work around it. I think it should still go into melodic

matthew-reynolds:
@bmagyar Yep, you're right. I also forgot that the joint limit logic requires `lower < upper`, so if people have `upper < lower` they've already got problems. Definitely good for Melodic.

bmagyar:
#466 was merged, closing this

swinterbotix:
Not sure if this is the right place to put this, but...

I've literally just ran into this exact issue on a robot that I'm working with. Gazebo behaves nicely when using effort_controllers/JointTrajectoryController, but not so much when using effort_controllers/JointPositionController. In my URDF, I specify limits for the revolute joint in question as...

`<limit effort="100" lower="${-pi}" upper="${pi}" velocity="${pi}"/>`

...but this causes the joint to rotate to pi right after the controllers are loaded and oscillate like mad around pi a few degrees. However, if I redefine the limits to...

`<limit effort="100" lower="-3.14" upper="3.14" velocity="${pi}"/>`

...the problem goes away. My guess is that there is a tiny bit of overlap between ${-pi} and ${pi} that's causing this issue. Not sure how this should be addressed, but thought I'd point it out.

I'm currently using Ubuntu 16.04 and ROS Kinetic.

francofusco:
Not sure about the joint moving to PI in the very beginning, but I was indeed experiencing oscillations when limits were slightly overlapping. The changes proposed here and in #466 should fix the problem - though I think they have been merged only in the melodic branch, not kinetic...

If you need a quick fix, perhaps the best is to clone the controllers locally and modify them manually to use `angular_distance_with_large_limits` instead of `angular_distance_with_limits`

PR #479, opened 2020-04-06
Fix warning dynamic_reconfigure

rbonghi:
Hi all,

I was working on my package when I approach this warning on my screen 
image

> [ WARN] [1586198292.260228386]: updateConfig() called on a dynamic_reconfigure::Server that provides its own mutex. This can lead to deadlocks if updateConfig() is called during an update. Providing a mutex to the constructor is highly recommended in this case. Please forward this message to the node author.

This pull-request fix the initialisation dynamic_reconfigure and now this warning is not show anymore.

PS: This pull-request should fix this answer https://answers.ros.org/question/310267/updateconfig-called-on-a-dynamic_reconfigureserver-that-provides-its-own-mutex/

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
@@ -234,8 +234,8 @@ namespace diff_drive_controller{
 
     /// Dynamic Reconfigure server
     typedef dynamic_reconfigure::Server<DiffDriveControllerConfig> ReconfigureServer;
-
     std::shared_ptr<ReconfigureServer> dyn_reconf_server_;
+    boost::recursive_mutex dyn_reconf_server_mutex_;
 
   private:
     /**
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -374,8 +374,13 @@ namespace diff_drive_controller{
     config.publish_rate = publish_rate;
     config.enable_odom_tf = enable_odom_tf_;
 
-    dyn_reconf_server_ = std::make_shared<ReconfigureServer>(controller_nh);
+    dyn_reconf_server_ = std::make_shared<ReconfigureServer>(dyn_reconf_server_mutex_, controller_nh);
+
+    // Update parameters
+    dyn_reconf_server_mutex_.lock();
     dyn_reconf_server_->updateConfig(config);
+    dyn_reconf_server_mutex_.unlock();
+
     dyn_reconf_server_->setCallback(boost::bind(&DiffDriveController::reconfCallback, this, _1, _2));
 
     return true;


diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
```diff
@@ -229,10 +229,12 @@ namespace diff_drive_controller{
 
     realtime_tools::RealtimeBuffer<DynamicParams> dynamic_params_;
 
-    /// Dynamic Reconfigure server
+
     typedef dynamic_reconfigure::Server<DiffDriveControllerConfig> ReconfigureServer;
-    
-    std::shared_ptr<ReconfigureServer> dyn_reconf_server_;
+    /// To avoid Dynamic Reconfigure Server warning
+    boost::recursive_mutex mDynServerMutex;
```
bmagyar: Please always follow the reigning coding style when contributing to a codebase.
```suggestion
    boost::recursive_mutex dyn_server_mutex_;
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -374,9 +374,15 @@ namespace diff_drive_controller{
     config.publish_rate = publish_rate;
     config.enable_odom_tf = enable_odom_tf_;
 
-    dyn_reconf_server_ = std::make_shared<ReconfigureServer>(controller_nh);
+    dyn_reconf_server_ = boost::make_shared<ReconfigureServer>(mDynServerMutex, controller_nh);
+    ReconfigureServer::CallbackType f;
+    f = boost::bind(&DiffDriveController::reconfCallback, this, _1, _2);
+    dyn_reconf_server_->setCallback(f);
```
bmagyar: No need to expand the one-liner of setting the callback into 3 lines, please change it back

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -374,9 +374,15 @@ namespace diff_drive_controller{
     config.publish_rate = publish_rate;
     config.enable_odom_tf = enable_odom_tf_;
 
-    dyn_reconf_server_ = std::make_shared<ReconfigureServer>(controller_nh);
+    dyn_reconf_server_ = boost::make_shared<ReconfigureServer>(mDynServerMutex, controller_nh);
```
bmagyar: I am not sure this won't cause other issues. The controller itself already has mutexes and there are mutex-protected structures in place. Besides, we'd like to move away from boost as much as possible and use C++11/14 features instead. @matthew-reynolds what do you think?

bmagyar (changes requested):

diff_drive_controller/src/diff_drive_controller.cpp
rbonghi: I totally agree with you, I would like to stay away from boost as well. With this commit anyway I found where was from the warning..
I can try to check without boost and see if works. ;-)

I keep you posted!

rbonghi: I am sorry for my second message, but my curiosity pushed me to compile again the code.

I am a little rusty, but I do not think that we can avoid to use boost here. The `dynamic_reconfigure::Server` use `boost::recursive_mutex` and I do not remember any trick to convert a `boost` mutex to `std`.

The reference is of `dynamic_reconfigure::Server` is here: https://github.com/ros/dynamic_reconfigure/blob/melodic-devel/include/dynamic_reconfigure/server.h#L75

Anyway I will sleep on it!
Tomorrow I will send you a commit with all reviewed parts.

matthew-reynolds: The `shared_ptr` can definitely still be `std`. Passing the `boost::recursive_mutex` as an argument to the constructor/`make_shared` doesn't affect the type of the `shared_ptr`.

But yes, we can't pass a `std::recursive_mutex` to that constructor since it asks for `boost`. We will have to live with that Boost usage.

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
matthew-reynolds: Or better yet, to match the name of server, `dyn_reconf_server_mutex_`

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -374,9 +374,15 @@ namespace diff_drive_controller{
...
+    ReconfigureServer::CallbackType f;
+    f = boost::bind(&DiffDriveController::reconfCallback, this, _1, _2);
+    dyn_reconf_server_->setCallback(f);
+    // Update parameters
+    dyn_reconf_server_->getConfigDefault(config);
```
matthew-reynolds: We don't want this `getConfigDefault()` - It will override the values we just set in `config`.

matthew-reynolds: I suspect this is why the tests failed.

rbonghi:
I updated the mutex `mDynServerMutex` to  `dyn_reconf_server_mutex_` following @matthew-reynolds .

I removed the line
```diff
-    dyn_reconf_server_->getConfigDefault(config);
```
Should be fine now.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -374,9 +374,12 @@ namespace diff_drive_controller{
...
     config.enable_odom_tf = enable_odom_tf_;
 
-    dyn_reconf_server_ = std::make_shared<ReconfigureServer>(controller_nh);
-    dyn_reconf_server_->updateConfig(config);
+    dyn_reconf_server_ = boost::make_shared<ReconfigureServer>(dyn_reconf_server_mutex_, controller_nh);
     dyn_reconf_server_->setCallback(boost::bind(&DiffDriveController::reconfCallback, this, _1, _2));
+    // Update parameters
+    dyn_reconf_server_mutex_.lock();
+    dyn_reconf_server_->updateConfig(config);
+    dyn_reconf_server_mutex_.unlock();
```
matthew-reynolds: Didn't notice this before, but let's make sure we call `updateConfig` before `setCallback`. Otherwise, the callback will be triggered when we update the config.

matthew-reynolds (changes requested):

matthew-reynolds:
Not sure what's up with Travis... I re-ran the build for the last commit (merge 5c3ff61 into `melodic-devel`) but it didn't update the status on GitHub, and now newest commit isn't being built either. @bmagyar thoughts?

Similar issue here: https://github.com/travis-ci/travis-ci/issues/7643, but since the GitHub-Travis integration is still working properly on other PRs in this repo (#466), I don't think that's the issue.

diff_drive_controller/src/diff_drive_controller.cpp
rbonghi: Fixed

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h
```diff
@@ -229,10 +229,11 @@ namespace diff_drive_controller{
...
     realtime_tools::RealtimeBuffer<DynamicParams> dynamic_params_;
 
-    /// Dynamic Reconfigure server
     typedef dynamic_reconfigure::Server<DiffDriveControllerConfig> ReconfigureServer;
-    
-    std::shared_ptr<ReconfigureServer> dyn_reconf_server_;
+    /// To avoid Dynamic Reconfigure Server warning
+    boost::recursive_mutex dyn_reconf_server_mutex_;
+    /// Dynamic Reconfigure server
+    boost::shared_ptr<ReconfigureServer> dyn_reconf_server_;
```
matthew-reynolds: As mentioned in https://github.com/ros-controls/ros_controllers/pull/479#discussion_r404443977, this `boost::shared_ptr` still needs to be switched back to `std::shared_ptr`.

Also, bit of a nit-pick: Even though it isn't perfect Doxygen, let's keep with this file's grouping comment pattern, and place the `typedef ... T` and `shared_ptr<T>` together before the mutex:

```
    /// Dynamic Reconfigure server:
    typedef dynamic_reconfigure::Server<DiffDriveControllerConfig> ReconfigureServer;
    std::shared_ptr<ReconfigureServer> dyn_reconf_server_;
    boost::recursive_mutex dyn_reconf_server_mutex_;
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -374,8 +374,12 @@ namespace diff_drive_controller{
     config.publish_rate = publish_rate;
     config.enable_odom_tf = enable_odom_tf_;
 
-    dyn_reconf_server_ = std::make_shared<ReconfigureServer>(controller_nh);
+    dyn_reconf_server_ = boost::make_shared<ReconfigureServer>(dyn_reconf_server_mutex_, controller_nh);
+    // Update parameters
+    dyn_reconf_server_mutex_.lock();
```
matthew-reynolds: Nitpick: blank line before comment to match the prevailing style please.

matthew-reynolds (changes requested):
Looks good to me, just need to switch the `boost::shared_ptr` back to `std::shared_ptr` and a couple style nitpicks.

rbonghi:
Hi @matthew-reynolds ,

I followed your hints and I updated all changes. In the first time https://github.com/ros-controls/ros_controllers/pull/479#discussion_r404443977 I thought was the same to use a `boost` and I didn't change it.

On travis is all green https://travis-ci.org/github/ros-controls/ros_controllers/builds/673500215 I do not why is in pending here.

matthew-reynolds (approved):
Yeah, Travis seems to be having some difficulty in the last few days.

Thanks for iterating, looks good to me.

bmagyar (approved):

bmagyar:
I've squashed & rebased on the latest melodic-devel, going to merge when the CI is happy

PR #481, opened 2020-04-09
Modernize xacro

matthew-reynolds:
Drop deprecated `xacro` features in preparation for Noetic. See https://github.com/ros/xacro/blob/noetic-devel/CHANGELOG.rst

This fixes three warnings we get a lot in our tests at the moment:
 - `xacro: in-order processing became default in ROS Melodic. You can drop the option.`
 - `Deprecated: xacro tag 'XYZ' w/o 'xacro:' xml namespace prefix (will be forbidden in Noetic)`
 - `xacro.py is deprecated; please use xacro instead`

Fixes for the second and third items could be backported to `kinetic-devel` if we want.


bmagyar (approved):
Thanks for this, I wouldn't spend time on putting this into kinetic at this point.

PR #484, opened 2020-04-16
Migrate to new industrial_ci

matthew-reynolds:
See also https://github.com/ros-controls/ros_control/pull/440

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,25 @@
 # This config file for Travis CI utilizes ros-industrial/industrial_ci package.
 # For more info for the package, see https://github.com/ros-industrial/industrial_ci/blob/master/README.rst
-dist: trusty
+language: generic
 services:
   - docker
-language: generic
+
 notifications:
   email:
-    on_success: change
-    on_failure: change
     recipients:
       - enrique.fernandez.perdomo@gmail.com
       - bence.magyar.robotics@gmail.com
       - gm130s@gmail.com
+    on_success: change #[always|never|change] # default: change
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,14 +13,11 @@ notifications:
...
       - gm130s@gmail.com
 env:
-  global:
-    - CATKIN_PARALLEL_TEST_JOBS=-p1
-    - ROS_PARALLEL_TEST_JOBS=-j1
   matrix:
     - ROS_DISTRO=melodic  ROS_REPO=ros
     - ROS_DISTRO=melodic  ROS_REPO=ros-testing
-    - ROS_DISTRO=melodic  UPSTREAM_WORKSPACE=https://raw.githubusercontent.com/ros-controls/ros_control/$ROS_DISTRO-devel/ros_control.rosinstall
+    - ROS_DISTRO=melodic  UPSTREAM_WORKSPACE='https://raw.githubusercontent.com/ros-controls/ros_control/$ROS_DISTRO-devel/ros_control.rosinstall -ros_controllers'
```
matthew-reynolds, line 20: Is there a reason we have this `rosinstall` setup as a third job, rather than applied to both other jobs (As is done in all the other `ros_control` repos?)

I think this approach actually might make more sense than the other repos, since we're testing against 3 different versions of the other `ros_control` packages rather than always pulling from GitHub.

bmagyar: I think the reason for this was to allow breakage in one direction but not in others.

bmagyar (approved):

PR #497, opened 2020-07-09
Fix: JointTrajectoryController tolerance checking using incorrect joint error

dougsm:
`state_joint_error_` is used as a temp variable inside a loop over joints in `updateStates`, so always contains the error values for the last joint in the loop.  As such, tolerance checking was only ever checking the error of the last joint, but against the tolerances for each joint. 

This simple fix just puts the correct error values into `state_joint_error_` before the tolerances are checked.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -382,6 +382,11 @@ update(const ros::Time& time, const ros::Duration& period)
       return;
     }
 
+    // Get state error for current joint
+    state_joint_error_.position[0] = state_error_.position[i];
+    state_joint_error_.velocity[0] = state_error_.velocity[i];
+    state_joint_error_.acceleration[0] = 0.0;
+
     //Check tolerances
     const RealtimeGoalHandlePtr rt_segment_goal = segment_it->getGoalHandle();
     if (rt_segment_goal && rt_segment_goal == rt_active_goal_)
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -385,7 +385,7 @@ update(const ros::Time& time, const ros::Duration& period)
     // Get state error for current joint
     state_joint_error_.position[0] = state_error_.position[i];
     state_joint_error_.velocity[0] = state_error_.velocity[i];
-    state_joint_error_.acceleration[0] = 0.0;
+    state_joint_error_.acceleration[0] = state_error_.acceleration[i];
 
     //Check tolerances
     const RealtimeGoalHandlePtr rt_segment_goal = segment_it->getGoalHandle();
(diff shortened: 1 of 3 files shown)


dougsm:
@bmagyar  I think this was initially introduced by your commit 99c035b1adaafb8c0e81ceae75ee0da0d95faf62

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -382,6 +382,11 @@ update(const ros::Time& time, const ros::Duration& period)
       return;
     }
 
+    // Get state error for current joint
+    state_joint_error_.position[0] = state_error_.position[i];
+    state_joint_error_.velocity[0] = state_error_.velocity[i];
+    state_joint_error_.acceleration[0] = 0.0;
```
bmagyar:
```suggestion
    state_joint_error_.acceleration[0] = state_joint_error_.acceleration[i];
```

we don't have to know at this point if this is set to 0.0

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -382,6 +382,11 @@ update(const ros::Time& time, const ros::Duration& period)
       return;
     }
 
+    // Get state error for current joint
+    state_joint_error_.position[0] = state_error_.position[i];
```
bmagyar, line 386: are you 100% sure that the indices here are correct?

bmagyar (changes requested):

bmagyar:
@dougsm sorry for the delay on this one

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
bmagyar: Ok I think you are right but I have a aslightly different idea for a solution.

`state_joint_error_` is truly a local variable and therefore shouldn't even be a member. The scope of it is reduced entirely to this section, it's updated by `updateStates` which is a useless update, the code could be removed from there...

let me put a commit on top of your PR here

bmagyar: due to StateSegment not being particularly realtime friendly I couldn't manage a good local var fix but I think this will do

bmagyar:
@dougsm could you give this a review/test please? I've rebased (for better CI) and added a commit on top

bmagyar:
@mateus-amarante @matthew-reynolds could you guys please give this a review/test?

mateus-amarante (commented):
The update sounds correct! Before, all joint tolerances were compared to the error of the last joint.

I think the [tests](https://github.com/ros-controls/ros_controllers/blob/b551a652147a2095a9d0391b100cbc0043fac2ec/joint_trajectory_controller/test/joint_trajectory_controller_test.cpp#L1179-L1258) were not catching this because the forced error is applied to both joints of the rrbot through the `smoothing` parameter ([impl](https://github.com/ros-controls/ros_controllers/blob/b551a652147a2095a9d0391b100cbc0043fac2ec/joint_trajectory_controller/test/rrbot.cpp#L113-L145)). The tests should check the tolerance violation of each individual joint.

One way to solve this would be by updating the interface of the rrbot to set the `smoothing` factor for each individual joint, then test the tolerance violation for each joint.

mateus-amarante:
> One way to solve this would be by updating the interface of the rrbot to set the smoothing factor for each individual joint, then test the tolerance violation for each joint.

#523 implements this idea

bmagyar:
@mateus-amarante your changes properly flag things up on #523 , I've cherry picked those tests onto this PR

bmagyar (approved):

bmagyar:
@dougsm thanks very much for spotting this and providing a fix.
@mateus-amarante many thanks for tracking down how we could properly test this!!

PR #503, opened 2020-08-25
[effort_controllers] JointGroupPositionController: Load URDF from namesdpace

wxmerkt:
Same as #245


bmagyar:
Good catch! The failing travis job is a flaky diff drive test, merging this one!

PR #504, opened 2020-08-25
JointGroupPositionController: Set current position as target in starting()

wxmerkt:
Currently, the target position upon starting is all zeros. This is not great if a command is not issued immediately (and even then). To be safer, set the goal position to the current sensed position and also reset the PID internal state upon starting.

(This PR is on top of #503)

### DIFF
--- a/effort_controllers/src/joint_group_position_controller.cpp
+++ b/effort_controllers/src/joint_group_position_controller.cpp
@@ -129,7 +129,7 @@ namespace effort_controllers
       enforceJointLimits(current_positions[i], i);
       pid_controllers_[i].reset();
     }
-    commands_buffer_.writeFromNonRT(current_positions);
+    commands_buffer_.initRT(current_positions);
   }
 
   void JointGroupPositionController::update(const ros::Time& time, const ros::Duration& period)
--- a/effort_controllers/src/joint_group_position_controller.cpp
+++ b/effort_controllers/src/joint_group_position_controller.cpp
@@ -120,6 +120,18 @@ namespace effort_controllers
     return true;
   }
 
+  void JointGroupPositionController::starting(const ros::Time& time)
+  {
+    std::vector<double> current_positions(n_joints_, 0.0);
+    for (std::size_t i = 0; i < n_joints_; ++i)
+    {
+      current_positions[i] = joints_[i].getPosition();
+      enforceJointLimits(current_positions[i], i);
+      pid_controllers_[i].reset();
+    }
+    commands_buffer_.writeFromNonRT(current_positions);
+  }
+
   void JointGroupPositionController::update(const ros::Time& time, const ros::Duration& period)
   {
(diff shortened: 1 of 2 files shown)


effort_controllers/src/joint_group_position_controller.cpp
```diff
@@ -120,14 +120,26 @@ namespace effort_controllers
...
+  void JointGroupPositionController::starting(const ros::Time& time)
+  {
+    std::vector<double> current_positions(n_joints_, 0.0);
+    for (std::size_t i = 0; i < n_joints_; ++i)
+    {
+      current_positions[i] = joints_[i].getPosition();
+      enforceJointLimits(current_positions[i], i);
+      pid_controllers_[i].reset();
+    }
+    commands_buffer_.writeFromNonRT(current_positions);
```
matthew-reynolds: `starting()` is in the RT domain, so rather than calling `writeFromNonRT()`, you should write to the command buffer directly. (The realtime buffer lacks a proper writeFromRT method since it's really intended for writing from a single non-RT thread, reading from a single RT thead. I haven't worked through the accesses and how safe this aooriach is, but from these examples by people much smarter than me, it seems this is a safe solution)

Examples:
https://github.com/ros-controls/ros_controllers/blob/1fd8d25707e289700743925c93644b4295b1c486/position_controllers/src/joint_group_position_controller.cpp#L42-L50
https://github.com/ros-controls/ros_controllers/blob/1fd8d25707e289700743925c93644b4295b1c486/velocity_controllers/src/joint_group_velocity_controller.cpp#L42-L46

matthew-reynolds: Actually, looking at this again, I think the right solution is to just use `initRT()`. In practical senses, it'll be pretty much the same thing, but with less overhead and also less hackey.

I think those examples I linked should probably be updated to `initRT()` as well. I'll look into that (#511).

matthew-reynolds (commented):
This change has no dependence on #503, could you please trim the branch down to just the relevant commit to make it easier to review and merge.

---

LGTM, especially because this is already what's done in most other position controllers. Set cmd=pos, reset PID.

wxmerkt:
Thank you for the review and feedback @matthew-reynolds. I updated the PR to use `initRT` and split it from the changes in #503.

matthew-reynolds (approved):
This looks great to me, ping @bmagyar for another set of eyes and potential merge

wxmerkt:
@bmagyar are we good to merge this?

bmagyar (approved):
Much appreciated guys, thank you!

bmagyar:
I've rebased this on the latest upstream which should have the CI fixed, merging when all turns green

bmagyar:
I still consider a diff_drive_controller failure green ;)

bmagyar:
Thanks for the fix @wxmerkt , merging

PR #505, opened 2020-08-26
Update ci to noetic

bmagyar:

PR #506, opened 2020-08-29
Fix test dependencies of effort_controllers package

mateus-amarante:
* Add missing dependency on joint_state_controller
* Remove unused transmission_interface

This is supposed to fix the errors detected in #505

### DIFF
--- a/effort_controllers/package.xml
+++ b/effort_controllers/package.xml
@@ -28,6 +28,7 @@
   <test_depend>rostest</test_depend>
   <test_depend>controller_manager</test_depend>
   <test_depend>hardware_interface</test_depend>
+  <test_depend>joint_state_controller</test_depend>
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
Thanks very much! This already fixes 2/3 ATM, will proceed with updating the CI to use noetic properly

mateus-amarante:
I was not able to reproduce the test failures of `diff_drive_controller`, even when I build and test the package in a fresh environment. Is there an easy way I can run CI tests locally? (I'm not so aware of the CI structure)

bmagyar:
You could try running that in a docker container but probably wouldn't get the kind of randomly limited resources that travis has. That test can be fixed by using a more reliable way to test topic output. Most of our tests are tolerant against limited resources when checking for topics, perhaps this one missed the update...

PR #515, opened 2020-09-13
Fix diff_drive_controller tests

mateus-amarante:
This is an attempt to fix an annoying hard-to-debug error present in jobs [998.2](https://travis-ci.org/github/ros-controls/ros_controllers/jobs/721105437), [1010.2](https://travis-ci.org/github/ros-controls/ros_controllers/jobs/724116882), [1010.3](https://travis-ci.org/github/ros-controls/ros_controllers/jobs/724116883).

The updates are:
* Move odom publisher to the end of the `init` method, so it is created after the optional publishers (twist commands and wheel states), then simplify `isControllerAlive` to check odom_pub only. The objective is to make the verification of the optional publishers more robust.
* Apply `waitForController` check (which considers `ros::ok()`) to all tests

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -301,8 +301,6 @@ namespace diff_drive_controller{
                           << ", left wheel radius "  << lwr
                           << ", right wheel radius " << rwr);
 
-    setOdomPubFields(root_nh, controller_nh);
-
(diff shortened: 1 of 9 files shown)


bmagyar:
Rebased for better CI

bmagyar (approved):
Thanks very much. I can confirm that this seems to fix the flaky test on the CI

Issue #513, opened 2020-09-13
Cleanup package.xml and CMakeLists.txt


mateus-amarante:
As suggested by @matthew-reynolds in https://github.com/ros-controls/ros_controllers/pull/510#issuecomment-687106430, I'm creating this "meta-issue" to map updates regarding cleanup of package.xml and CMakeLists.txt files of the ros_controllers' packages. The idea is to reproduce https://github.com/ros-controls/ros_control/pull/404. Generally, it includes:
- Clean dependencies (find missing/unnecessary dependencies)
- Apply consistent format to package.xml and CMakeListst.txt
- Fix major inconsistencies detected by `catkin_lint` (e.g. double `find_package`)

Checklist:
- [x] ackermann_steering_controller
- [x] diff_drive_controller
- [x] effort_controllers
- [x] force_torque_sensor_controller
- [x] forward_command_controller
- [x] four_wheel_steering_controller
- [x] gripper_action_controller
- [x] imu_sensor_controller
- [x] joint_state_controller
- [x] joint_trajectory_controller
- [x] position_controllers
- [x] ros_controllers
- [x] rqt_joint_trajectory_controller
- [x] velocity_controllers

PR #516, opened 2020-09-13
Clean CMakeLists.txt and package.xml files of diff_drive_controller package

mateus-amarante:
Apply updates suggested in #513:

Clean dependencies (find missing/unnecessary dependencies)
Apply consistent format to package.xml and CMakeListst.txt
Fix major inconsistencies detected by catkin_lint (e.g. double find_package)

### DIFF
--- a/ackermann_steering_controller/CMakeLists.txt
+++ b/ackermann_steering_controller/CMakeLists.txt
@@ -18,7 +18,7 @@ find_package(catkin REQUIRED COMPONENTS
   tf
 )
 
-find_package(Boost REQUIRED COMPONENTS accumulators bind function)
+find_package(Boost REQUIRED COMPONENTS headers)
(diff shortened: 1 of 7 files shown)


bmagyar:
Looks good to me! I've rebased it to have the benefits of the new CI updates

bmagyar:
boost details seem to be an overkill

mateus-amarante:
> boost details seem to be an overkill

Actually, there was a mistake there. All the header-only modules of Boost are gathered in the `headers` component. The last update fix it. If you still think it is an overkill, I can remove the dependency.

bmagyar (approved):
Having the boost stuff fully declared is nice, it should keep the package more future-proof, it's just not what I'm used to ;) 

Cheers for this, merging...

PR #521, opened 2020-10-03
Clean CMakeLists.txt and package.xml files of joint_state_controller package

mateus-amarante:
Apply updates suggested in #513:

* Clean dependencies (find missing/unnecessary dependencies)
* Apply consistent format to package.xml and CMakeListst.txt
* Fix major inconsistencies detected by catkin_lint (e.g. double find_package)

**OBS:** I noticed there are two misplaced files: [joint_state_controller.yaml](https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_state_controller/joint_state_controller.yaml) and [joint_state_controller.launch](https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_state_controller/joint_state_controller.launch). I think they must be removed (or at least moved to a launch/ and config/ folders). I didn't do it because they were already being installed. If you prefer to keep it to preserve the API, then we should also add an exec_depend on controller_manager.


bmagyar (approved):
Thank you!

bmagyar:
I'd keep those files but indeed we could move them into folders. The install locations can be kept so that installations won't look different. I found it useful in the past to keep simplistic files in these packages as examples people can copy and use. I know it's not fully proper but it worked in the past ;)

bmagyar:
Thanks a lot, merging!

PR #532, opened 2020-11-05
Fix null pointer checks in diff_drive_controller

DLu:
If you need to calculate wheel radius but not wheel separation, it is possible to pass null pointers to the `getWheelRadius` function because the checks for null pointers are contained within the `if lookup_wheel_separation` block.


bmagyar:
Nice catch! Do you have a urdf I could use in a test?

bmagyar (approved):
Created https://github.com/ros-controls/ros_controllers/issues/540

DLu:
Thanks for merging! Apologies for not getting you a test case sooner.

PR #536, opened 2020-11-23
joint_trajectory_controller: add time_from_start feedback

### LINKED ISSUE #535: JointTrajectoryController: time_from_start in state feedback messages is empty
The `JointTrajectoryController` doesn't fill the `time_from_start` field of the `JointTrajectoryPoint` in the `JointTrajectoryControllerState` message published via the state topic and action interface.

Filling out the `time_from_start` field would be useful to get some feedback on which part of the executed trajectory is currently running and thus how far we are along the trajectory path.

machinekoder:
The `current` state feedback gets the actual `time_from_start` and the interpolated time between the segments.
The `desired` state feedback gets the `time_from_start` of the beginning of the current segment.

I'm open for discussion what makes most sense, The current `time_from_start` is a must in my opinion. The actual start of the segment we are currently executing makes finding the source `JointTrajectoryPoint` in the trajectory easier. However, the desired states time is actually the same as the current one, as we are sampling with the current time.

An alternative option would be to make the end of the segment the desired `time_from_start` and also feed back the error `time_from_start` as the difference between the desired and current states time. But overall, that sounds confusing.

The current `time_from_start` is good enough for a client to find the source segment. However, requires more work, since we need to compare the timestamps.

fixes https://github.com/ros-controls/ros_controllers/issues/535

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -784,7 +784,6 @@ updateStates(const ros::Time& sample_time, const Trajectory* const traj)
       state_error_.time_from_start = desired_state_.time_from_start - current_state_.time_from_start;
     }
 
-    state_error_.time_from_start = desired_state_.time_from_start - current_state_.time_from_start;
   }
 }
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -716,8 +716,10 @@ publishState(const ros::Time& time)
       state_publisher_->msg_.desired.positions     = desired_state_.position;
       state_publisher_->msg_.desired.velocities    = desired_state_.velocity;
       state_publisher_->msg_.desired.accelerations = desired_state_.acceleration;
+      state_publisher_->msg_.desired.time_from_start = ros::Duration(desired_state_.time_from_start);
       state_publisher_->msg_.actual.positions      = current_state_.position;
       state_publisher_->msg_.actual.velocities     = current_state_.velocity;
+      state_publisher_->msg_.actual.time_from_start = ros::Duration(current_state_.time_from_start);
       state_publisher_->msg_.error.positions       = state_error_.position;
       state_publisher_->msg_.error.velocities      = state_error_.velocity;
 
@@ -760,7 +762,7 @@ updateStates(const ros::Time& sample_time, const Trajectory* const traj)
 
   for (unsigned int joint_index = 0; joint_index < getNumberOfJoints(); ++joint_index)
   {
-    sample( (*traj)[joint_index], sample_time.toSec(), desired_joint_state_);
+    const auto segment = sample( (*traj)[joint_index], sample_time.toSec(), desired_joint_state_);
 
     current_state_.position[joint_index] = joints_[joint_index].getPosition();
     current_state_.velocity[joint_index] = joints_[joint_index].getVelocity();
@@ -773,6 +775,13 @@ updateStates(const ros::Time& sample_time, const Trajectory* const traj)
     state_error_.position[joint_index] = angles::shortest_angular_distance(current_state_.position[joint_index],desired_joint_state_.position[0]);
     state_error_.velocity[joint_index] = desired_joint_state_.velocity[0] - current_state_.velocity[joint_index];
     state_error_.acceleration[joint_index] = 0.0;
+
+    if (joint_index == 0)
+    {
+      const auto time_from_start = segment->timeFromStart();
+      current_state_.time_from_start = sample_time.toSec() - segment->startTime() + time_from_start;
+      desired_state_.time_from_start = time_from_start;
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


machinekoder:
Note that I tested an alternative implementation using the `begin()` of the trajectory as time reference, since every segment has `startTime` and an `endTime` already. However, this doesn't work reliably, since the `JointTrajectoryController` combines different trajectories, so the first segment may be from another leftover trajectory and the one we currently execute. Therefore, we need to store the `timeFromStart` in the actual segment.

bmagyar:
This seems like a harmless change to me, filling these state fields with *anything* is better than the current approach. Since you are the first one who explained this use-case clearly so far, have you had a good experience with your proposed solution? I don't have a strong opinion on this

matthew-reynolds:
This looks good to me, especially the `current` time_from_start. I agree that anything is better than what the blank fields it publishes right now.

> An alternative option would be to make the end of the segment the desired time_from_start and also feed back the error time_from_start as the difference between the desired and current states time. **But overall, that sounds confusing.**

I agree, I think this offers more information but it's confusing and doesn't line up with the normal usage of the `JointTrajectoryPoint`. Your current approach for the `desired.time_from_start` is better since it aligns better with the intent of the msg (The `time_from_start` in the msg should match the time of the pos/vel/accel/efforts), even if it is a bit redundant.

I would propose also making `error.time_from_start = desired.time_from_start - current.time_from_start`. As you pointed out, this would always be 0, but it's nice in case someone down the road adapts the code in some way where current time != desired time in that message.

bmagyar:
I've rebased this branch to latest noetic and implemented your suggested change @matthew-reynolds . If you could approve, I'm happy to merge

bmagyar (approved):
@machinekoder thanks very much for this and sorry for the delay. These are very sensible changes

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -790,8 +801,10 @@ setActionFeedback()
   current_active_goal->preallocated_feedback_->desired.positions     = desired_state_.position;
   current_active_goal->preallocated_feedback_->desired.velocities    = desired_state_.velocity;
   current_active_goal->preallocated_feedback_->desired.accelerations = desired_state_.acceleration;
+  current_active_goal->preallocated_feedback_->desired.time_from_start = ros::Duration(desired_state_.time_from_start);
   current_active_goal->preallocated_feedback_->actual.positions      = current_state_.position;
   current_active_goal->preallocated_feedback_->actual.velocities     = current_state_.velocity;
+  current_active_goal->preallocated_feedback_->actual.time_from_start = ros::Duration(current_state_.time_from_start);
   current_active_goal->preallocated_feedback_->error.positions       = state_error_.position;
   current_active_goal->preallocated_feedback_->error.velocities      = state_error_.velocity;
```
matthew-reynolds, line 809: Let's stick the newly-calculated into the message too

```suggestion
  current_active_goal->preallocated_feedback_->error.velocities      = state_error_.velocity;
  current_active_goal->preallocated_feedback_->error.time_from_start = ros::Duration(state_error_.time_from_start);
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -773,6 +775,15 @@ updateStates(const ros::Time& sample_time, const Trajectory* const traj)
...
     state_error_.acceleration[joint_index] = 0.0;
+
+    if (joint_index == 0)
+    {
+      const auto time_from_start = segment->timeFromStart();
+      current_state_.time_from_start = sample_time.toSec() - segment->startTime() + time_from_start;
+      desired_state_.time_from_start = time_from_start;
+    }
+
+    state_error_.time_from_start = desired_state_.time_from_start - current_state_.time_from_start;
```
matthew-reynolds: I think this is better in the `if` statement above.

matthew-reynolds (changes requested):

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -773,6 +775,15 @@ updateStates(const ros::Time& sample_time, const Trajectory* const traj)
     state_error_.position[joint_index] = angles::shortest_angular_distance(current_state_.position[joint_index],desired_joint_state_.position[0]);
     state_error_.velocity[joint_index] = desired_joint_state_.velocity[0] - current_state_.velocity[joint_index];
     state_error_.acceleration[joint_index] = 0.0;
+
+    if (joint_index == 0)
+    {
+      const auto time_from_start = segment->timeFromStart();
+      current_state_.time_from_start = sample_time.toSec() - segment->startTime() + time_from_start;
+      desired_state_.time_from_start = time_from_start;
```
bmagyar, line 784:
```suggestion
      desired_state_.time_from_start = time_from_start;
      state_error_.time_from_start = desired_state_.time_from_start - current_state_.time_from_start;
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
bmagyar: good catch, added it there, removed from here!

```suggestion
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -773,6 +775,15 @@ updateStates(const ros::Time& sample_time, const Trajectory* const traj)
...
+      state_error_.time_from_start = desired_state_.time_from_start - current_state_.time_from_start;
+    }
+
```
bmagyar:
```suggestion
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -716,8 +716,10 @@ publishState(const ros::Time& time)
       state_publisher_->msg_.desired.positions     = desired_state_.position;
       state_publisher_->msg_.desired.velocities    = desired_state_.velocity;
       state_publisher_->msg_.desired.accelerations = desired_state_.acceleration;
+      state_publisher_->msg_.desired.time_from_start = ros::Duration(desired_state_.time_from_start);
       state_publisher_->msg_.actual.positions      = current_state_.position;
       state_publisher_->msg_.actual.velocities     = current_state_.velocity;
+      state_publisher_->msg_.actual.time_from_start = ros::Duration(current_state_.time_from_start);
       state_publisher_->msg_.error.positions       = state_error_.position;
       state_publisher_->msg_.error.velocities      = state_error_.velocity;
```
matthew-reynolds, line 724:
```suggestion
      state_publisher_->msg_.error.velocities      = state_error_.velocity;
      state_publisher_->msg_.error.time_from_start = ros::Duration(state_error_.time_from_start);
```

matthew-reynolds (approved):
Whoops, I missed this one too in my last review.

LGTM, pending this 1 addition :)

matthew-reynolds (approved):

bmagyar:
Thanks again guys, merging...

PR #542, opened 2020-12-06
Add test for #532

caioaamaral:
**Description:**
  - This adds the rostest `diff_drive_get_wheel_radius_fail.test` that covers the problem solved by  #532.

**Without the patch:** 
  -  Test would fail due a rostest timeout since DiffDriveController crashes when tries to dereference a nullpointer.

**Now:**
  - Test pass since it accomplishes the requirements stated by [diff_drive_fail_test.cpp](https://github.com/ros-controls/ros_controllers/blob/noetic-devel/diff_drive_controller/test/diff_drive_fail_test.cpp)


bmagyar:
Only a flaky test failing, thanks for this! Merging...

PR #546, opened 2020-12-30
Add codecov bot

matthew-reynolds:
Draft until next release of `setup-ros`

I mainly opened this draft early since I wanted to see if the GH actions were working on PRs :smile: Looks like things are good!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,10 +30,10 @@ jobs:
     strategy:
       fail-fast: false
     steps:
-      - uses: ros-tooling/setup-ros@0.1.1
+      - uses: ros-tooling/setup-ros@v0.1
         with:
           required-ros-distributions: noetic
-      - uses: ros-tooling/action-ros-ci@0.1.0
+      - uses: ros-tooling/action-ros-ci@v0.1
         with:
           target-ros1-distro: noetic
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -9,16 +9,59 @@ on:
...
+            gripper_action_controller
+            imu_sensor_controller
+            joint_state_controller
+            joint_trajectory_controller
+            position_controllers
+            ros_controllers
+            rqt_joint_trajectory_controller
+            velocity_controllers
+          vcs-repo-file-url: |
+            https://raw.githubusercontent.com/ros-controls/ros_control/noetic-devel/ros_control.rosinstall
```
matthew-reynolds, line 56: Unlike in the industrial_ci `UPSTREAM_WORKSPACE`, we don't need to exclude ros_controllers itself, since `action-ros-ci` does this automatically (See [here](https://github.com/ros-tooling/action-ros-ci/blob/2d437cb171c1b95b338c22ff67a4e8a2b87394a5/src/action-ros-ci.ts#L208)).

Unfortunately, we can't use the `$ROS_DISTRO` macro here since that's provided by industrial_ci, so instead we just specify the `noetic-devel` branch directly. No big deal.

PR #551, opened 2021-01-07
Update ackermann_steering_controller.cpp

OCILY:
include ros.h


bmagyar:
There are a couple of include in the header file of this controller that include ros.h from multiple sources. This is not required but thank you for caring.

PR #552, opened 2021-01-18
User/wmmc88/fix diff drive controller nan check

wmmc88:
diff_drive controller was checking the wrong member for finite values. The existing test was written such that it didn't test each branch in the if statement so it didn't catch the typo.

### DIFF
--- a/diff_drive_controller/test/diff_drive_nan_test.cpp
+++ b/diff_drive_controller/test/diff_drive_nan_test.cpp
@@ -91,6 +91,16 @@ TEST_F(DiffDriveControllerTest, testNaNCmd) {
   for (int i = 0; i < 10; ++i) {
     geometry_msgs::Twist cmd_vel;
     cmd_vel.linear.x = NAN;
+    cmd_vel.angular.z = 0.0;
+    publish(cmd_vel);
+    geometry_msgs::TwistStamped odom_msg = getLastCmdVelOut();
+    EXPECT_FALSE(std::isnan(odom_msg.twist.linear.x));
+    EXPECT_FALSE(std::isnan(odom_msg.twist.angular.z));
+    ros::Duration(0.1).sleep();
+  }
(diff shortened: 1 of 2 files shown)


diff_drive_controller/test/diff_drive_nan_test.cpp
```diff
@@ -91,6 +91,16 @@ TEST_F(DiffDriveControllerTest, testNaNCmd) {
...
+    cmd_vel.angular.z = 0.0;
+    publish(cmd_vel);
+    geometry_msgs::TwistStamped odom_msg = getLastCmdVelOut();
+    EXPECT_FALSE(std::isnan(odom_msg.twist.linear.x));
+    EXPECT_FALSE(std::isnan(odom_msg.twist.angular.z));
+    ros::Duration(0.1).sleep();
+  }
+  for (int i = 0; i < 10; ++i) {
+    geometry_msgs::Twist cmd_vel;
+    cmd_vel.linear.x = 0.0;
```
wmmc88, line 103: something i noticed when testing the fix for the test was that it still doesn't catch the error unless the other test in this file is commented. Anyone have any ideas about this? Is rostest reusing the same master and running the tests in parallel or something?

wmmc88:
unrelated test in `ackermann_steering_controller` seems to be failing

diff_drive_controller/test/diff_drive_nan_test.cpp
matthew-reynolds: As we discussed offline, the tests within a test suite are run sequentially on the same master. However, each `.test` is run with its own master.

matthew-reynolds (approved):
This LGTM!

> unrelated test in ackermann_steering_controller seems to be failing

Guess I should hurry up with those flakey fixes I'm working on...

bmagyar:
Ingoring the tests, thanks for the contribution!

PR #560, opened 2021-03-29
Check if max_allowed_effort is set before applying limits

ikalevatykh:
By default, the max_effort value is zero, and it does not allow to the gripper to move in the current implementation.

### DIFF
--- a/gripper_action_controller/include/gripper_action_controller/hardware_interface_adapter.h
+++ b/gripper_action_controller/include/gripper_action_controller/hardware_interface_adapter.h
@@ -175,7 +175,10 @@ public:
 
     // Update PIDs
       double command = pid_->computeCommand(error_position, error_velocity, period);
-      command = std::min<double>(fabs(max_allowed_effort), std::max<double>(-fabs(max_allowed_effort), command));
+      if (max_allowed_effort)
+      {
+        command = std::min<double>(fabs(max_allowed_effort), std::max<double>(-fabs(max_allowed_effort), command));
+      }
       (*joint_handle_ptr_).setCommand(command);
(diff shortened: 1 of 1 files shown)


matthew-reynolds:
I'm not sure I understand the intention behind this change.

The `default_max_effort_` is used whenever the controller is started but hasn't yet received a command, or when the goal is finished. Here, a max effort of 0 usually makes sense, since we don't want the gripper to move. (Alternatively, a non-zero default `max_effort` could be provided so that the gripper actively maintains position.)

Whenever we _do_ want the gripper to move, we send it a command with a desired position _and max effort_. Since the command itself contains a max effort, the `default_max_effort_` isn't used. It's the responsibility of whatever is sending the command to specify the max effort for the duration of the command.

---

(Disclaimer: I haven't ever personally used the gripper controller)

ikalevatykh:
Thanks, @matthew-reynolds 

Yes, I understand. The question is, can I set max_effort to zero in the GripperCommand to skip the effort clipping.

matthew-reynolds:
I don't think we should add a special case for 0 here. This is a mature controller and I would not be surprised if people out there have valid uses for max_effort=0 (eg. maybe going limp for safety purposes, but still wanting a non-zero default max effort to hold position between goals).

I think you're best off simply commanding some appropriately large limit. (Or, if for some reason it's unknown, an arbitrarily large limit). That way we aren't breaking anyone's existing behaviour, or introducing discontinuities in the controller by jumping from a low limit to infinite limit as `max_limit` approaches zero.

ikalevatykh:
OK, I agree

Issue #561, opened 2021-04-09
[Kinetic diff_drive_controller] Publishes backward driving command to wheels even with zero input command


wienans:
Hi Guys,

i faced a really weird issue of the diff_drive_controller in the kinetic version. And as i think it is safety related i post this issue even if kinetic runs into EOL this month. 
After Startup of the robot the robot was driving backwards and left for maybe a second starting with -1.5 m/s in linear and angular direction. This ramped down over time and every second message is a zero velocity message. Here you can see a rqt Plot of this behavior. This topic is the cmd_vel_out debug topic.
![grafik](https://user-images.githubusercontent.com/40465543/114139613-90ede100-990f-11eb-82c2-28f0bada4179.png)
As we don't publish a zero velocity command while standing and it only appeared rarely after startup i first thought it was something going crazy internally because some variable is may be not defined correctly and it graps some random data from the RAM. So i published contentious zeros while standing and the issue occured again. The biggest problem with it, it is not really reproducable, i didn't faced the issue after restarting the robot after it occured once at the start of the day. And it only reappeared at some days and only at the first startup of the day.

bmagyar:
It'd be nice if you could get to the bottom of this. I'd add logs for the
input joint states too, perhaps your encoder values were spinning like
crazy and this is to compensate?

wienans:
Hey,
i didn’t do a deep dive into the code, but I found this lines https://github.com/ros-controls/ros_controllers/blob/45486f82a509a1e8af588d23844ba01e7685cf96/diff_drive_controller/src/diff_drive_controller.cpp#L424
There the command is read out and later published to the debug topic. So I thought it is only forward Control. Subscribe to input topic save in real-time handler read from real time handler publish debug velocity and then set wheel velocity’s. So I would assume there is no feedback from the wheels or any other system which would course a change of cmd_vel_out

matthew-reynolds:
Hmm, that sorta looks like the controller is timing out the command.

 - Is it possible you're specifying a very short command timeout?  ("cmd_vel_timeout" param)
 - Are you passing in the correct time into the controller_manager update? You need to use `ros::Time::now()` rather than any other time source in order to match the controller's internal timeout calculation.
 
 Alternatively, is it possible that you're passing in an inconsistent update period to the controller_manager update?
 
 I think those are the only two factors that could affect the command between the `readFromRT()` and the publisher, but I haven't dug deep.

wienans:
[...]

PR #563, opened 2021-05-12
Add guard to dt negative case and break the controller for safety

alcantara09:
Hi,

This PR adds a guard to the case when the time step on the diff_drive_controller is computed as negative due to some error coming from the hardware interface, as in https://github.com/ros-controls/ros_controllers/issues/561

At my company we are experiencing the same backward movement behavior but less consistent and less predicable. This safe guards the controller to this type of error, what can be common given we are also experiencing the same issue.

### DIFF
--- a/ackermann_steering_controller/src/ackermann_steering_controller.cpp
+++ b/ackermann_steering_controller/src/ackermann_steering_controller.cpp
@@ -322,7 +322,7 @@ namespace ackermann_steering_controller{
 
     // Limit velocities and accelerations:
     const double cmd_dt = (period.toSec() > 0) ? period.toSec() : 0.0;
-    if(cmd_dt < 0)
+    if(period.toSec() < 0)
     {
       ROS_ERROR("Invalid time interval, delta time cannot be negative");
     }
--- a/ackermann_steering_controller/src/ackermann_steering_controller.cpp
+++ b/ackermann_steering_controller/src/ackermann_steering_controller.cpp
@@ -321,7 +321,7 @@ namespace ackermann_steering_controller{
     }
 
     // Limit velocities and accelerations:
-    const double cmd_dt(period.toSec());
+    const double cmd_dt = (period.toSec() > 0) ? period.toSec() : 0.0;
     if(cmd_dt < 0)
     {
       ROS_ERROR("Invalid time interval, delta time cannot be negative");
--- a/ackermann_steering_controller/src/ackermann_steering_controller.cpp
+++ b/ackermann_steering_controller/src/ackermann_steering_controller.cpp
@@ -322,6 +322,12 @@ namespace ackermann_steering_controller{
 
     // Limit velocities and accelerations:
     const double cmd_dt(period.toSec());
+    if(cmd_dt < 0)
+    {
+      ROS_ERROR("Invalid time interval, delta time cannot be negative");
+      curr_cmd.lin = 0.0;
+      curr_cmd.ang = 0.0;
+    }
 
     limiter_lin_.limit(curr_cmd.lin, last0_cmd_.lin, last1_cmd_.lin, cmd_dt);
     limiter_ang_.limit(curr_cmd.ang, last0_cmd_.ang, last1_cmd_.ang, cmd_dt);
--- a/ackermann_steering_controller/src/ackermann_steering_controller.cpp
+++ b/ackermann_steering_controller/src/ackermann_steering_controller.cpp
@@ -325,8 +325,6 @@ namespace ackermann_steering_controller{
     if(cmd_dt < 0)
     {
       ROS_ERROR("Invalid time interval, delta time cannot be negative");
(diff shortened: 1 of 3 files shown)


matthew-reynolds (changes requested):
We should guard on `period` being negative, which corresponds to `cmd_dt` rather than `dt`, Could you also duplicate this over to `ackermann_steering_controller` and `four_wheel_steering_controller`? Those controllers all use the period in their `SpeedLimiters`, which is where this particularly nasty behaviour is coming from.

The other controllers only use the period for updating PIDs, which will still have some odd effects but shouldn't be nearly as big of a deal.

I will add, though, that if you're seeing this issue then the real solution is to fix your `ControllerManager::Update()` to use a monotonic time source. Adding warnings here is a bit of a fall back and diagnostic, it shouldn't be relied on to provide good behaviour. Passing in a negative period isn't something the framework is designed to handle, since it makes no sense.

alcantara09:
> We should guard on `period` being negative, which corresponds to `cmd_dt` rather than `dt`, Could you also duplicate this over to `ackermann_steering_controller` and `four_wheel_steering_controller`? Those controllers all use the period in their `SpeedLimiters`, which is where this particularly nasty behaviour is coming from.
> 
> The other controllers only use the period for updating PIDs, which will still have some odd effects but shouldn't be nearly as big of a deal.
> 

Ok, I will do the following changes, no problem.

> I will add, though, that if you're seeing this issue then the real solution is to fix your `ControllerManager::Update()` to use a monotonic time source. Adding warnings here is a bit of a fall back and diagnostic, it shouldn't be relied on to provide good behaviour. Passing in a negative period isn't something the framework is designed to handle, since it makes no sense.

You are absolutely correct, we are debugging our update to correct the problem. But given we were not the only ones to hit this issue, I thought it would be useful to have the guard

alcantara09:
Hi,

I did the asked changes. 

However I have one question still. I was getting a lot of negative values for `dt`. What is the meaning of this? should we add a `std::abs` to check the timeout with a positive value?

alcantara09:
Hi,

I was just testing the entire chain and I just realized that setting curr_cmd to zero doesn't guard against the bug because it will be still inserted on the limiter. 

I see two options then:
1. We can set curr_cmd to zero and skip the limiter step
2. We can set curr_cmd to zero and set cmd_dt to a positive value (The question then would be which value)

An alternative would be to throw an exception

What is your opinion on that? 

(The unit tests are running normally locally, I do not understand why it failed on the pipeline)

matthew-reynolds:
I think we probably want to just set the `cmd_dt` to 0. That should force the acceleration and jerk to be 0 - In other words, maintain the same vel as the prev update. Not perfect, but I think that's the safest solution.

Skipping the limiter step opens up a hole for a dangerous command to slip through, which I don't think is acceptable.

alcantara09:
Fair enough, I did the changes

ackermann_steering_controller/src/ackermann_steering_controller.cpp
```diff
@@ -321,7 +321,13 @@ namespace ackermann_steering_controller{
...
 
     // Limit velocities and accelerations:
-    const double cmd_dt(period.toSec());
+    const double cmd_dt = (period.toSec() > 0) ? period.toSec() : 0.0;
+    if(cmd_dt < 0)
+    {
+      ROS_ERROR("Invalid time interval, delta time cannot be negative");
+      curr_cmd.lin = 0.0;
+      curr_cmd.ang = 0.0;
+    }
```
matthew-reynolds: Your proposed change won't issue the warning, and we shouldn't touch the curr_cmd.

Something like this makes more sense.

```suggestion
    double cmd_dt(period.toSec());
    if (cmd_dt < 0.0)
    {
      ROS_ERROR("Invalid time interval, delta time cannot be negative");
      cmd_dt = 0.0;
    }
```

matthew-reynolds (changes requested):
Since we don't have tests to verify this behaviour, please test the changes locally to ensure the PR has the desired effect. Your last commit doesn't appear to satisfy the intended behaviour.

Please also retarget this PR to the `noetic-devel` branch. It can then be backported from there to Melodic.

bmagyar:
Any update on this @alcantara09 ? I'd like to merge a working solution if you have one. The current errors printed are not bad to start but if you could set the values to 0.0 on top of reporting the error that'd make things a little safer.

alcantara09:
Hi @bmagyar and @matthew-reynolds,

I am forcing cmd_dt to zero when we have a negative period. 
This forces all jerk and acceleration to be zero and then we are forwarding last cmd as current cmd, as it was agreed as expected behavior. Only on the jerk calculation, `v = dv0 + da`, that the contributtion of the previous velocities still can alter the current cmd. What I do not believe should be set to zero.

I believe the code is in conformity with the issues raised by you, otherwise could you elaborate a little more the aspect that I am still missing so I can change?

PR #565, opened 2021-05-14
[joint_state_controller] Allow specification of joints or set specific order

wxmerkt:
Introduces a 'joints' parameter similar to other controllers in order to
be able to specify a subset of joints or a specific order of joints.
This is useful when the HardwareInterface exposes joints in an order
that is not desired or if certain joints are not meant to be published
by a controller. If not provided, the controller publishes all joints
and preserves backwards-compatibility.

### DIFF
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -53,7 +53,7 @@ namespace joint_state_controller
       ROS_INFO_STREAM("Joints parameter specified, publishing specified joints in desired order.");
     } else {
       // get all joint names from the hardware interface
-      std::vector<std::string> joint_names = hw->getNames();
+      joint_names = hw->getNames();
     }
 
     num_hw_joints_ = joint_names.size();
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -42,8 +42,20 @@ namespace joint_state_controller
                                   ros::NodeHandle&                         root_nh,
                                   ros::NodeHandle&                         controller_nh)
   {
-    // get all joint names from the hardware interface
-    const std::vector<std::string>& joint_names = hw->getNames();
+    // List of joints to be published
+    std::vector<std::string> joint_names;
+
+    // Get list of joints: This allows specifying a desired order, or
+    // alternatively, only publish states for a subset of joints. If the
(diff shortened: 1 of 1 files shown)


joint_state_controller/src/joint_state_controller.cpp
```diff
@@ -42,8 +42,20 @@ namespace joint_state_controller
...
+
+    // Get list of joints: This allows specifying a desired order, or
+    // alternatively, only publish states for a subset of joints. If the
+    // parameter is not set, all joint states will be published in the order
+    // specified by the hardware interface.
+    if (controller_nh.getParam("joints", joint_names)) {
+      ROS_INFO_STREAM("Joints parameter specified, publishing specified joints in desired order.");
+    } else {
+      // get all joint names from the hardware interface
+      std::vector<std::string> joint_names = hw->getNames();
```
matthew-reynolds:
```suggestion
      joint_names = hw->getNames();
```

matthew-reynolds (changes requested):
This seems like a reasonable feature.

Personally I don't see a lot of value in this (Why would you care about the order of the joints? Is the small overhead of transporting all the joint states instead of a subset really an issue?), but I also see no downside to giving the option to users, so LGTM pending requested changes.

wxmerkt:
Thanks for spotting the error!

> Why would you care about the order of the joints? 

Unfortunately not all downstream consumers of `joint_states` correctly set the order of the vector based on the joint names and have implicit assumptions on joint orders (including when URDF parsers order joints differently).

> Is the small overhead of transporting all the joint states instead of a subset really an issue?

Some downstream use cases we encountered don't select the required value from the vector, so having the option to have a subset e.g. under a different topic is a suitable workaround without republishing.

matthew-reynolds (approved):
LGTM. Shame we have to add upstream workarounds for poorly behaving downstream packages, but it's such a small and easy feature that we may as well.

Thanks!

wxmerkt:
@matthew-reynolds I added the suggested fix. Is this good to merge?

matthew-reynolds:
Let's ping @bmagyar

wxmerkt:
@bmagyar could you please take a look at this?

bmagyar (approved):

bmagyar:
Thanks guys, looks good! I will try to put the release out today

bmagyar:
https://github.com/ros/rosdistro/pull/29881

PR #570, opened 2021-07-29
Added acceleration, velocity and jerk as dynamic parameters to diff_drive_controller

ivan140:
Hi,

I experience cases on my diff drive robot, where it is necessary to dynamically change the acceleration, velocity and jerk parameters of the diff drive controller (i.e. when the robot is carrying some extra weight). Therefore I added them to the dynamic parameters.

Would be nice if you find the time to review and merge, if possible.


pradhanmeharraj:
WANT TO WORK ON IT

PR #573, opened 2021-09-12
Add <?xml version=1.0?> to every .launch and .test file

lucasw:


bmagyar:
I guess this is ok but may I ask why? :)

lucasw:
vim and maybe other text editors benefit from it for syntax highlighting without setting up special rules for each file type.

bmagyar (approved):

bmagyar:
thanks!

PR #588, opened 2022-02-22
changes for modern system environment (Ubuntu 22.04+)

v4hn:
These changes are needed to compile&use the packages sucessfully for [ros-o](https://github.com/ros-o) on Ubuntu 22.04 and upwards.

This PR subsumes #587 .

### DIFF
--- a/ackermann_steering_controller/CMakeLists.txt
+++ b/ackermann_steering_controller/CMakeLists.txt
@@ -1,11 +1,6 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(ackermann_steering_controller)
 
-# Default to C++14
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
-endif()
-
 # Load catkin and all dependencies required for this package
 find_package(catkin REQUIRED COMPONENTS
   controller_interface
--- a/diff_drive_controller/CMakeLists.txt
+++ b/diff_drive_controller/CMakeLists.txt
@@ -1,11 +1,6 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(diff_drive_controller)
 
-# Default to C++14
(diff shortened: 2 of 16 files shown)


ackermann_steering_controller/CMakeLists.txt
```diff
@@ -1,11 +1,6 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(ackermann_steering_controller)
 
-# Default to C++14
-if(NOT CMAKE_CXX_STANDARD)
```
wxmerkt, line 5: @v4hn @jspricke Very excited about the ROS-O effort. Since I've seen a few of these PRs around, for my understanding, if the `set(CMAKE_CXX_STANDARD` is guarded with `if(NOT CMAKE_CXX_STANDARD)`, should it still be removed?

jspricke: Yes, `CMAKE_CXX_STANDARD` is not set by default and I recommend not doing so. Note that C++14 is the default in Ubuntu 20 anyhow so all this code does is forcing an older standard for newer compilers ;).

wxmerkt: Thank you, all clear. I was wondering about older operating system releases as well (as in other projects we don't branch for every release / try to keep one branch for as many distros as possible).

jspricke: C++14 "is the default in GCC 6.1 up until GCC 10 (including)"
https://gcc.gnu.org/projects/cxx-status.html#cxx14

Ubuntu 18 has GCC 7.4:

https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=gcc

ROS Melodic targets C++14:

https://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023

bmagyar (approved):
Thanks!

PR #589, opened 2022-03-06
Updated to use nodehandle in urdf parsing, which allows to use namespace for robot_description.

aarsht7:
Refer the ROS question where I was facing an error with the namespaces while working with the group position controller.
https://answers.ros.org/question/397126/robot_description-not-found-on-parameter-server-while-launching-the-controller/


destogl (approved):
Sounds reasonable.

PR #594, opened 2022-05-05
Increase spin box to six decimals

Levi-Armstrong:
Currently the spin box only provides 2 decimals which does not seem to be enough especially if you are dealing with linear axes(1 cm resolution).


Levi-Armstrong:
More information on why this was needed: I am using this tool to preform a multiple point tool center point calibration. This involve jogging the robot to the same point but from different orientation. Given the current resolution it is not possible at least for my application.

bmagyar (approved):

bmagyar:
I'll try to get a release out asap, thx for the PR

PR #595, opened 2022-05-07
use boost::placeholders::_1/_2 in remaining instances

lucasw:
Also include boost/bind/bind.hpp instead of boost/bind.hpp, eliminated unnecessary bind.hpp include.  Follows #588


bmagyar:
How about removing boost bind altogether?

bmagyar (approved):
thanks!

PR #605, opened 2022-07-27
Store trajectory segment being executed and store in feedback message

Levi-Armstrong:
This is to open up discussions related to this [PR](https://github.com/ros-controls/control_msgs/pull/67). Is there any possibility to get this into the ROS1 version


destogl (approved):
I would say this is fine since no functionality is changed, but just extended.

PR #608, opened 2022-10-22
Install headers of ackermann_steering_controller

peci1:
diff_drive_controller does it, so there's no reason for ackermann to not do it. I use the headers in a library that would like to utilize the odometry classes defined by these controllers.

### DIFF
--- a/ackermann_steering_controller/CMakeLists.txt
+++ b/ackermann_steering_controller/CMakeLists.txt
@@ -126,6 +126,11 @@ endif()
 ## Install ##
 #############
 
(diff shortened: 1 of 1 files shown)


bmagyar (approved):
Don't see why not...

peci1:
Thanks. Will you please consider Melodic backport?

bmagyar:
let's see!

bmagyar:
@mergifyio backport melodic

bmagyar:
@mergifyio backport melodic-devel

PR #619, opened 2023-08-25
joint_state_controller: Clear joint_state_ in init() before assignment

onionsflying:
It seems like `joint_state_` should be cleared before assignment. Because the `init` function will be called again when the controller is reloaded.


bmagyar (approved):
Thank you!

PR #623, opened 2023-11-13
JointTrajectory: Don't use shortest_angular_distance for linear joints

### LINKED ISSUE #432: Does JointTrajectoryController assume that joints are rotational?
Does `JointTrajectoryController` assume that joints are rotational - continuous, revolute - instead of linear - prismatic -? I cannot find any such assumption documented on the wiki or in the code, but maybe I missed something.
I see the [`angle_wraparound_` member](https://github.com/ros-controls/ros_controllers/blob/melodic-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.h#L182), apparently used to differentiate continuous from revolute joints.
The calculation for the `state_joint_error_` on position at https://github.com/ros-controls/ros_controllers/blob/melodic-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L390 and that for the `state_error` on position at https://github.com/ros-controls/ros_controllers/blob/melodic-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L394 use [`angles::shortest_angular_distance`](http://docs.ros.org/indigo/api/angles/html/namespaceangles.html#a83debbedd331511aa914f393db576818) which will always give you a result in [-pi;pi].
It would be confusing to find that the position error on my linear joint mysteriously doesn't want to go beyond [-3.14meters;3.14meters], although a linear joint with such a length is probably not very common...

bmagyar:
Fixes https://github.com/ros-controls/ros_controllers/issues/432

Port #591 to Noetic

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
@@ -254,6 +254,7 @@ bool JointTrajectoryController<SegmentImpl, HardwareInterface>::init(HardwareInt
   // Initialize members
   joints_.resize(n_joints);
   angle_wraparound_.resize(n_joints);
+  is_linear_.resize(n_joints);
   for (unsigned int i = 0; i < n_joints; ++i)
   {
     // Joint handle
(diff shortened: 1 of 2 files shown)


bmagyar:
@steinmn could you please nod in agreement? ;)

steinmn (approved):
Nod :)

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h
```diff
@@ -779,6 +783,17 @@ updateStates(const ros::Time& sample_time, const Trajectory* const traj)
...
+      state_joint_error_.position[0] = desired_joint_state_.position[0] - current_state_.position[joint_index];
+    }
+    else
+    {
+      state_joint_error_.position[0] = angles::shortest_angular_distance(current_state_.position[joint_index],desired_joint_state_.position[0]);
+    }
+    state_joint_error_.velocity[0] = desired_joint_state_.velocity[0] - current_state_.velocity[joint_index];
+    state_joint_error_.acceleration[0] = 0.0;
+
     state_error_.position[joint_index] = angles::shortest_angular_distance(current_state_.position[joint_index],desired_joint_state_.position[0]);
```
steinmn, line 797:
```suggestion
    if (is_linear_[joint_index])
    {
      state_error_.position[joint_index] = desired_joint_state_.position[0] - current_state_.position[joint_index];
    }
    else
    {
      state_error_.position[joint_index] = angles::shortest_angular_distance(current_state_.position[joint_index],desired_joint_state_.position[0]);
    }
```

steinmn (changes requested):
On second thought, something doesn't look quite right here

PR #582, opened 2021-11-11
[velocity_controllers] Add timeout to JointGroupVelocityController

wxmerkt:
Relates to #363

Issue #363, opened 2018-10-15
No timeout in JointGroupVelocityController


AndyZe:
From testing in Gazebo, it seems that JointGroupVelocityController never stops when commands go stale. For example, say it received a velocity message 200 seconds ago. It will continue moving according to this last command, until something breaks/cannot move farther.

It seems like a timeout should be added to handle this case. Please inform me if a PR would be welcomed to add this. And also, how should the timeout period be established? I guess it should be one of the parameters in the controller.yaml file?

AndyZe:
Here's a video to show what I mean. I stop sending commands to the controller after a few seconds but the arm keeps drifting upward.

[https://drive.google.com/file/d/1vXsQ-oxcefxUcGOVdtnXHXSZ5CW9eB4g/view?usp=sharing](https://drive.google.com/file/d/1vXsQ-oxcefxUcGOVdtnXHXSZ5CW9eB4g/view?usp=sharing)

wxmerkt:
I agree with @AndyZe - we should have a time-out for velocity forward command controllers. I am happy to add it - would a PR be accepted?

Proposal to amend `velocity_controllers/JointGroupVelocityController`:
- Define the timeout (double, in seconds) in the forward_command_controller with a default of -1. If timeout < 0, it's not being applied (e.g. for position). 
- When timeout is a positive value, set the command to zero if no command has been received in the given timeframe. Zero is acceptable for velocity, but not for effort control (but that is a separate issue)

bmagyar:
Very welcome!

PR #625, opened 2023-11-24
JointGroupVelocityController: stop motion if no commands received

rhaschke:
As a safety measure, this PR adapts the `JointGroupVelocityController` to reset control velocities to zero if no new command was received within a configurable interval (`watchdog_timeout`). If the steering process crashes or just finishes without resetting the velocity itself, the robot would otherwise just continue moving with the last speed commanded (and crash into something probably).
I have chosen a rather large default timeout (1s) to limit interferences with existing code.


wxmerkt:
Related issue and PR: #363 / #582

rhaschke:
It's a pity that your initial PR wasn't merged yet, @wxmerkt. @bmagyar, how should we proceed with the two PRs addressing the same original issue?

rhaschke:
Friendly ping, @bmagyar

PR #629, opened 2023-12-24
Fix joint state controller exception starting with 0 time

### LINKED ISSUE #628: Joint state controller throws exception when start time is less than update rate
The joint state controller in `joint_state_controller.cpp` makes an assumption in its `JointStateController::starting(const ros::Time& time)` method that the time value passed to this function is always greater than its publish interval, which is set in Hz.

```
last_publish_time_ = time - ros::Duration(1.001/publish_rate_);
```

If the `time` passed into this function is less than `0.02` seconds for a `50` Hz publish rate, for example, this will result in `last_publish_time` being negative after the subtraction, which would throw a `std::runtime_error` exception causing the joint state controller to die:

```
terminate called after throwing an instance of 'std::runtime_error'
what():  Time is out of dual 32-bit range
```

The exception is thrown because the time in ROS is represented as an unsigned value, which would cause it wrap to the maximum possible value of the number type if a negative value was set.

*Proposed fix*

If we calculate the publish interval...

```
auto interval = ros::Duration(1.001/publish_rate_);
```

...and compare to `time` before subtracting...

```
if (time.toSec() > interval.toSec())
```

Then we can set the `last_publish_time` to `0` if the interval is larger than `time`, or set to their difference as before if the `time` is greater than `interval`. I see that Dave Coleman also recommends setting last update time to this in ROS Controllers Boilerplate:

```
// Get current time for use with first update
 clock_gettime(CLOCK_MONOTONIC, &last_time_);
```

I don't know what this means yet, will investigate if I should do this here instead of simply setting to `0`.

*Testing*

I will submit a pull request for this issue, and test the joint state controller with Panda arm used in MoveIt tutorials to ensure everything is working as expected, and test results are not polluted by my own robot configuration which could be broken.

Hopefully setting last publish time to zero will not break anything.

You may have unit tests as well, I will see if I can add one.

01binary:
Fixes #628, crash in joint state controller when given start time less than the publish interval.

## Reproduction steps:

- Build joint state controller from source (the published package does not have the bug yet)
- Set publish rate less than 100 Hz (i.e. 50 Hz should give you 0.02 seconds publish interval)
- Start the joint state controller (arbitrary start time will be around 0.01 seconds)
- Note that if you set the publish rate to 100Hz that would give you 0.01 seconds publish interval, which is not less than start time, so the bug would not be exposed.

## Testing

I haven't updated unit tests for joint state controller yet, but the functionality was tested and appears to work:

- The issue was fixed
- The problem the bug was trying to solve remains fixed: joint state controller will update immediately on the first update as the author apparently desired, and then wait the publish interval on subsequent updates

### DIFF
--- a/joint_state_controller/include/joint_state_controller/joint_state_controller.h
+++ b/joint_state_controller/include/joint_state_controller/joint_state_controller.h
@@ -93,7 +93,6 @@ private:
   ros::Time last_publish_time_;
   double publish_rate_;
   unsigned int num_hw_joints_; ///< Number of joints present in the JointStateInterface, excluding extra joints
-  bool pub_time_initialized_;
 
   void addExtraJoints(const ros::NodeHandle& nh, sensor_msgs::JointState& msg);
 };
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -80,26 +80,17 @@ namespace joint_state_controller
     }
     addExtraJoints(controller_nh, realtime_pub_->msg_);
 
-    pub_time_initialized_=false;
-
     return true;
   }
 
   void JointStateController::starting(const ros::Time& time)
   {
-    // initialize time exactly once, to maintain publish rate through controller resets
-    if (!pub_time_initialized_)
-    {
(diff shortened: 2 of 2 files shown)


01binary:
This is now a duplicate of #630 .
I failed to get unit tests passing, and the issue has been solved in the other PR.

PR #630, opened 2024-02-27
Fix uncaught exception in joint_state_controller

rhaschke:
Using sim time in Gazebo, the starting time is close to zero, such that `time - ros::Duration(1/publish_rate_)` becomes negative. This causes an exception to be thrown: "Time is out of dual 32-bit range".

### DIFF
--- a/joint_state_controller/src/joint_state_controller.cpp
+++ b/joint_state_controller/src/joint_state_controller.cpp
@@ -90,7 +90,11 @@ namespace joint_state_controller
     // initialize time exactly once, to maintain publish rate through controller resets
     if (!pub_time_initialized_)
     {
-      last_publish_time_ = time - ros::Duration(1.001/publish_rate_); //ensure publish on first cycle
+      try {
+        last_publish_time_ = time - ros::Duration(1.001/publish_rate_); //ensure publish on first cycle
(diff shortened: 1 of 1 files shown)


bmagyar (approved):

bmagyar:
Thank you!

Would you like a release of this or happy with it being merged for now?

rhaschke:
A merge would be fine for now. Sorry for pushing broken code initially. It should work now.

01binary:
Totally forgot about this. I also have a fix in https://github.com/ros-controls/ros_controllers/pull/629, but the unit tests are failing and I couldn't find a way to fix them.

We can close that pull request and delete the branch if this version of the fix gets merged instead.

One small note: I see that you fixed the issue by catching an exception instead of ensuring it doesn't occur by performing a greater-than check. This code could potentially execute in a tight (update) loop, so throwing an exception and catching it every loop iteration, in a loop that's supposed to run at a specific frequency may not be the best thing long-term.

rhaschke:
@bmagyar, can you please rerun the CI tests? I think the test `JointStateControllerTest.valuesOk` is flaky. Locally, it works for me most of the time but occasionally fails.

